PR #1, opened 2022-07-27
feat(HACBS-852): add script to copy applications

davidmogar:
The new script allows to copy applications (and their components) to a target namespace. This will be used to create copies of the dev applications (and components) into the managed namespaces.

Signed-off-by: David Moreno García <damoreno@redhat.com>

### DIFF
--- a/hacbscopy.sh
+++ b/hacbscopy.sh
@@ -1,0 +1,195 @@
+#!/usr/bin/env bash
+#
+# Perform a copy of applications and components to a target namespace.
+
+SHORT_OPTS=a:,h
+LONG_OPTS=applications:,all,help
+OPTS=$(getopt --alternative --name hacbscopy --options $SHORT_OPTS --longoptions $LONG_OPTS -- "$@")
+
+CLEANUP_ANNOTATIONS=(
+    "\"com.redhat.appstudio/component-initial-build-processed\"": "\"true\""
+)
+CLEANUP_KEYS=(
+    ".metadata.finalizers",
+    ".metadata.namespace",
+    ".metadata.ownerReferences",
+    ".metadata.resourceVersion",
+    ".metadata.uid",
+    ".spec.gitOpsRepository",
+    ".status"
+)
+CLEANUP_COMMAND="jq 'del(${CLEANUP_KEYS[@]})' | jq '.metadata.annotations += {${CLEANUP_ANNOTATIONS[@]}}'"
+NAMESPACED_NAME_REGEX="^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$"
+QUERY_ARGS="--no-headers -o custom-columns=NAMESPACE:.metadata.namespace,NAME:.metadata.name"
+REQUIREMENTS="kubectl jq oc"
+#######################################
+# Check if all required tools are installed
+# Globals:
+#   REQUIREMENTS
+# Arguments:
+#   None
+# Outputs:
+#   Writes errors messages to stderr
+#######################################
+check_requirements() {
+    for tool in $REQUIREMENTS; do
+        if ! [ -x "$(command -v $tool)" ]; then
+            echo "Error: $tool is not installed" >&2
+            exit 3
+        fi
+    done
+
+    oc status &>/dev/null || echo "Error: Not logged in into a cluster" >&2 && exit 4
+}
+
+#######################################
+# Copy a given application to a the target workspace.
+# Globals:
+#   CLEANUP_COMMAND
+#   NAMESPACED_NAME_REGEX
+#   WORKSPACE
+# Arguments:
+#   Namespaced name of the application to copy
+#######################################
+copy_application() {
+    if [[ ! "$1" =~ $NAMESPACED_NAME_REGEX ]]; then
+        echo "Ignoring application '$1' as it's not a valid namespaced name"
+    fi
+
+    namespace=$(echo "$1" | cut -d '/' -f 1)
+    application=$(echo "$1" | cut -d '/' -f 2)
+
+    echo "Copying application '$application' to namespace '$WORKSPACE'"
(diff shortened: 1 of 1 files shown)


hacbscopy.sh
```diff
@@ -0,0 +1,104 @@
...
+    WORKSPACE=${@:OPTIND:1}
+}
+
+show_help() {
+    echo "Usage:"
+    echo "  hacbscopy [OPTION]... WORKSPACE"
+    echo -e "\nCopy all applications and related resources from one workspace to another.\n"
+    echo "Options:"
+    echo "  -a,--applications if specified, only the given applications will be targeted"
+    echo "  --all             copy resources from all namespaces"
```
johnbieren, line 99: Can we expand on this help a little bit? As a user, I don't understand. Copy all resources from all namespaces? On line 96 we say all applications, let's do that here too

hacbscopy.sh
```diff
@@ -0,0 +1,104 @@
...
+        exit 2
+    fi
+
+    WORKSPACE=${@:OPTIND:1}
+}
+
+show_help() {
+    echo "Usage:"
+    echo "  hacbscopy [OPTION]... WORKSPACE"
+    echo -e "\nCopy all applications and related resources from one workspace to another.\n"
```
johnbieren, line 96: Isn't "related resources" just components?

hacbscopy.sh
```diff
@@ -0,0 +1,104 @@
...
+        exit 2
+    fi
+
+    if [ "$#" -lt 1 ] ; then
+        echo "Target workpace has to be specified as a positional argument"
+        exit 2
+    fi
+
+    if [ "$#" -gt 1 ] ; then
+        echo "Only the target workspace positional argument is required"
```
johnbieren, line 86: I don't like "required" here. IIUC, this is checking you didn't pass too many arguments. Maybe `s/required/accepted` or "Passing more than the target workspace positional argument is not supported"

hacbscopy.sh
johnbieren: This is also the only line of the help ending in a period

hacbscopy.sh
```diff
@@ -0,0 +1,104 @@
...
+    fi
+
+    for application in "${applications[@]}"
+    do
+        copy_application "$application"
+    done
+}
+
+copy_components() {
+    readarray -t components < <(kubectl get component -n "$1" -o jsonpath="{range .items[?(@.spec.application==\"$2\")]}{.metadata.name}{\"\n\"}{end}")
```
johnbieren, line 53: I think future devs or users of this script would appreciate a comment for this line to say what is going on

johnbieren (changes requested):

hacbscopy.sh
davidmogar: Regarding the period, I'm copying the format of `getopt` help. The description has a period.

hacbscopy.sh
davidmogar: Changed to `Only one positional argument (target workspace) is accepted`. What do you think?

johnbieren: works for me

johnbieren (approved):

hacbscopy.sh
```diff
@@ -0,0 +1,195 @@
...
+# Arguments:
+#   Namespace of the application owning the components
+#   Name of the application owning the components
+#######################################
+copy_components() {
+    # Read all components into an array
+    readarray -t components < <(kubectl get component -n "$1" -o jsonpath="{range .items[?(@.spec.application==\"$2\")]}{.metadata.name}{\"\n\"}{end}")
+
+    for component in "${components[@]}"; do
+        echo "Copying component '$component' to namepsace '$WORKSPACE'"
```
theflockers, line 115: s/namepsace/namespace/g

hacbscopy.sh
```diff
@@ -0,0 +1,195 @@
...
+#######################################
+show_help() {
+    echo "Usage:"
+    echo "  hacbscopy [OPTION]... WORKSPACE"
+    echo -e "\nCopy all applications and components from one workspace to another.\n"
+    echo "Options:"
+    echo "  -a,--applications if specified, only the given applications will be targeted"
+    echo "  --all             copy all applications and components from all namespaces"
+    echo "                    to the target one"
+    echo -e "\n  -h, --help        display this help"
```
theflockers, line 190: You could use just one echo here to not create unnecessary procs just to output a text. You could also use `cat` instead. Just a suggestion as well.

```
$ cat <<EOF 
crazy long text
EOF
```

theflockers (approved):
Some comments added as suggestions. Overall LGTM.

hacbscopy.sh
johnbieren: can't believe I missed this one

hacbscopy.sh
```diff
@@ -0,0 +1,195 @@
...
+# Arguments:
+#   Namespaced name of the application to copy
+#######################################
+copy_application() {
+    if [[ ! "$1" =~ $NAMESPACED_NAME_REGEX ]]; then
+        echo "Ignoring application '$1' as it's not a valid namespaced name"
+    fi
+
+    namespace=$(echo "$1" | cut -d '/' -f 1)
+    application=$(echo "$1" | cut -d '/' -f 2)
```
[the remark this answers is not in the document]
johnbieren: As someone who has certainly spent time in bash and saw that notation in Jon's PR and had no idea what it did, I am in favor of the cut method. Very much just my opinion, but I think cut is more commonly understood and easier to read

hacbscopy.sh
davidmogar: True! I like to avoid using that though. It fucks up indentation :sweat_smile: 

Same function would be:
```
show_help() {
    echo <<EOF
Usage:
...
EOF
}
```

hacbscopy.sh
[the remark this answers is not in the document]
davidmogar: I didn't know this trick until recently actually. I also prefer the cut. Using that is like dancing with a wizard and waking up naked and realizing you are a lynx now.

hacbscopy.sh
davidmogar: Good eyes! Fixed.

PR #7, opened 2023-01-12
feat(HACBS-1259): add python script to upload data to pyxis

johnbieren:
This commit adds a reworked version of the create-container-image script from the openshift ecosystem team to enable pushing container image metadata to pyxis.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -1,0 +1,170 @@
+#!/usr/bin/env python
+import argparse
+from urllib.parse import quote
+from datetime import datetime
+import json
+import logging
+from typing import Any, Dict, List
+from urllib.parse import urljoin
+
+import pyxis
+
+LOGGER = logging.getLogger("pyxis")
+
+
+def setup_argparser() -> Any:  # pragma: no cover
+    """Setup argument parser
+
+    :return: Initialized argument parser
+    """
+
+    parser = argparse.ArgumentParser(description="ContainerImage resource creator.")
+
+    parser.add_argument(
+        "--pyxis-url",
+        default="https://pyxis.com",
+        help="Base URL for Pyxis container metadata API",
+    )
+    parser.add_argument(
+        "--certified", help="Is the ContainerImage certified?", required=True
+    )
+    parser.add_argument(
+        "--tag",
+        help="The ContainerImage tag name to upload",
+        required=True,
+    )
+    parser.add_argument(
+        "--skopeo-result",
+        help="File with result of `skopeo inspect` running against image"
+        " represented by ContainerImage to be created",
+        required=True,
+    )
+    parser.add_argument(
+        "--is-latest",
+        help="Should the `latest` tag of the ContainerImage be overwritten?",
+        required=True,
+    )
+    parser.add_argument("--verbose", action="store_true", help="Verbose output")
+    return parser
+
+
+def image_already_exists(args, digest: str) -> bool:
+    """Function to check if a containerImage with the given digest
+    already exists in the pyxis instance
+
+    :return: True if one exists, else false
+    """
+
+    # quote is needed to urlparse the quotation marks
+    filter_str = quote(f'docker_image_digest=="{digest}";' f"not(deleted==true)")
+
+    check_url = urljoin(args.pyxis_url, f"v1/images?page_size=1&filter={filter_str}")
+
+    # Get the list of the ContainerImages with given parameters
+    rsp = pyxis.get(check_url)
+    rsp.raise_for_status()
+
+    query_results = rsp.json()["data"]
+
+    if len(query_results) == 0:
+        LOGGER.info("Image with given docker_image_digest doesn't exist yet")
+        return False
+
+    LOGGER.info(
+        "Image with given docker_image_digest already exists."
+        "Skipping the image creation."
+    )
+    if "_id" in query_results[0]:
+        LOGGER.info(f"The image id is: {query_results[0]['_id']}")
+
+    return True
+
+
+def prepare_parsed_data(skopeo_result: Dict[str, Any]) -> Dict[str, Any]:
+    """Function to extract the data this script needs from provided skopeo inspect output
+
+    :return: Dict of tuples containing pertinent data
+    """
+
+    return {
+        "digest": skopeo_result.get("Digest", ""),
+        "docker_version": skopeo_result.get("DockerVersion", ""),
+        "layers": skopeo_result.get("Layers", []),
+        "name": skopeo_result.get("Name", ""),
+        "architecture": skopeo_result.get("Architecture", ""),
+        "env_variables": skopeo_result.get("Env", []),
+    }
+
+
+def create_container_image(args, parsed_data: Dict[str, Any]):
+    """Function to create a new containerImage entry in a pyxis instance"""
+
+    LOGGER.info("Creating new container image")
+
+    date_now = datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%f+00:00")
+
+    if "digest" not in parsed_data:
+        raise Exception(f"Digest was not found in the passed skopeo inspect json")
+    if "name" not in parsed_data:
+        raise Exception(f"Name was not found in the passed skopeo inspect json")
+    docker_image_digest = parsed_data["digest"]
+    # digest isn't accepted in the parsed_data payload to pyxis
+    del parsed_data["digest"]
+    docker_image_registry = parsed_data["name"].split("/")[0]
+    docker_image_repo = parsed_data["name"].split("/", 1)[1]
+    # name isn't accepted in the parsed_data payload to pyxis
+    del parsed_data["name"]
+
+    upload_url = urljoin(args.pyxis_url, f"v1/images")
+    container_image_payload = {
+        "repositories": [
+            {
+                "published": False,
+                "registry": docker_image_registry,
+                "repository": docker_image_repo,
+                "push_date": date_now,
+                "tags": [
+                    {
+                        "added_date": date_now,
+                        "name": args.tag,
+                    },
+                ],
+            }
+        ],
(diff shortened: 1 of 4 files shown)


pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+from urllib.parse import quote
+from datetime import datetime
+import json
+import logging
+from typing import Any, Dict, List
+from urllib.parse import urljoin
+
+import pyxis
+
+LOGGER = logging.getLogger("operator-cert")
```
scoheb, line 11: minor...but this does not make sense.

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+    del parsed_data["digest"]
+    docker_image_repo = parsed_data["name"].split('/', 1)[1]
+    del parsed_data["name"]
+
+    upload_url = urljoin(args.pyxis_url, f"v1/images")
+    container_image_payload = {
+        "repositories": [
+            {
+                "published": False,
+                "registry": "quay.io",
```
scoheb, line 105: will this always be true? can't we detect the registry but the pull spec?

pyxis/pyxis.py
```diff
@@ -0,0 +1,189 @@
...
+import os
+import sys
+from typing import Any, Dict, Optional, Tuple
+from urllib.parse import urljoin
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+import requests
+
+LOGGER = logging.getLogger("operator-cert")
```
scoheb, line 11: same as above

scoheb (changes requested):

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+        " represented by ContainerImage to be created",
+        required=True,
+    )
+    parser.add_argument("--is-latest", help="Should the `latest` tag of the ContainerImage be overwritten?", required=True)
+    parser.add_argument("--verbose", action="store_true", help="Verbose output")
+    return parser
+
+
+def check_if_image_already_exists(args, digest: str) -> bool:
+    # quote is needed to urlparse the quotation marks
```
davidmogar, line 48: Missing docstring (there's another function with it).

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+
+    LOGGER.info(
+        "Image with given docker_image_digest already exists."
+        "Skipping the image creation."
+    )
+    return True
+
+
+def prepare_parsed_data(skopeo_result: Dict[str, Any]) -> Dict[str, Any]:
+    return {
```
davidmogar, line 76: Missing docstring (there's another function with it).

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+        "architecture": skopeo_result.get("Architecture", ""),
+        "env_variables": skopeo_result.get("Env", []),
+    }
+
+
+def create_container_image(
+    args,
+    parsed_data : Dict[str, Any],
+):
+    LOGGER.info("Creating new container image")
```
davidmogar, line 90: Missing docstring (there's another function with it).

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+
+    # Get the list of the ContainerImages with given parameters
+    rsp = pyxis.get(check_url)
+    rsp.raise_for_status()
+
+    query_results = rsp.json()["data"]
+
+    if len(query_results) == 0:
+        LOGGER.info(
+            "Image with given docker_image_digest doesn't exist yet"
```
davidmogar, line 64: Without breaking the line, this would be shorted than other lines in this very same file.

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+        "digest": skopeo_result.get("Digest", ""),
+        "docker_version": skopeo_result.get("DockerVersion", ""),
+        "layers": skopeo_result.get("Layers", []),
+        "name": skopeo_result.get("Name", ""),
+        "architecture": skopeo_result.get("Architecture", ""),
+        "env_variables": skopeo_result.get("Env", []),
+    }
+
+
+def create_container_image(
```
davidmogar, line 86: Without breaking the line, this would be shorted than other lines in this very same file.

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+    args,
+    parsed_data : Dict[str, Any],
+):
+    LOGGER.info("Creating new container image")
+
+    date_now = datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%f+00:00")
+
+    # digest and name aren't accepted in the parsed_data payload to pyxis
+    docker_image_digest = parsed_data["digest"]
```
davidmogar, line 95: Are you ok with the script failing on runtime if the key doesn't exist?

pyxis/create_container_image.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+def main():  # pragma: no cover
+    """
+    Main func
+    """
+
+    parser = setup_argparser()
+    args = parser.parse_args()
+    log_level = "DEBUG" if args.verbose else "INFO"
```
davidmogar, line 142: I might be wrong, but I think there are actual variables in `logging` for the different levels. Cleaner if they exist.

pyxis/create_container_image.py
johnbieren: Good catch, fixed

pyxis/create_container_image.py
johnbieren: Fixed, thanks

pyxis/pyxis.py
johnbieren: Good catch

pyxis/create_container_image.py
```diff
@@ -0,0 +1,157 @@
...
+    pyxis.setup_logger(level=log_level)
+
+    with open(args.skopeo_result) as json_file:
+        skopeo_result = json.load(json_file)
+
+    parsed_data = prepare_parsed_data(skopeo_result)
+
+    exists = check_if_image_already_exists(args, parsed_data["digest"])
```
davidmogar, line 150: A better name could be ``image_eixst` or does_image_exist` maybe as it returns a boolean. You could then use it directly on the if without creating a variable that will not be used afterwards.

pyxis/pyxis.py
```diff
@@ -0,0 +1,189 @@
...
+        url (str): Pyxis API URL
+        auth_required (bool): Whether authentication should be required for the session
+
+    Raises:
+        Exception: Exception is raised when auth ENV variables are missing.
+
+    Returns:
+        requests.Session: Pyxis session
+    """
+    cert = os.environ.get("PYXIS_CERT_PATH")
```
davidmogar, line 30: Should this two be constants?

pyxis/pyxis.py
```diff
@@ -0,0 +1,189 @@
...
+    key = os.environ.get("PYXIS_KEY_PATH")
+
+    session = requests.Session()
+    add_session_retries(session)
+
+    if auth_required:
```
davidmogar, line 36: Early returns would make this code more manageable.

```
if not auth_required:
    LOGGER.debug("Pyxis session without authentication is created")
    return session

if cert and key:
....
```

pyxis/create_container_image.py
```diff
@@ -0,0 +1,157 @@
...
+    Setup argument parser
+
+    Returns:
+        Any: Initialized argument parser
+    """
+    parser = argparse.ArgumentParser(description="ContainerImage resource creator.")
+
+    parser.add_argument(
+        "--pyxis-url",
+        default="https://pyxis.stage.engineering.redhat.com",
```
johnbieren, line 25: @scoheb can I put the public endpoint here? Or is it private

pyxis/pyxis.py
johnbieren: not sure what you mean by this

pyxis/create_container_image.py
johnbieren: Yes, it should fail and do nothing if the key doesn't exist. Do you think I should raise a custom exception saying as such in this case, or just leave it and let the normal python failure tell the story?

pyxis/test_create_container_image.py
```diff
@@ -0,0 +1,170 @@
...
+
+    # Image already exist
+    mock_rsp.json.return_value = {"data": [{}]}
+
+    # Act
+    exists = image_already_exists(args, digest)
+    # Assert
+    assert exists
+    mock_get.assert_called_with(
+        "https://catalog.redhat.com/api/containers/v1/images?page_size=1&filter=docker_image_digest%3D%3D%22some_digest%22%3Bnot%28deleted%3D%3Dtrue%29"
```
theflockers, line 31: should the url be a variable or constant as it repeats several times?

pyxis/create_container_image.py
```diff
@@ -0,0 +1,162 @@
+import argparse
```
theflockers, line 1: Maybe add the python _shebang_

pyxis/create_container_image.py
davidmogar: Probably a cleaner error would be better, but I'm also ok with leaving it like this. Your call :wink:

pyxis/pyxis.py
davidmogar: The strings (eg. PYXIS_CERT_PATH) being stored in constants.

pyxis/create_container_image.py
johnbieren: fixed

pyxis/create_container_image.py
johnbieren: fixed

davidmogar (approved):

PR #8, opened 2023-01-18
feat(HACBS-1635): add release-utils image that updates on push

johnbieren:

### DIFF
--- a/.github/workflows/dockerfile_push.yaml
+++ b/.github/workflows/dockerfile_push.yaml
@@ -1,0 +1,32 @@
+name: Push Dockerfile to Quay
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'Dockerfile'
+jobs:
+  push-dockerfile:
+    name: Push dockerfile
+    runs-on: ubuntu-20.04
+    env:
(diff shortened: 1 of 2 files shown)


.github/workflows/dockerfile_push.yaml
```diff
@@ -0,0 +1,32 @@
...
+    branches:
+      - main
+    paths:
+      - 'Dockerfile'
+jobs:
+  push-dockerfile:
+    name: Push dockerfile
+    runs-on: ubuntu-20.04
+    env:
+      QUAY_ORG: hacbs-release
```
davidmogar, line 13: Maybe we could define this two as environment variables in GitHub itself (Settings > Secrets and variables > Actions > Variables > Environment variables) but I'm ok with this as well.

davidmogar (approved):

.github/workflows/dockerfile_push.yaml
johnbieren: I personally like to have it as self contained as possible. I'd put the secrets in there too if it weren't for them needing to be secret. So since you are fine with this, I merged as is

PR #9, opened 2023-01-23
fix: clone repo in dockerfile to /home

johnbieren:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,4 +8,5 @@ RUN dnf -y --setopt=tsflags=nodocs install \
 # Set HOME variable to something else than `/` to avoid 'permission denied' problems when writing files.
 ENV HOME=/tekton/home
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,4 +8,5 @@ RUN dnf -y --setopt=tsflags=nodocs install \
 # Set HOME variable to something else than `/` to avoid 'permission denied' problems when writing files.
 ENV HOME=/tekton/home
 
-RUN git clone https://github.com/hacbs-release/release-utils ~/release-utils
+# The ~ dir seems to be mounted over in tekton tasks, so put in /home
+RUN git clone https://github.com/hacbs-release/release-utils /home/release-utils
```
theflockers, line 12: Has the run user permissions to write at `/home`? Normally it is only writable by root.

johnbieren: In my test, yes, I am able to write to `/home/foo` in a tekton task

scoheb (approved):

PR #10, opened 2023-01-24
feat: use python 3.9 in release-utils image

johnbieren:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,8 @@ FROM registry.access.redhat.com/ubi8/ubi
 
 RUN dnf -y --setopt=tsflags=nodocs install \
     git \
-    python3 \
+    python39-devel \
+    python39-requests \
     && dnf clean all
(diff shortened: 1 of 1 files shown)


scoheb (approved):

mmalina:
@johnbieren why not use ubi9 which comes with python 3.9 as default? Also, it would have been nice if the commit message explained why this change is needed. There is no jira linked either, so it would have been nice to have some context.

mmalina:
Yeah, ok, maybe we can switch to ubi9 in the future - it will have lots of packages at newer versions which is usually preferred.

johnbieren:
> Yeah, ok, maybe we can switch to ubi9 in the future - it will have lots of packages at newer versions which is usually preferred.

Definitely agree that switching to ubi9 in the future would be a good thing. Maybe we create a low prio ticket for it. WDYT @davidmogar ?

PR #12, opened 2023-01-25
fix(HACBS-1261): ensure pyxis image was successfully pushed

johnbieren:
After the post is done to pyxis, execute a get to make sure the image metadata was successfully added. This also has the added benefit of the containerImageID being printed out in all scenarios.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -165,6 +165,10 @@ def main():  # pragma: no cover
     if not image_already_exists(args, parsed_data["digest"]):
         create_container_image(args, parsed_data)
 
+    # Make sure image is now available
+    if not image_already_exists(args, parsed_data["digest"]):
+        raise Exception("Image metadata was not successfully added to Pyxis.")
+
 
 if __name__ == "__main__":  # pragma: no cover
     main()


pyxis/create_container_image.py
```diff
@@ -165,6 +165,10 @@ def main():  # pragma: no cover
     if not image_already_exists(args, parsed_data["digest"]):
         create_container_image(args, parsed_data)
 
+    # Make sure image is now available
+    if not image_already_exists(args, parsed_data["digest"]):
+        raise Exception("Image metadata was not successfully added to pyxis.")
```
davidmogar, line 170:
```suggestion
        raise Exception("Image metadata was not successfully added to Pyxis.")
```

davidmogar: Also, why not to include it in the previous if? It's exactly the same check.

davidmogar (changes requested):

pyxis/create_container_image.py
johnbieren: Not sure I understand. The previous if pushes it if it is not there. After the push, we check to see if it the metadata is current and if it is not then raise an exception

davidmogar: Si the parsed_data gets changed? That's confusing to me.

johnbieren: No, the parsed_data doesn't get changed. The parsed_data is the data it needs to check if the image exists. It is just
```
if image doesn't exist
    create it

if image still doesn't exist
    create it didn't work, raise exception
```

davidmogar: Mental fart :dash:

davidmogar (approved):

PR #15, opened 2023-01-31
refactor(HACBS-1261): log container image ID on pyxis post

johnbieren:
This commit reworks the logic of how the container image ID is logged in all instances. It reduces the calls to the api endpoint by parsing the returned data of the pyxis.post.

### DIFF
--- a/pyxis/create_container_image
+++ b/pyxis/create_container_image
@@ -76,6 +76,8 @@ def image_already_exists(args, digest: str) -> bool:
     )
     if "_id" in query_results[0]:
         LOGGER.info(f"The image id is: {query_results[0]['_id']}")
+    else:
+        raise Exception("Image metadata was found in Pyxis, but the id key was missing.")
 
     return True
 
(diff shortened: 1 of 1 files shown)


pyxis/create_container_image.py
```diff
@@ -146,7 +146,13 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
             }
         )
 
-    return pyxis.post(upload_url, container_image_payload)
+    rsp = pyxis.post(upload_url, container_image_payload)
+
+    # Make sure container metadata was successfully added to Pyxis
+    if "_id" in rsp:
+        LOGGER.info(f"The image id is: {rsp['_id']}")
```
mmalina, line 153: This is good. But in the other instance of this `LOGGER.info` in `image_already_exists()` line 77 (it seems I cannot add a comment there directly, since it's not being change), should we perhaps also fail if we can't find the id? It's not very likely there will be a problem there - if the returned list is empty, we return `False`. But I wonder what good the `if` is there. I think we should probably fail as well if we can't find the id there.

johnbieren: Good idea, fixed

mmalina: Cool, thanks.

mmalina:
LGTM!

mmalina (approved):

PR #17, opened 2023-02-09
Improve Github workflow

mmalina:
There are two commits.

The first one will just fix warnings that the workflow currently produces by updating the actions versions.

The second one adds more improvements - the workflow will run on both main branch and PRs and it will also test that the python script(s) can run.


happybhati (approved):

mmalina:
@happybhati thanks for approving, but as you can see, it's not working yet ;) Marking as Draft now.

mmalina:
Hmm, the docker login action fails on:
```
Error: Username and password required
```
That's because when the pull_request event triggers the workflow and it's run in a fork, Github will not pass the secrets.

One option would be to use the `pull_request_target` event:

https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks
> In order to solve this, we’ve added a new pull_request_target event, which behaves in an almost identical way to the pull_request event with the same set of filters and payload. However, instead of running against the workflow and code from the merge commit, the event runs against the workflow and code from the base of the pull request. 

So it would be more complicated to do. I will close this for now and revisit later.

PR #18, opened 2023-02-09
fix: update github workflow to remove warnings

mmalina:
There were some warnings, mainly that the action
versions we were using were using Node.js 12 which is now deprecated.

I also renamed the workflow to more accurately reflect what it's doing.

Also, I'm adding the latest tag for the image. It's nice to have since it's the default when pulling the image without specifying a tag.

Signed-off-by: Martin Malina <mmalina@redhat.com>


happybhati (approved):
LGTM :+1:

johnbieren (approved):

PR #23, opened 2023-02-23
fix: set up logging correctly

mmalina:
The way it was set up, LOGGER calls in pyxis/create_container_image would actually not print anything to stdout. Now it works.

### DIFF
--- a/pyxis/pyxis.py
+++ b/pyxis/pyxis.py
@@ -152,25 +152,21 @@ def add_session_retries(
     session.mount("https://", adapter)
 
 
-def setup_logger(level: int = logging.INFO, log_format: Any = None) -> Any:
+def setup_logger(level: int = logging.INFO, log_format: Any = None):
     """Set up and configure 'pyxis' logger.
     Args:
         level (str, optional): Logging level. Defaults to logging.INFO.
         log_format (Any, optional): Logging message format. Defaults to None.
     :return: Logger object
     """
-
-    logger = logging.getLogger("pyxis")
-    logger.propagate = False
-    logger.setLevel(level)
-
     if log_format is None:
         log_format = "%(asctime)s [%(name)s] %(levelname)s %(message)s"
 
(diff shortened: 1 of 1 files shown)


davidmogar (approved):

mmalina:
> I was thinking the pyxis task was a bit quiet.... thanks!

@scoheb but keep in mind that in order for this to take effect in the pipeline, its reference to the `release-utils` image would need to be updated. Maybe once we have e2e tests for the pipelines we can switch to always using the latest image... It would also provide better security - we wouldn't be using old images that are bound to contain vulnerabilities sooner or later.

davidmogar:
> > I was thinking the pyxis task was a bit quiet.... thanks!
> 
> @scoheb but keep in mind that in order for this to take effect in the pipeline, its reference to the `release-utils` image would need to be updated. Maybe once we have e2e tests for the pipelines we can switch to always using the latest image... It would also provide better security - we wouldn't be using old images that are bound to contain vulnerabilities sooner or later.

Using the latest image is fine only if we don't use `:latest`. We need to avoid this. First because that's a moving tag. Second because it's more difficult to track problems.

mmalina:
> Using the latest image is fine only if we don't use `:latest`. We need to avoid this. First because that's a moving tag. Second because it's more difficult to track problems.

I would think "Using the latest image" = using `:latest` ;) How else would we do it? Having some automation to create a PR for release-bundles to automatically update the reference everywhere?

> First because that's a moving tag. 

Why is that a problem? I know that nowadays it's popular to always specify versions/tags everywhere, but both have pros and cons. I understand the concern if you're using an external resource, but here we're using our own image that we're in charge of. And if we have sufficient tests, it should be ok, no?

> Second because it's more difficult to track problems.

You will still see what image digest/sha was pulled to see exactly which one it was. But yeah, maybe the approach with always pointing to some tag and updating the reference regularly would be better for this.

PR #24, opened 2023-02-23
feat(HACBS-1551): install cosign binary

mmalina:
cosign is needed for sbom download in our tekton tasks. Previously, we thought we could use the cosign container directly in a tekton step, but because we'll have multiple components/images, we will need the binary to call it repeatedly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,7 @@
 FROM registry.access.redhat.com/ubi8/ubi
 
+ARG COSIGNVERSION=1.13.1
+
 RUN dnf -y --setopt=tsflags=nodocs install \
     git \
     jq \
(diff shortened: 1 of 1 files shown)


happybhati (approved):

Dockerfile
```diff
@@ -8,6 +8,9 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     skopeo \
     && dnf clean all
 
+# cosign is used for sbom download
+RUN rpm -ivh https://github.com/sigstore/cosign/releases/download/v1.13.1/cosign-1.13.1.x86_64.rpm
```
scoheb, line 12: should the version be a variable at top?

davidmogar: You mean an argument with a default value? It's a good idea IMHO.

scoheb: or an `ENV COSIGNVERSION=1.13.1`

scoheb: `RUN rpm -ivh https://github.com/sigstore/cosign/releases/download/v${COSIGNVERSION}/cosign-${COSIGNVERSION}.x86_64.rpm`

davidmogar: If that variable is required in the container, why not. If not it's better to use ARG as it allows you to pass a different value on build time.

davidmogar: Actually you can do the same for environment variables... So up to you :wink:

kasemAlem (approved):

scoheb (approved):

PR #26, opened 2023-02-24
feat(HACBS-1772): add black format check github action

johnbieren:


johnbieren:
Depends on https://github.com/hacbs-release/release-utils/pull/25

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,13 @@
...
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
mmalina, line 12: I can see that this is what they suggest in the docs: https://black.readthedocs.io/en/stable/integrations/github_actions.html

But the latest is v3, so I think we should go with that: https://github.com/actions/checkout

I actually saw warnings when using v2 in [a personal project](https://github.com/mmalina/strava-mapping/commit/c0d0f809d09f8aa301e67ce79f6cee4d55e4cff2) recently, not sure which it doesn't show anything here.

johnbieren: fixed

mmalina: Cool, thanks.

mmalina (approved):

PR #27, opened 2023-02-27
fix(HACBS-1772): fix current flake8 errors

johnbieren:
I used a line length of 88 for checking (flake8 default is 79, but black defaults to 88 and we are also using black).


mmalina:
> Sure, the line here is still over 95 though. So, I will leave this PR as is and in the PR adding the flake8 action, I will also set it at 95 instead of 88

Sounds good.

happybhati (approved):

PR #28, opened 2023-02-27
Add flake8 action and set max line length to 95

johnbieren:
This PR has two commits. One changes the max line length for black and flake8 in this repo to be 95 (and adds config files doing that). The second adds a github action to enforce flake8 formatting

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -9,5 +9,9 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: psf/black@stable
+      - name: Check out repo
+        uses: actions/checkout@v3
+      - name: Black Lint
+        uses: psf/black@stable
+        with:
+          options: "--line-length 95"
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -13,5 +13,9 @@ jobs:
         uses: actions/checkout@v3
       - name: Black Lint
         uses: psf/black@stable
+      - name: Setup python environment for flake8 check
+        uses: actions/setup-python@v4
         with:
-          options: "--line-length 95"
+          python-version: "3.9" # Same as in Dockerfile
+      - name: flake8 Lint
+        uses: py-actions/flake8@v2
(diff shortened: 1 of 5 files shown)


.github/workflows/lint.yaml
```diff
@@ -9,5 +9,9 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
-      - uses: psf/black@stable
+      - name: Check out repo
+        uses: actions/checkout@v3
+      - name: Black Lint
+        uses: psf/black@stable
+        with:
+          options: "--line-length 95"
```
mmalina, line 17: Is this really needed? I would have hoped it would pick it up from the config file. But if it's needed, then this should probably also include `--check --diff` which is the default. 

https://black.readthedocs.io/en/stable/integrations/github_actions.html#usage
> Please note that the [--check flag](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#labels-exit-code) is required so that the workflow fails if Black finds files that need to be formatted.

.github/workflows/lint.yaml
```diff
@@ -15,3 +15,11 @@ jobs:
...
         with:
           options: "--line-length 95"
+      - name: Setup python environment for flake8 check
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.9" # Same as in Dockerfile
+      - name: flake8 Lint
+        uses: py-actions/flake8@v2
+        with:
+          max-line-length: 95
```
mmalina, line 25: Same as with black: Is this needed? I would have hoped it would pick it up from the `.flake8` file.

johnbieren: I'll test it on my fork and report back

johnbieren: You seem to be correct. I got `./pyxis/test_create_container_image.py:36:96: E501 line too long (112 > 95 characters)` on a PR on my fork with the options removed (you can also see now that nothing fails in the action although we have lines longer than 79 characters). Nice catch

mmalina: Awesome :)

mmalina (approved):

PR #29, opened 2023-02-28
fix(HACBS-1803): fix current pytest errors

johnbieren:
This PR includes renaming `create_container_image` back to have its .py suffix. This is because the imports don't work properly without it. There are workarounds (see https://stackoverflow.com/questions/2601047/import-a-python-module-without-the-py-extension) but I think it would be nice to just put the suffix back. Open to discussion on it

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -25,9 +25,7 @@ def setup_argparser() -> Any:  # pragma: no cover
         default="https://pyxis.com",
         help="Base URL for Pyxis container metadata API",
     )
-    parser.add_argument(
-        "--certified", help="Is the ContainerImage certified?", required=True
-    )
+    parser.add_argument("--certified", help="Is the ContainerImage certified?", required=True)
     parser.add_argument(
         "--tag",
         help="The ContainerImage tag name to upload",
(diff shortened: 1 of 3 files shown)


pyxis/create_container_image.py
```diff
@@ -25,9 +25,7 @@ def setup_argparser() -> Any:  # pragma: no cover
         default="https://pyxis.com",
         help="Base URL for Pyxis container metadata API",
     )
-    parser.add_argument(
-        "--certified", help="Is the ContainerImage certified?", required=True
-    )
+    parser.add_argument("--certified", help="Is the ContainerImage certified?", required=True)
```
davidmogar, line 28: Is this a flag? In that case is missing the `action='store_true'` isn't it? You can remove the `required` if you have the action.

johnbieren:
Added commit to copy the `create_container_image` script as part of Dockerfile so we can still have the .py suffix on the file, but the `create_container_image` binary will exist for `release-bundles` tasks

pyxis/create_container_image.py
johnbieren: It could be a flag, but it isn't currently used like that (because when I took it from the operator team, they did not have it as a flag)

davidmogar: I saw in the test you assign a boolean (string) value to it. It's weird that is not a flag... anyway :man_shrugging:

happybhati (approved):
LGTM 👍

PR #38, opened 2023-06-14
feat(HACBS-2225): add new utils

davidmogar:
As part of this commit, the Dockerfile is changed so it doesn't depend on a remote repository.

### DIFF
--- a/utils/get-resource
+++ b/utils/get-resource
@@ -1,0 +1,43 @@
+#!/usr/bin/env sh
+#
+# Script: kubectl_script.sh
+#
+# Description: This script receives three parameters: <resource_type>,
+#              <namespaced_name>, and [jsonpath]. It uses `kubectl` to load
+#              the specified Kubernetes resource and print it. If a jsonpath
+#              is supplied, only that part will be printed.
+#
+# Usage: kubectl_script.sh <resource_type> <namespaced_name> [jsonpath]
+#
+# Parameters:
+#   <resource_type>   - The Kubernetes resource type (e.g., pod, deployment, service)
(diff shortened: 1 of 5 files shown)


johnbieren (dismissed):
I think `bash-utils` is a better name than just `utils` but if you don't think mentioning that they are bash, I suppose that is fine

utils/merge-json
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env sh
+
```
mmalina, line 2: Could we have some basic description? Like what this is used for and possibly which task uses it.

davidmogar: Done.

utils/get-resource
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env sh
+
```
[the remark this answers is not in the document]
davidmogar: First done. README for a different PR maybe.

utils/get-resource
```diff
@@ -0,0 +1,43 @@
...
+# Usage: kubectl_script.sh <resource_type> <namespaced_name> [jsonpath]
+#
+# Parameters:
+#   <resource_type>   - The Kubernetes resource type (e.g., pod, deployment, service).
+#   <namespaced_name> - The namespaced name of the resource in the form "namespace/name".
+#   [jsonpath]        - Optional JSONPath expression to filter the output.
+#
+# Requirements:
+#   - This script requires the 'kubectl' cli. Ensure that 'kubectl' is
+#     on yourinstalled
```
johnbieren, line 19: typo

johnbieren (dismissed):

johnbieren (approved):

PR #40, opened 2023-07-10
feat(HACBS-2226): installs glab and adds functions to handle gitlab repos

theflockers:
Jira: https://issues.redhat.com/browse/HACBS-2226

- installs the glab tool
- adds functions to manage git and gitlab repos
- adds Red Hat IT Root CA

Relates to https://github.com/hacbs-release/app-interface-deployments/pull/30

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,13 @@ ARG COSIGN_VERSION=1.13.1
 ARG KUBECTL_VERSION=1.27.2
 ARG OCP_VERSION=4.13.3
 ARG YQ_VERSION=4.34.1
+ARG GLAB_VERSION=1.31.0
 
 RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
     curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl &&\
     curl -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_VERSION}/opm-linux.tar.gz |tar -C /usr/bin -xzf - &&\
-    chmod +x /usr/bin/{yq,kubectl,opm} &&\
+    curl -L https://gitlab.com/gitlab-org/cli/-/releases/v${GLAB_VERSION}/downloads/glab_${GLAB_VERSION}_Linux_x86_64.tar.gz | tar -C /usr -xzf - bin/glab &&\
+    chmod +x /usr/bin/{yq,kubectl,opm,glab} &&\
     rpm -ivh https://github.com/sigstore/cosign/releases/download/v${COSIGN_VERSION}/cosign-${COSIGN_VERSION}.x86_64.rpm
 
 RUN dnf -y --setopt=tsflags=nodocs install \
@@ -19,6 +21,9 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     skopeo \
     && dnf clean all
 
+# Adds Red Hat IT Root CAs
+ADD https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem /etc/pki/ca-trust/source/anchors/
+RUN update-ca-trust
 
 COPY pyxis /home/pyxis
 COPY utils /home/utils
--- a/utils/git-functions
+++ b/utils/git-functions
@@ -1,0 +1,82 @@
+#!/usr/bin/env sh
+#
+# Functions to manage git repositories
+
+git_functions_usage() {
+    echo -e "Usage:"
+    echo -e "\t~$ export GIT_AUTHOR_NAME=<name>"
+    echo -e "\t~$ export GIT_AUTHOR_EMAIL=<email>"
+    echo -e "\t~$ export ACCESS_TOKEN=<github key>"
+    echo -e "\t~$ . git-functions"
+}
+
+git_config_ident() {
+    git config user.name "${GIT_AUTHOR_NAME}" 
+    git config user.email "${GIT_AUTHOR_EMAIL}"
+}
+
+git_commit_and_push() {
+    OPTIONS=$(getopt -l "branch:,message:" -o "b:m:" -a -- "$@")
+    eval set -- "$OPTIONS"
+    while true; do
+        case "$1" in
+            -b|--branch)
+                shift
+                BRANCH=$1
+                ;;
+            -m|--message)
+                shift
+                COMMIT_MSG="$1"
+                ;;
+            --)
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+    git checkout -b $BRANCH
+    git_config_ident
+    git commit -a -m "$COMMIT_MSG"
+    git push origin $BRANCH
+}
+
+git_clone_and_checkout() {
+    OPTIONS=$(getopt -l "repository:,revision:" -o "r:v:" -a -- "$@")
+    eval set -- "$OPTIONS"
+    while true; do
+        case "$1" in
+            -r|--repository)
+                shift
+                REPOSITORY=$1
+                ;;
+            -v|--revision)
+                shift
+                REVISION="$1"
+                ;;
+            --)
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+
+    echo "Cloning the Repository..."
+
+    REPO_DIR=$(basename ${REPOSITORY} .git)
+
+    # rewrites the repo url with the auth token
+    CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPOSITORY#*://}"
+    git clone $CLONE_URL
+    cd $REPO_DIR
+    git checkout $REVISION
+}
+
+git_functions_init() {
+  # an actual user should be set
+  [ -z "${GIT_AUTHOR_NAME}" -o -z "${GIT_AUTHOR_EMAIL}" ] && git_functions_usage && return 1
+
+  # it should not continue if ACCESS_TOKEN
+  [ -z "${ACCESS_TOKEN}" ] && git_functions_usage && return 1
+}
--- a/utils/gitlab-functions
+++ b/utils/gitlab-functions
@@ -1,0 +1,52 @@
+#!/usr/bin/env sh
+#
+# Functions to manage gitlab repos.
+#
+# This library requires the `glab` cli (https://docs.gitlab.com/ee/integration/glab/)
+
+gitlab_usage() {
+    echo -e "Usage:"
+    echo -e "\t~$ export GITLAB_HOST=<host>"
+    echo -e "\t~$ export ACCESS_TOKEN=<github key>"
+    echo -e "\t~$ . gitlab-functions"
+}
+
+gitlab_auth() {
+    gitlab_host=$1
+    glab auth login -h $gitlab_host --stdin <<< "${ACCESS_TOKEN}"
+}
+
+gitlab_create_mr() {
+    OPTIONS=$(getopt -l "head:,title:,description:" -o "h:t:d:" -a -- "$@")
+    eval set -- "$OPTIONS"
+    while true; do
+        case "$1" in
+            -h|--head)
+                shift
+                HEAD="$1";
+                ;;
+            -t|--title)
+                shift
+                TITLE="$1";
+                ;;
+            -d|--description)
+                shift
+                DESCRIPTION="$1";
+                ;;
+            --)
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+    glab mr create --title "${TITLE}" -b ${HEAD} --description "${DESCRIPTION}" | \
+        awk '/merge_request/ {print "merge_request: "$1}' |yq -o json
+}
+
+gitlab_init() {
+  # it should not continue if ACCESS_TOKEN or GITLAB_HOST is missing
+  [ -z "${ACCESS_TOKEN}" -o -z "${GITLAB_HOST}" ] && gitlab_usage && return 1
+
+  gitlab_auth ${GITLAB_HOST}
+}


functions/gitlab
```diff
@@ -0,0 +1,71 @@
...
+  if [ "$#" -lt 1 -a "$1" == "_x" ]; then
+      show_help
+      return
+  else 
+      GITLAB_HOST=$1
+      glab_auth $GITLAB_HOST
+  fi
+}
+
+init_gitlab $1
```
mmalina, line 71: The way this is set up, I would expect this file to only contain function definitions, so a user of this would source the file (`source functions/gitlab`) and then run whichever functions they wish. But here we call one of the functions already - that doesn't seem right. Same way as a python script shouldn't run anything if you merely import it.

So I think this line should be removed and the `init_gitlab` function should be called in the task where this is used.

functions/gitlab
```diff
@@ -0,0 +1,71 @@
...
+# Library: gitlab
+#
+# Shell library for GitLab Operations
+#
+#
+show_help() {
+    echo -e "Usage:"
+    echo -e "\t~$ export GIT_AUTHOR_NAME=<name>"
+    echo -e "\t~$ export GIT_AUTHOR_EMAIL=<email>"
+    echo -e "\t~$ export ACCESS_KEY=<github key>\n\t~$ . gitlab <gitlab host>"
```
mmalina, line 12: Any reason why the lines above are separate, but here two lines are combined (with `\n`)? I think it would improve readability if all the lines were separate.

functions/gitlab
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env sh
+#
+# Library: gitlab
+#
+# Shell library for GitLab Operations
+#
+#
+show_help() {
```
mmalina, line 8: Once the function file is loaded, these functions will be globally available in the given shell session. I think it would be better to have a more specific name, e.g. `gitlab_usage`.

functions/gitlab
mmalina: Also, here it says `ACCESS_KEY`, but in the init function, I can see that `ACCESS_TOKEN` is used.

functions/gitlab
```diff
@@ -0,0 +1,71 @@
...
+    CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPO_URL#*://}"
+    test ! -d $REPO_DIR && git -c http.sslVerify=false clone $CLONE_URL
+    cd $REPO_DIR
+    git remote set-url origin $CLONE_URL
+    git checkout $REV
+}
+
+init_gitlab() {
+  # an actual user should be set
+  [ -z "${GIT_AUTHOR_NAME}" -o -z "${GIT_AUTHOR_EMAIL}" ] && show_help && return
```
mmalina, line 56: I think this should return a non-zero code in such case. Same applies to the other ones below as well.

johnbieren (commented):
Can you change the commit title? I know that is the ticket description, but this commit is just about adding functions to interact with gitlab. It doesn't have anything to do with tekton tasks

functions/gitlab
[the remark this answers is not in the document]
theflockers: Misread your last phrase. I had the impression that it had only executables, not include files. If that is the case, indeed. I can move it to utils.

functions/gitlab
```diff
@@ -0,0 +1,71 @@
...
+
+glab_auth() {
+    gitlab_host=$1
+    glab config set skip_tls_verify true --host $gitlab_host
+    glab auth login -h $gitlab_host --stdin <<< "${ACCESS_TOKEN}"
+}
+
+glab_create_mr() {
+    read BRANCH HEAD <<< "${1} ${2}"
+    glab mr create --title "chore: fileUpdates changes" -b $HEAD --description "chore: fileUpdates changes" | \
```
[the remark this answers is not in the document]
theflockers: I noticed (after having a bunch of MR open) that I will also need to add some unique identification in the title. @mmalina @johnbieren . I will change this as well.

functions/gitlab
mmalina: I think this concern is still valid. The file was moved to `utils/`, but it's called `gitlab-functions` and it says that it's a library at the top of the file. So I still believe it shouldn't really do anything upon being sourced.

functions/gitlab
mmalina: It's better now. But on related topic, I think it would be nice if all the functions had a common prefix, say `gitlab_`, that way it's clearer that it all comes from this one place - especially when looking at the script that uses the functions in the tekton task.

functions/gitlab
theflockers: No actual reason for the separation. Just missed to format it properly. Fixed also to `ACCESS_TOKEN`  instead.

mmalina: cool, thanks.

functions/gitlab
mmalina: Yeah, it would probably make sense to have a separate one for `git-` functions.

johnbieren (dismissed):
I'm good with the code once Martin's concerns are addressed

functions/gitlab
theflockers: done. Now the init is called in the task.

functions/gitlab
theflockers: done. Added `git-functions` file.

utils/git-functions
```diff
@@ -0,0 +1,85 @@
...
+            --)
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+    git checkout -b $BRANCH
+    git_config_ident
+    git add . && git commit -m "$COMMIT_MSG"
```
mmalina, line 42: You can just do `git commit -a -m "$COMMIT_MSG"` and the `git add .` is not needed.

If you did want to do this separately for some reason, I would suggest just running each command on a separate line - there is no reason to combine the two this way.

utils/git-functions
```diff
@@ -0,0 +1,85 @@
...
+        shift
+    done
+
+    echo "Cloning the Repository..."
+
+    REPO_DIR=$(basename ${REPOSITORY} .git)
+
+    # rewrites the repo url with the auth token
+    CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPO_URL#*://}"
+    test ! -d $REPO_DIR && git -c http.sslVerify=false clone $CLONE_URL
```
mmalina, line 73: Based on this, I assume $REPO_DIR is required to exist already. Why is that the case? The `git clone` will create it for you, right?

Also, why are specifically these two commands combined with `&&`? If the test fails, the whole function will hopefully fail and exit anyway, right? Also, in general in such tests it's usually nicer to use `if` and explicitly print what's wrong, like:
```sh
if ! [ -d $REPO_DIR ]
then
    echo Error: $REPO_DIR does not exist.
    return 1
fi
```

utils/git-functions
```diff
@@ -0,0 +1,85 @@
...
+    cd $REPO_DIR
+    git remote set-url origin $CLONE_URL
```
mmalina, line 75: Will this not be done automatically? AFAIK, when you clone a new repo, you automatically have a remote called `origin` with the url set to what you cloned. Is this not the case here? Perhaps because of the non-standard url with oauth2? If that's the case, I think it warrants a comment.

utils/gitlab-functions
```diff
@@ -0,0 +1,54 @@
...
+gitlab_usage() {
+    echo -e "Usage:"
+    echo -e "\t~$ export GITLAB_HOST=<host>"
+    echo -e "\t~$ export ACCESS_TOKEN=<github key>"
+    echo -e "\t~$ . gitlab-functions"
+}
+
+gitlab_auth() {
+    gitlab_host=$1
+    glab config set skip_tls_verify true --host $gitlab_host
```
mmalina, line 17: Again, we should not have to skip ssl/tls verification. And if we can get rid of that, then this function is probably not needed, since it would only call one command, right?

utils/gitlab-functions
```diff
@@ -0,0 +1,54 @@
...
+                DESCRIPTION="$1";
+                ;;
+            --)
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+    glab mr create --title "${TITLE}" -b ${HEAD} --description "${DESCRIPTION}" | \
```
mmalina, line 45: Hmm, I'm starting to wonder if we need any gitlab functions at all - if we remove `gitlab_init` and `gitlab_usage` as I suggested elsewhere. And `gitlab_auth` too. Then all we're left with is this `glab mr create` wrapper which can easily be done inline.

Do we at least anticipate to call this from multiple places? I wonder what you and others think.

utils/git-functions
theflockers: No reason. I will change to use `-a` instead.

utils/git-functions
```diff
@@ -0,0 +1,85 @@
...
+                shift
+                break
+                ;;
+        esac
+        shift
+    done
+    git checkout -b $BRANCH
+    git_config_ident
+    git add . && git commit -m "$COMMIT_MSG"
+    git -c http.sslVerify=false push origin $BRANCH
```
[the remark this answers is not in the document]
theflockers: `glab` does not like the certificate.  I will search where I might pull the CA to add to the image.

utils/git-functions
theflockers: Thanks for notice this. This is actually a test leftover.

In my tests I was listing the same repo twice - so each cycle it would need to clone the repo again. If the dir exists, git fails and the script exits, which is not desirable.

This is not the case anymore because now the script always cleans up the workspace after the MR is created, so for the next clone it starts freshly. I will remove the `test ! -d`

utils/git-functions
theflockers: I have to set this otherwise git will expose the credentials when pushing.

utils/gitlab-functions
theflockers: same problem as above. I will find out where I can fetch the CA and add it to the image.

utils/gitlab-functions
theflockers: The `gitlab_init` and `gitlab_usage` it is just a way to let the user know that all vars required are set. 

The end goal was having something that can be reusable otherwise I would just leave everything in line in the task. 

If all this looks too much, I am (not so) happy to make it simpler.

utils/git-functions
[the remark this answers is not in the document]
theflockers: done. CA added and ssl verify bypass removed.

utils/git-functions
theflockers: `test ! -d` and ssl verify bypass removed.

utils/git-functions
theflockers: I will remove it. It looked to build old builds, it happened only on github (my first PR had it). If we need it in the future, we can add this again.

utils/gitlab-functions
johnbieren: I think we should keep a gitlab functions file. Once we are doing things like automerging or whatever, it will likely be different between gitlab and github

utils/gitlab-functions
```diff
@@ -0,0 +1,54 @@
...
+        awk '/merge_request/ {print "merge_request: "$1}' |yq -o json
+}
+
+gitlab_init() {
```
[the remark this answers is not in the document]
theflockers: BTW, should we conform to a styleguide? Maybe https://google.github.io/styleguide/shellguide.html

Dockerfile
```diff
@@ -19,6 +21,9 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     skopeo \
     && dnf clean all
 
+# Adds Red Hat IT Root CAs
+ADD https://certs.corp.redhat.com/certs/2015-IT-Root-CA.pem https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem /etc/pki/ca-trust/source/anchors/
```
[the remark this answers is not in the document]
theflockers: Not  sure TBH. Anyhow the CA has only public information as the public key, for instance.

utils/gitlab-functions
theflockers: +1. I prefer to keep them.

functions/gitlab
mmalina: Thanks.

utils/gitlab-functions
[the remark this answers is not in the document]
mmalina: Thanks for linking Google's style guide. It definitely looks nice. Not sure if we need to strictly follow it, we can discuss later.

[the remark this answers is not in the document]
mmalina: > Just made a simpler comment in the top of the two files. I think it looks better now. Please take a look.

I agree it's better - at least it's consistent now. But I think it should still at least say that `glab` is required to be installed. Don't you agree?

Dockerfile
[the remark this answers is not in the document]
mmalina: Yep, I agree this should be fine - it's done in other places as well.

utils/git-functions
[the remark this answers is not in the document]
mmalina: Awesome, thanks!

utils/git-functions
mmalina: Great, thanks.

utils/git-functions
mmalina: Thanks.

utils/git-functions
mmalina: Cool.

utils/gitlab-functions
mmalina: Alright.

utils/gitlab-functions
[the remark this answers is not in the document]
theflockers: Sure, I agree. I actually forgot to add the info. Done now.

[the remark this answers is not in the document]
mmalina: Cool, thanks.

utils/git-functions
```diff
@@ -0,0 +1,82 @@
...
+        esac
+        shift
+    done
+
+    echo "Cloning the Repository..."
+
+    REPO_DIR=$(basename ${REPOSITORY} .git)
+
+    # rewrites the repo url with the auth token
+    CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPO_URL#*://}"
```
mmalina, line 70: s/REPO_URL/REPOSITORY/ , right? `REPO_URL` doesn't seem to be defined anywhere.

theflockers: Fixed! Please take a look.

mmalina (changes requested):
@theflockers one last thing: Can you alter the commit message so that the title fits and is not split? I think the limit is 72 characters. Also, can you mention that red hat cert was added?

mmalina (approved):

PR #41, opened 2023-07-11
refactor(HACBS-2375): adjust sbom script for pyxis changes

mmalina:
The image objects in Pyxis will no longer contain
embedded content_manifest_components objects in them. We were relying on this, so we have to use
edges instead.

### DIFF
--- a/pyxis/upload_sbom.py
+++ b/pyxis/upload_sbom.py
@@ -127,8 +127,19 @@ query ($id: ObjectIDFilterScalar!) {
             content_manifest {
                 _id
             }
-            content_manifest_components {
-                _id
+            edges {
+                content_manifest {
+                    data {
+                        _id
+                        edges {
+                            components {
+                                data {
+                                    _id
+                                }
+                            }
+                        }
+                    }
+                }
             }
         }
         error {
@@ -242,10 +253,11 @@ mutation ($id: ObjectIDFilterScalar!, $input: ContentManifestComponentInput! ) {
 
 
 def get_existing_component_count(image: dict) -> int:
-    if image["content_manifest_components"] is not None:
-        return len(image["content_manifest_components"])
-    else:
+    content_manifest_data = image["edges"]["content_manifest"]["data"]
+    if content_manifest_data is None:
         return 0
+    else:
+        return len(content_manifest_data["edges"]["components"]["data"])
 
 
 def load_sbom_components(sbom_path: str) -> list[dict]:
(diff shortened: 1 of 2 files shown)


johnbieren (dismissed):

kasemAlem:
so based on this PR , I think I'll need to update in release e2e-tests suite [here](https://github.com/redhat-appstudio/e2e-tests/blob/f2f01144083a9348a74dfb3f1e0853f6b5cb9a3e/pkg/utils/release/sbom.go#L86) only ?

pyxis/upload_sbom.py
```diff
@@ -127,8 +127,19 @@ def get_image(graphql_api: str, image_id: str) -> dict:
             content_manifest {
                 _id
             }
-            content_manifest_components {
-                _id
+            edges {
```
ezopezo, line 130: If your aim is to get all components for specified image, without need to access content manifest, this might be simplified, because containerImage already contains `content_manifest_components` edge returning associated components. 
Also please keep in mid that this edge (and all edges) is not returning all components but paginated result, that you need to iterate to get all associated components. Parametrization of edges is same as parametrization of regular queries in this case; ...content_manifest_components(page: 0, page_size: 20)...

mmalina: Thanks for the feedback. I refactored it - now there is just one level of `edges` and pagination is dealt with (a `while` loop) and `bom_ref` is included as well, so I could get rid of the additional components loading.

mmalina:
> so based on this PR , I think I'll need to update in release e2e-tests suite [here](https://github.com/redhat-appstudio/e2e-tests/blob/f2f01144083a9348a74dfb3f1e0853f6b5cb9a3e/pkg/utils/release/sbom.go#L86) only ?

@kasemAlem yeah, I guess that should be removed. But it's not really used or affecting anything, right?

pyxis/upload_sbom.py
```diff
@@ -137,10 +148,21 @@ def get_image(graphql_api: str, image_id: str) -> dict:
...
+    while has_more:
+        variables = {"id": image_id, "page": page, "page_size": page_size}
+        body = {"query": query, "variables": variables}
 
-    image = pyxis.graphql_query(graphql_api, body, "get_image")
+        image = pyxis.graphql_query(graphql_api, body, "get_image")
+
+        components_batch = image["edges"]["content_manifest_components"]["data"]
+        components.extend(components_batch)
+        has_more = len(components_batch) == page_size
```
ezopezo, line 163: I believe this can ommit last page because  it is not guaranteed that count of components will be exactly page_size. I would with has_more = len(components_batch) only

ezopezo: Yes, that makes a prefect sense, thanks for explanation!

johnbieren (approved):

mmalina:
@johnbieren please disregard my re-request for review. I clicked it and then noticed you approved 1 hour ago :) Sorry about that. And thanks!

PR #44, opened 2023-07-18
fix: remove fetching of unreachable CA

theflockers:
Red Hat IT Root CA is unreachable from actions so this PR removes the fetching
of it to unblock image builds until we find a permanent solution.


theflockers:
done @mmalina. Please take a look.

theflockers:
@mmalina I think we should go for this one https://github.com/redhat-appstudio/release-service-utils/pull/45 what do you think?

PR #45, opened 2023-07-18
fix: add CA files to Dockerfile

theflockers:
Gitlab tool requires the RedHat IT Root public CAs to enable secure
connection to internal gitlab endpoint.

This PR adds the two required CA files to build the release-base-image.


theflockers:
Requires new secret `REDHAT_IT_ROOT_CA_BUNDLE`

mmalina:
This would make the setup more complex and it would no longer be easy to simply run `docker build .` locally without any prior setup. Personally I would prefer to simply copy the certs in the repo. I wonder what others thing.

theflockers:
PR changed to include the actual CA files instead of using the secret.

Dockerfile
```diff
@@ -21,6 +21,9 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     skopeo \
     && dnf clean all
 
+ADD pki/certs/2015-IT-Root-CA.pem pki/certs/2022-IT-Root-CA.pem /etc/pki/ca-trust/source/anchors/
```
mmalina, line 24: Can we use a different directory in the repo for this? Something to make it more obvious what it's for when you look at the repo root folder. I'd suggest `data/certs`, `data/pki/certs`. Or just `certs` as they do here: https://github.com/redhat-aqe/tox-container/tree/master/certs

theflockers: I would go for `data/certs` then. If we need to add anything in the future, having a`data` dir might be useful.

mmalina (approved):

PR #46, opened 2023-07-19
feat(HACBS-2271): improve reliability of upload_sbom script

mmalina:
This PR adds two improvements:

1. Print pyxis trace_id on graphql failure - this will help with debugging if the Pyxis request fails
2. Add retry option for upload_sbom script - this option will be used in the Tekton task and it will improve the reliability of the pipeline

### DIFF
--- a/pyxis/pyxis.py
+++ b/pyxis/pyxis.py
@@ -53,7 +53,7 @@ def _get_session(pyxis_url: str, auth_required: bool = True) -> requests.Session
     return session
 
 
-def post(url: str, body: Dict[str, Any]) -> Dict[str, Any]:
+def post(url: str, body: Dict[str, Any]) -> requests.Response:
     """POST pyxis API request to given URL with given payload
 
     Args:
@@ -76,7 +76,7 @@ def post(url: str, body: Dict[str, Any]) -> Dict[str, Any]:
             f"Pyxis POST query failed with {url} - {resp.status_code} - {resp.text}"
         )
         raise
-    return resp.json()
+    return resp
 
 
 def graphql_query(graphql_api: str, body: Dict[str, Any], query: str) -> Dict[str, Any]:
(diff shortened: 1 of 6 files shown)


kasemAlem:
reviewed it, 
I see you made good test coverage,  
just a question, how many retries  is implemented ? is it a defined value ?

pyxis/pyxis.py
```diff
@@ -76,7 +76,7 @@ def post(url: str, body: Dict[str, Any]) -> Dict[str, Any]:
             f"Pyxis POST query failed with {url} - {resp.status_code} - {resp.text}"
         )
         raise
-    return resp.json()
+    return resp
```
johnbieren, line 79: Little confused with this change. It seems you are just removing `.json()` here and adding it wherever the `post` function is called. What am I missing?

pyxis/upload_sbom.py
```diff
@@ -59,9 +60,35 @@ def parse_arguments() -> argparse.Namespace:  # pragma: no cover
     )
     parser.add_argument("--sbom-path", help="Path to the sbom file", required=True)
     parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
+    parser.add_argument(
+        "--retry",
+        "-r",
+        action="store_true",
+        help="Retry the whole sbom uploading in case it fails",
```
johnbieren, line 67: Maybe `number of times to retry uploading the entire sbom if it fails` Sounds like a boolean as written

mmalina: It is a boolean as implemented now. Do you think it would make more sense to have it as `int` instead (and rename to `--retries`) and default to `0`?

johnbieren: Oh, I looked too quickly and thought you were passing the # of retries. I think passing the number of retries you want makes the most sense. However, I am totally fine with keeping the code as is and just changing the help to be a little more clear... `Whether or not to retry the whole sbom uploading in case it fails` or something

pyxis/pyxis.py
johnbieren: ah, i missed that. blended in with the rest of the green. got it, thanks

johnbieren (approved):
I am good with whatever you decide on the retry thing

pyxis/upload_sbom.py
mmalina: OK, cool, I kept is as bool for now. I went with: `If set, retry the whole sbom uploading in case it fails`

mmalina: Oops, I forgot to push the help message change before merging. Here's a followup PR: https://github.com/redhat-appstudio/release-service-utils/pull/47

kasemAlem:
thanks @mmalina  for explanation . 
agree with Johnny

PR #49, opened 2023-07-24
fix: adds yamlpath and yamlline.py script to Dockerfile

theflockers:
Required by: https://github.com/hacbs-release/app-interface-deployments/pull/30/files

This PR installs the `yamlpath` utils and also the `yamlline.py` script to enable the task in the linked PR
to make changes in the input file using `sed` instead of `yq`, as the latter fully changes the input file due to a `go-yaml` issue.

### DIFF
--- a/utils/yamlline.py
+++ b/utils/yamlline.py
@@ -1,11 +1,8 @@
 #!/usr/bin/env python3
-"""Yamlline
+"""yamlline.py
 
-Simple helper script to return the actual line number of a yaml path in
-a yaml file.
+Script to return the line number of a yaml path in a yaml file.
 
-    * get_path_line_num - return the line number of the yaml path
-    * main - the main function of the script
 """
 
 import argparse
@@ -25,30 +22,35 @@ def get_path_line_num(yamlfile, yamlpath):
     try:
         with open(yamlfile, encoding="utf8") as file_name:
             found_root = False
-            # read the file line by line
             for line in iter(file_name.readline, ""):
                 # leave the loop and return the line number
-                # if the yaml path is found
+                # if there are no more strings to search
                 if len(search_path) == 0:
                     return line_number
 
                 try:
+                    # check if the search string is a root key of a yaml path
                     if re.search(f"^{search_path[0]}:", line):
                         search_path.pop(0)
                         found_root = True
-                        # return if it is the only key
+
+                        # return early if the path has a single key
                         if len(search_path) == 0:
                             line_number += 1
                             return line_number
 
+                    # if the root key was already found, search only for the
+                    # keys that are deeper
                     if found_root is True:
                         if re.search(f"\\s+.*{search_path[0]}:", line):
                             search_path.pop(0)
 
                 except re.error:
+                    # in case the key is a composite of [key=value] it should
+                    # rebuild the string as `key: val` and then search for it
                     rawkey = search_path[0][1:-1]
                     key, val = rawkey.split("=")
-                    search = f"{key}.*{val}$"
+                    search = f"{key}:.*{val}$"
                     if re.search(search, line):
                         search_path.pop(0)
 
@@ -71,7 +73,7 @@ def main():
     parser.add_argument(
         "-f", "--file", dest="filename", help="YAML file to read", metavar="FILE"
     )
-    parser.add_argument("-p", "--path", dest="path", help="YAML path to get the line number")
+    parser.add_argument("-p", "--path", dest="path", help="YAML path that we want to know the line number")
     args = parser.parse_args()
 
     if args.filename is None or args.path is None:


Dockerfile
```diff
@@ -21,6 +21,8 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     skopeo \
     && dnf clean all
 
+RUN pip3 install yamlpath
```
davidmogar, line 24: No available through dnf right?

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+"""Yamlline
+
+Simple helper script to return the actual line number of a yaml path in
+a yaml file.
+
+    * get_path_line_num - return the line number of the yaml path
```
johnbieren, line 7: I don't think these are worth adding here. They can just go as docstrings with their functions and that is fine IMO

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+"""Yamlline
+
+Simple helper script to return the actual line number of a yaml path in
```
johnbieren, line 4: I think `Script to return the line number of a yaml path in a yaml file` is sufficient

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+    * get_path_line_num - return the line number of the yaml path
+    * main - the main function of the script
+"""
+
+import argparse
+import re
+import sys
+
+
+def get_path_line_num(yamlfile, yamlpath):
```
johnbieren, line 16: can you add type hints?

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+                try:
+                    if re.search(f"^{search_path[0]}:", line):
+                        search_path.pop(0)
+                        found_root = True
+                        # return if it is the only key
+                        if len(search_path) == 0:
+                            line_number += 1
+                            return line_number
+
+                    if found_root is True:
```
johnbieren, line 44: I think this logic could use some more comments

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+    if len(search_path) == 1:
+        search_path = yamlpath.split("/")
+
+    line_number = 0
+    try:
+        with open(yamlfile, encoding="utf8") as file_name:
+            found_root = False
+            # read the file line by line
+            for line in iter(file_name.readline, ""):
+                # leave the loop and return the line number
```
johnbieren, line 30: what if it exists again further down in the yaml?

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def main():
+    """main function"""
+
+    parser = argparse.ArgumentParser(description="Parameters for yamlline.")
+    parser.add_argument(
+        "-f", "--file", dest="filename", help="YAML file to read", metavar="FILE"
+    )
+    parser.add_argument("-p", "--path", dest="path", help="YAML path to get the line number")
```
johnbieren, line 74: I am not following the help string on this one

Dockerfile
mmalina: Yep, seems that way:
```
# yum search yamlpath
Updating Subscription Management repositories.
Unable to read consumer identity

This system is not registered with an entitlement server. You can use subscription-manager to register.

Red Hat Universal Base Image 8 (RPMs) - BaseO 820 kB/s | 612 kB     00:00
Red Hat Universal Base Image 8 (RPMs) - AppSt 1.3 MB/s | 3.0 MB     00:02
Red Hat Universal Base Image 8 (RPMs) - CodeR  32 kB/s |  69 kB     00:02
No matches found.
```

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+
+import argparse
+import re
+import sys
+
+
+def get_path_line_num(yamlfile, yamlpath):
+    """Returns the line number of the given yaml path in a yaml file"""
+
+    # try split the path using ".", otherwise tries "/"
```
mmalina, line 19: What's the use case for using `/` instead of `.`? Why would that happen? Also, what if there is a `.` in a key name? Like this:
```yaml
my.test:
  - something
```

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+    search_path = yamlpath.split(".")
+    if len(search_path) == 1:
+        search_path = yamlpath.split("/")
+
+    line_number = 0
+    try:
+        with open(yamlfile, encoding="utf8") as file_name:
```
mmalina, line 26: `file_name` is not a good name - it's a handle of the open file, it's not the file name. I'd suggest just `file` or `f`.

utils/yamlline.py
theflockers: if the `search_path` length is 0 it means that there is no more strings to search. The minimum strings for search_path should be 1.

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+            found_root = False
+            # read the file line by line
+            for line in iter(file_name.readline, ""):
```
mmalina, line 29: Why not just `for line in file`?

utils/yamlline.py
theflockers: indeed. I will improve it.

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+                # leave the loop and return the line number
+                # if the yaml path is found
+                if len(search_path) == 0:
+                    return line_number
+
+                try:
+                    if re.search(f"^{search_path[0]}:", line):
```
mmalina, line 36: This will only check if the current line starts with `search_path[0]`, right? Why not use `line.startswith()`? But also, what if the yaml doesn't even keys at the root level? E.g.:
```yaml
- one
- two
```

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+                        search_path.pop(0)
+                        found_root = True
+                        # return if it is the only key
+                        if len(search_path) == 0:
+                            line_number += 1
+                            return line_number
+
+                    if found_root is True:
+                        if re.search(f"\\s+.*{search_path[0]}:", line):
```
mmalina, line 45: This is very unreliable. This assumes that the next time we encounter the next key in the file, it's the line we're looking for. But it may not be the case. Let's consider a file like this:
```yaml
rootkey:
  one:
    two: "something"
  two:
    one: "some other thing"
```
We have a path of `"rootkey.two.one"`, but the script will find the first "two" and `pop()` it from the path and carry on. 

Also, how will this work with lists, e.g. when the current `search_path[0]` is something like `"key[5]"`?

And we haven't touched the possibility that the yaml file contains inline json snippets (json is completely valid yaml).

To be honest I don't think this approach is a good idea in general.

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+                            search_path.pop(0)
+
+                except re.error:
```
mmalina, line 48: When will this happen? What's the scenario when there will be an error? And what happens in that case? I don't understand what the code does.

utils/yamlline.py
```diff
@@ -0,0 +1,86 @@
...
+                    rawkey = search_path[0][1:-1]
+                    key, val = rawkey.split("=")
+                    search = f"{key}.*{val}$"
+                    if re.search(search, line):
+                        search_path.pop(0)
+
+                except ValueError:
```
mmalina, line 55: When will this happen? What's the scenario when there's a `ValueError`?

utils/yamlline.py
theflockers: Nice catch. Also noticed some keys in this format in k8s yamls `deployment/container-port: xxx`

mmalina:
BTW, `yq` can actually print the line number of a selected field for you:
```
$ cat test.yaml | yq '(.resourceTemplates[0].targets[0].parameters.HYPERSHIFT_OPERATOR_OVERRIDING_DATA) '
  # openshift/hypershift.git branch release-4.13 commit 3d04abc15dc669129902979f8d9d97a8f269f5c7 / tag v0.1.6
  hypershift-operator: registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-hypershift-rhel8-operator@sha256:b287cb560cb8c1711176d292918ea04a20641730dd3a0e7b4f4fba8697ccd841
  apiserver-network-proxy: registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-apiserver-network-proxy-rhel8@sha256:281c31c439d4116a7180519370955254d1845e57fb0c2b9b55a84e4798bec9ce
  cluster-api-provider-agent: registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-cluster-api-provider-agent-rhel8@sha256:7dd4c639450284a8a4893fad89b49183a244deeaf05bc25370446316593a51e3

$ cat test.yaml | yq '(.resourceTemplates[0].targets[0].parameters.HYPERSHIFT_OPERATOR_OVERRIDING_DATA) | line'
10
```

theflockers:
Closed as `yq` can give the approximate line number.

PR #50, opened 2023-07-25
feat(HACBS-2225): add script to create InternalRequests

davidmogar:
This script allows to create new InternalRequests and wait for them to finish.

### DIFF
--- a/utils/internal-request
+++ b/utils/internal-request
@@ -1,0 +1,136 @@
+#!/bin/bash
+
+# This script creates an InternalRequest resource in a Kubernetes cluster
+# using the 'kubectl' command line tool. The resource is created with
+# parameters passed to the script, and optionally waits for the resource to
+# reach a 'completed' status.
+#
+# Usage:
+#   ./internal-request.sh -r request [-p <key=value> ...] [-s sync] [-t timeout]
+#
+# Parameters:
+#   -r  Request: the name of the request.
+#   -p  Parameters: can be specified multiple times. Each '-p' flag represents a
+#       parameter that will be added to the 'parameters' field in the
+#       InternalRequest resource. The value of the parameter is treated as a string,
+#       and it can be a valid JSON object or array. When passing complex parameter
+#       values, make sure to enclose them in quotes.
+#   -s  Sync: a flag that indicates whether the script has to finish to complete
+#       the tasks or exit immediately after creating the resource. Default is true.
+#   -t  Timeout: Defaults to 600 seconds.
+#   -h  Display this help message.
+#
+# Prerequisites:
+#   - kubectl: The Kubernetes command line tool must be installed and properly
+#     configured to communicate with your cluster.
+#   - jq: This script uses jq to parse JSON. It must be installed on the system
+#     running the script.
+#
+# Note: This script is intended to be used with a specific Kubernetes API
+# that includes the 'InternalRequest' resource type.
+
+set -e
+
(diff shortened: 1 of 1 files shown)


utils/internal-request
```diff
@@ -0,0 +1,122 @@
...
+# Check if mandatory parameters are set
+if [ -z "$REQUEST" ]
+then
+    usage
+fi
+
+# Convert parameters to JSON format
+IFS=","
+PARAM_JSON=""
+for param in $PARAMS
```
theflockers, line 68: For fileUpdates the VALUE will be a json string that contains several commas on it. I tested locally and it splits the json value as it was different params.  I tested it like:

`./internal-request  -r something -p "fileUpdates=<the escaped json string>"

It did not like it.

davidmogar: Jsons have to be passed as `-p 'fileUpdates={\"foo\": \"bar\"}'`.

utils/internal-request
```diff
@@ -0,0 +1,122 @@
...
+
+TIMEOUT=600
+SYNC=true
+PARAMS=""
+
+function usage {
+    echo "Usage: $0 -r request [-p parameters] [-s sync] [-t timeout]"
+    echo
+    echo "  -r  Request: the name of the request."
+    echo "  -p  Params: list of parameters to pass to the request (key, value). Default is empty."
```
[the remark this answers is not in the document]
davidmogar: Yeah, good feedback. I'll change it.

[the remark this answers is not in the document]
theflockers: The new way with multiple `-p` works fine, but the `usage` still needs to be updated.

utils/internal-request
```diff
@@ -0,0 +1,135 @@
...
+      },
+      "spec": {
+        "request": $request,
+        "parameters": $parameters
+      }
+    }'
+)
+
+echo "$PAYLOAD"
+exit 0
```
theflockers, line 103: exiting earlier.

utils/internal-request
```diff
@@ -0,0 +1,135 @@
...
+    --argjson parameters "$PARAM_JSON" \
+    '{
+      "apiVersion": "v1",
+      "kind": "InternalRequest",
+      "metadata": {
+        "generateName": ($request + "-")
+      },
+      "spec": {
+        "request": $request,
+        "parameters": $parameters
```
theflockers, line 97: should be `spec.params`

utils/internal-request
```diff
@@ -0,0 +1,135 @@
...
+      }
+    }'
+)
+
+echo "$PAYLOAD"
+exit 0
+
+# Create InternalRequest using kubectl
+RESOURCE=$(echo "$PAYLOAD" | kubectl create -f - -o json)
+INTERNAL_REQUEST_NAME=$(echo "$RESOURCE" | jq -r '.metadata.name')
```
theflockers, line 107: It would be nice to print the internalrequest name after create it. It is handy  when running in the command line.

utils/internal-request
davidmogar: Replace... Fixed.

utils/internal-request
davidmogar: I just pushed my test code. Fixed.

utils/internal-request
davidmogar: Done.

theflockers (approved):
LGTM

PR #51, opened 2023-07-28
chore: use git shallow clone

theflockers:
Use git shallow clone to save disk space and clone time.


mmalina:
@theflockers the change looks good. But can you add at least a brief explanation to the commit message? Like:

> This will make the clone much smaller and the clone is only used for creating a MR/PR, so history is not needed.

Otherwise it LGTM.

mmalina (approved):

PR #52, opened 2023-08-10
feat(HACBS-2405): add a setup script for tekton task testing

mmalina:
The new tekton dir will be used for storing init scripts for tekton task testing in the release-service-bundles repo.

This change is needed by
https://github.com/redhat-appstudio/release-service-bundles/pull/159

### DIFF
--- a/tekton/test-push-sbom-to-pyxis-no-snapshot
+++ b/tekton/test-push-sbom-to-pyxis-no-snapshot
@@ -1,0 +1,16 @@
+# Description: This script is meant to set up mocks at the beginning
+#              of each step of the push-sbom-to-pyxis task.
+
+function cosign() {
+  echo Mock cosign called with: $*
+
+  echo Error: cosign should not be called
+  exit 1
+}
+
+function upload_sbom() {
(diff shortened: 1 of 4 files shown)


tekton/test-push-sbom-to-pyxis-no-snapshot
```diff
@@ -0,0 +1,16 @@
+# Description: This script is meant to set up mocks at the beginning
+#              of each step of the push-sbom-to-pyxis task.
+
+function cosign() {
```
mmalina, line 4: For these negative tests that are expected to fail even before the mocked commands are called it doesn't really provide much value to have this at all - because even if the mock was unexpectedly called, the task would still fail which is what the test expects anyway (in this case it tests that the task fails if there is no snapshot).

It would only really make sense to have these if we were able to have more granular checks of how exactly the task failed.

But anyway, this is to show what you can do to init different test scenarios.

mmalina:
Closing this. We decided to go with Solution 2: https://github.com/redhat-appstudio/release-service-bundles/pull/162

PR #53, opened 2023-08-15
feat(HACBS-2425): add jq and krb5-workstation to dockerfile

Troy876:
Update Dockerfile to include what's missing from the bundles Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,14 +3,16 @@ FROM registry.access.redhat.com/ubi8/ubi
 ARG COSIGN_VERSION=2.1.1
 ARG KUBECTL_VERSION=1.27.2
 ARG OCP_VERSION=4.13.3
+ARG JQ_VERSION=1.6
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
 
-RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -6,11 +6,12 @@ ARG OCP_VERSION=4.13.3
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
 
-RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
+RUN curl -L https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -o /usr/bin/jq &&\
```
johnbieren, line 9: Can you make the jq version an `ARG` like we have for the other tools?

johnbieren (changes requested):
The commit isn't the proper format. It should be something like 
`feat(HACBS-2425): add jq and krb5-workstation to dockerfile`

Dockerfile
Troy876: ARG JQ_VERSION=1.6 has been added to the Dockerfile

johnbieren: Can you use the variable in the url?

happybhati: +1 Johnny
`RUN curl -L https://github.com/stedolan/jq/releases/download/jq-${JQ_VERSION}/jq-linux64 -o /usr/bin/jq &&\`

johnbieren (approved):

Dockerfile
Troy876: Should be updated now

PR #54, opened 2023-08-21
feat(HACBS-2227): snapshot data model can be used in update expressions

theflockers:
This PR adds the update-paths script to enable updating
yq expression in fileUpdates paths strings with the result
of these expressions

### DIFF
--- a/utils/update-paths
+++ b/utils/update-paths
@@ -1,0 +1,68 @@
+#!/bin/bash
+#
+# script:      update-paths
+# 
+# description: This script searches for "{{ }}" enclosed yq expression in fileUpdates paths then
+#              executes and replaces it with its result
+#
+# parameters:  
+#
+# -p, --paths  Json String containing fileUpdates paths with optional yq expression(s) enclosed by "{{ }}"
+#              The `paths` json string should be properly quoted
+#              Example:  --paths='[{"path":"foo.yaml","replacements":[{"key":"bar","replacement":"{{ .components[0]| .containerImage }}"}]}]'
+#
+# -f, --file:  Path to the mapping_snapshots.json file
+#
+print_help(){
+    echo -e "$0 [ -p, --paths ] PATHS [ -f, --file ] FILE\n"
+    echo -e "\t-p, --paths\tJson String containing fileUpdates paths with optional yq expression(s) enclosed by \"\${{ }}\""
+    echo -e "\t\t\tThe \`paths\` json string should be properly escaped/quoted"
+    echo -e "\t\t\tExample:  --paths='[{\"path\":\"foo.yaml\",\"replacements\":[{\"key\":\"bar\",\"replacement\":\"{{ .components[0]| .containerImage }}\"}]}]'"
+    echo -e "\t-f, --file\tPath to mapping_snapshots.json file\n"
+}
+
+OPTIONS=$(getopt -l "paths:,file:" -o "p:f:" -a -- "$@")
+eval set -- "$OPTIONS"
+while true; do
+    case "$1" in
+        -p|--paths)
+            shift
+            PATHS=$1
+            ;;
+        -f|--file)
+            shift
+            FILE="$1"
+            ;;
+        --)
+            shift
+            break
+            ;;
+    esac
+    shift
+done
+
+if [ -z "${PATHS}" ] || [ -z "${FILE}" ]; then
+    print_help
+    exit
+fi
+
+pathsLength=$(jq -cr '. |length' <<< "${PATHS}")
+for(( p=0; p<pathsLength; p++)); do
+
+  path=$(jq ".[${p}]" <<< "${PATHS}")
+  replacementsLength=$(jq '.replacements | length' <<< "${path}")
+  for(( r=0; r<replacementsLength; r++ )); do
+      replacement=$(jq -cr ".replacements[${r}].replacement" <<< "${path}")
+      if grep -q '{{.*}}' <<< "${replacement}"; then
+          yqExp=$(sed 's|.*{{\(.*\)}}.*|\1|' <<< "${replacement}")
+
+          # execute the yq expression and replace the "replacement string" with the result
+          yqResult=$(yq -oy -r "${yqExp}" "${FILE}")
+          newReplacement=$(sed "s|{{.*}}|"${yqResult}"|g" <<< "${replacement}")
+
+          newReplacementJson=$(jq -R <<< ${newReplacement})
+          PATHS=$(jq -cr ".[${p}].replacements[$r].replacement = ${newReplacementJson}" <<< "${PATHS}")
+      fi
+  done
+done
+echo "${PATHS}"


utils/update-paths
```diff
@@ -0,0 +1,86 @@
+#!/bin/bash
+#
+# script:      update-paths
+# 
+# description: This script searches for "${{ }}" enclosed yq expression in fileUpdates paths then
+#              executes and replaces it with its result
+#
+# parameters:  
+#
+# -p, --paths: Json String container fileUpdates paths
```
johnbieren, line 10: I am having trouble understanding what this is saying

utils/update-paths
```diff
@@ -0,0 +1,86 @@
...
+fi
+
+newPaths=('{"paths": []}')
+
+IFS=$'\n' # This is required because the JSON string might contain spaces
+for path in $(jq -cr '.[]' <<< "${PATHS}"); do
+  newPath=$(jq '{"path": .path}' <<< "${path}")
+  newReplacements=('{"replacements": []}')
+
+  replacementsLenght=$(jq '.replacements | length' <<< "${path}")
```
johnbieren, line 54: Lenght, not Lenght

utils/update-paths
theflockers: it refers to this - but it is written wrong - should be "_containing_":

```
            paths:
                - path: deploy2.yaml
                  replacements:
                      - key: "(.resourceTemplates[] | select(.name==\"osd-fleet-manager-production\").targets[] | select(.name==\"osd-fleet-manager-production-us-east-2\") | .parameters.HYPERSHIFT_OPERATOR_OVERRIDING_DATA)"
                        replacement: "|cluster-api-provider-agent.*|cluster-api-provider-agent: SOME_IMAGE|"
                      - key: "(.resourceTemplates[] | select(.name==\"osd-fleet-manager-production\").targets[] | select(.name==\"osd-fleet-manager-production-blue\") | .parameters.HYPERSHIFT_OPERATOR_OVERRIDING_DATA)"
                        replacement: "|registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-cluster-api-provider-kubevirt-rhel8@sha256:8d7f489b7040e2c528e51e46b290074fc161a42cdf3120c2cc0544b3ca97dde9|${{'.components[] | select(.name==\"work\") | .containerImage'}}|"
                      - key: "(.resourceTemplates[] | select(.name==\"osd-fleet-manager-production\").targets[] | select(.name==\"osd-fleet-manager-production-canary\") | .parameters.CLUSTER_TEMPLATES)"
                        replacement: "|effect.*NoSchedule|effect: NoExecute|"
```
                        
I will try to make it more clear.

utils/update-paths
theflockers: Caught you.  You also wrote it wrong :D

utils/update-paths
johnbieren: I think putting an example (as simple/minimal) as possible in the description would be nice. The help says it is a json but you just pasted yaml, so I am still a little unclear

johnbieren (commented):
Other than the paths thing, it lgtm. I didn't test the bash itself, I am assuming you did and it does what you want

utils/update-paths
theflockers: Added a snippet in the header of the file and as well in the help.

johnbieren (approved):
I'm good with this. I think it would be nice to get another team member to review it as well, but lgtm

utils/update-paths
```diff
@@ -0,0 +1,88 @@
...
+IFS=$'\n' # This is required because the JSON string might contain spaces
+for path in $(jq -cr '.[]' <<< "${PATHS}"); do
+  newPath=$(jq '{"path": .path}' <<< "${path}")
+  newReplacements=('{"replacements": []}')
+
+  replacementsLength=$(jq '.replacements | length' <<< "${path}")
+  for(( i=0; i<replacementsLength; i++ )); do
+      replacement=$(jq -cr ".replacements[${i}]" <<< "${path}")
+      # Here we are searching for the enclosed by "${{ }}" yq exp to later execute it.
+      # The PADDING = 4 means the begining of the expression "${{" +1 (the "'" char)
```
mmalina, line 60: What's the `'` in the expression needed for? Maybe it can be removed? I don't really understand how the awk command works exactly. But it seems like something that could potentially break.

Here's another idea: What if we supported variables like `$1` in the replacement string and there would be an array with the values, e.g.: `"vars": [".components[0].containerImage", "".components[1].containerImage"]` The advantage would be that it would be much simpler to find the variable references in the `replacement` string (we would support up to 9 of them). Just a thought.

utils/update-paths
```diff
@@ -0,0 +1,88 @@
...
+
+          EXP = substr($0, RSTART + PADDING, RLENGTH - (PADDING*2)+1)
+      } { 
+          gsub(/\\/, "", EXP); print EXP
+      }' <<< "${replacement}")
+
+      if [ -n "${yqExp}" ]; then
+          # execute the yq expression and replace the "replacement string" with the result
+          yqResult=$(yq "${yqExp}" "${FILE}")
+          newReplacement=$(sed "s|\${{.*}}|"${yqResult}"|g" <<< "${replacement}")
```
mmalina, line 74: If there is more than one yq expression, this will remove the section from the beginning of the first one to the end of the last one, no? (by default, regular expressions are greedy)

utils/update-paths
theflockers: About the `'` -  the expression should be passed like this: `${{'.components[] | select(.name==\"work\") | .containerImage'}}` - They are not required. Just thought the expression would look "loose" inside the curly braces. 

About the second phrase I did not get what you mean. Could you elaborate a bit more? Maybe giving some more graphic example.

utils/update-paths
theflockers: It should be one yq expression per replacement. 
You can do multiple replacements in per (path) file, but it should be in another key/replacement pair.

```
- repo: some.git
  path: foobar.yaml
    replacements:
      - key: "somekey"
        replacement: "|foo|${{'.containerImage'}}|" #  this is ok
      - key: "somekey2"
        replacement: "|bar|${{'.containerImage'}}|" # this is ok too
      - key: "somekey3"
        replacement: "|jeez|${{'.containerImage'}} ${{'.containerImage'}}|"" # <= this is not ok and will be documented.
```

utils/update-paths
```diff
@@ -0,0 +1,88 @@
...
+#              The `paths` json string should be properly quoted
+#              Example: --paths="[{\"path\":\"foo.yaml\",\"replacements\":[{\"key\":\"bar\",\"replacement\":\"\${{'.components[0].containerImage'}}\"}]}]"
+#
+# -f, --file:  Path to the mapping_snapshots.json file
+#
+print_help(){
+    echo -e "$0 [ -p, --paths ] PATHS [ -f, --file ] FILE\n
+            \t-p, --paths\tJson String containing fileUpdates paths with optional yq expression(s) enclosed by \"\${{ }}\"
+            \t\t\tThe \`paths\` json string should be properly escaped/quoted
+            \t\t\tExample: --paths=\"[{\\\"path\\\":\\\"foo.yaml\\\",\\\"replacements\\\":[{\\\"key\\\":\\\"bar\\\",\\\"replacement\\\":\\\"\\\${{'.components[0].containerImage'}}\\\"}]}]\"\n
```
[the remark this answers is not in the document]
mmalina: But here you have even more backslashes than in the actual example 8 lines above. Can't we at least eliminate those? Here the only thing that matters is what's printed, no? Sorry, but I just find the amount of escaping here insane :)

utils/update-paths
mmalina: What I had in mind is something like:

```yaml
- repo: some.git
  path: foobar.yaml
    replacements:
      - key: "somekey"
        replacement: "|foo|$1 abc $2|"
        vars: [".containerImage[0]", ".containerImage[1]"]
```

Because $1 is much easier to replace. But now that you say there will be just one at most, it's probably ok. What seems really funky in the awk command is the padding. Why not just match exactly what we want using a capture group?

Apparently original awk cannot do that. gawk can (which is what you rely on anyway likely). But perl can do it as well.

```sh
echo '${{ .my_component }}'| perl -n -e'/\$\{\{(.*)\}\}/ && print $1'
.my_component
```

Or you can also easily do it with `sed`:
```sh
echo 'bla ${{ .my_component }} bla'| sed 's/.*\${{\(.*\)}}.*/\1/'
 .my_component
```

I would really omit the `' '` as having it just makes things more complicated with the matching.

On top of this, just a tiny suggestion: I would propose to use just `{{ }}` instead of `${{ }}`. It doesn't really matter, but this is also what jinja uses, so it looks more familiar. But yours is ok as well.

utils/update-paths
mmalina: OK, see my other comment. If only one `yq` expression is allowed, that makes things easier.

utils/update-paths
```diff
@@ -0,0 +1,88 @@
...
+done
+
+if [ -z "${PATHS}" ] || [ -z "${FILE}" ]; then
+    print_help
+    exit
+fi
+
+newPaths=('{"paths": []}')
+
+IFS=$'\n' # This is required because the JSON string might contain spaces
```
mmalina, line 51: If it's only needed for that one for on the following line, just define it there (at the beginning of the line). But I would actually propose to use indexes, just like you do in the next one. It's much safer.

utils/update-paths
```diff
@@ -0,0 +1,88 @@
...
+          newReplacements[0]=$(jq -cr ".replacements += [${newReplacement}]" <<< "${newReplacements[0]}")
+      else
+          newReplacements[0]=$(jq -cr ".replacements += [${replacement}]" <<< "${newReplacements[0]}")
+      fi
+
+  done
+
+  # merge the two hashes and append to the newPaths array
+  merged=$(jq -cr -s add  <<< "${newPath[0]} ${newReplacements[0]}")
+  newPaths[0]=$(jq -cr ".paths += [ ${merged} ]" <<< "${newPaths[0]}")
```
mmalina, line 85: Why is `newPaths` defined as an array if we only ever use the first item in that array?

mmalina: BTW, if you use indexing for both for loops as I suggest elsewhere, then you can always work with the original json string and you can replace each replacement directly in place and there is no need for any merging or anything like that.

theflockers: I used _shellcheck_ to check the script. It gave an advice that I thought reasoneable  to use array for the `newPaths` and others variables.

Some example: 
```
$ cat test.sh
#!/bin/bash

test='{ "abc": "test" }'

echo $test
```

```
$ shellcheck test.sh
In test.sh line 3:
test='{ "abc": "test" }'
     ^-----------------^ SC2089: Quotes/backslashes will be treated literally. Use an array.
```

mmalina: But ... that doesn't seem to apply to our use case at all. It says that if you intend to actually use the original string as separate space separated arguments, you should use an array. First of all, that's not what we want to do here at all - we want to keep the string as is, including all the quotes (otherwise the json is invalid). And second, wrapping the whole string in an array containing just one item doesn't do anything at all - the idea in that example is to have each "argument" in the original string stored as a separate item in the array.

utils/update-paths
theflockers: I removed the single quotes and also removed the `awk` in favor of the `sed`version. Also a small change in the logic as the `sed` one keeps the full string when it does not match.

I kept the it without the `vars` so far as it is just a single one.

utils/update-paths
theflockers: changed to use indexes.

utils/update-paths
theflockers: ok. removed the arrays.

theflockers:
PTAL @mmalina

utils/update-paths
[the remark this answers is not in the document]
mmalina: Much better, thanks!

utils/update-paths
mmalina: Much simpler, cool.

utils/update-paths
```diff
@@ -0,0 +1,79 @@
...
+  path=$(jq ".[${p}]" <<< "${PATHS}")
+  newPath=$(jq '{"path": .path}' <<< "${path}")
+
+  newReplacements='{"replacements": []}'
+  replacementsLength=$(jq '.replacements | length' <<< "${path}")
+  for(( r=0; r<replacementsLength; r++ )); do
+
+      replacement=$(jq -cr ".replacements[${r}]" <<< "${path}")
+      if grep -q '{{.*}}' <<< "${replacement}"; then
+          # get the yq expression and use xargs to remove the trailling whitespaces
```
mmalina, line 62: What whitespace is generated here and why does it matter? If it's the white space surrounding the actual yq expression inside `{{}}`, then you could improve the regexp to also eat it. But I can't see why it could be a problem - I'm sure the `yq` tool doesn't care if there are a few white spaces here and there.

utils/update-paths
mmalina: Cool.

utils/update-paths
```diff
@@ -0,0 +1,79 @@
...
+          yqExp=$(sed 's|.*{{\(.*\)}}.*|\1|' <<< "${replacement}"| xargs)
+
+          # execute the yq expression and replace the "replacement string" with the result
+          yqResult=$(yq -oy -r "${yqExp}" "${FILE}")
+          newReplacement=$(sed "s|{{.*}}|"${yqResult}"|g" <<< "${replacement}")
```
mmalina, line 67: Here you can just do:
```sh
PATHS=$(jq -cr ".[${p}].replacements[$r].replacement = \"$newReplacement\"")
```
or
```sh
PATHS=$(jq -cr ".[${p}].replacements[$r] = \"$newReplacement\"")
```
if you work with the whole replacement (including the key).
And then no merging is needed at all - at the end you will have all the updates in `PATHS` directly.

utils/update-paths
```diff
@@ -0,0 +1,79 @@
...
+for(( p=0; p<pathsLength; p++)); do
+
+  path=$(jq ".[${p}]" <<< "${PATHS}")
+  newPath=$(jq '{"path": .path}' <<< "${path}")
+
+  newReplacements='{"replacements": []}'
+  replacementsLength=$(jq '.replacements | length' <<< "${path}")
+  for(( r=0; r<replacementsLength; r++ )); do
+
+      replacement=$(jq -cr ".replacements[${r}]" <<< "${path}")
```
mmalina, line 60: Shouldn't we be working with just the replacement, and not the {key, replacement} pair?

utils/update-paths
theflockers: Bash added `\ `  to the space and this broke the yq. The trim avoided that to happen.

utils/update-paths
theflockers: Initially that was the idea, but the `yq` problems discussed in the other PR made us need to use yq to locate and sed to replace. So using key, replacement made it easier.

mmalina: I'm not sure you understand what I mean. In context of this script, you don't do anything about the key. But the lookup for `{{}}` should only apply to the replacement.replacement bit, not the whole `{ "key": ".some.key", "replacement": "my {{ .expression[1] }}"}`.

utils/update-paths
mmalina: Bash added `\`? Where? How? Do you have an example? I'd like to see it for myself.

utils/update-paths
theflockers: I still need the key to proper build the replacements later in the script.
```
 70           newReplacements=$(jq -cr ".replacements += [${newReplacement}]" <<< "${newReplacements}")                                                                                                                                                                                                                       
 71       else                                                                                          
 72           newReplacements=$(jq -cr ".replacements += [${replacement}]" <<< "${newReplacements}")    
```

mmalina: I don't follow. I'm saying that the yq sed replacement should only be applied to the `replacement` and not the `key`. Of course the `key` part is still there in the json.

utils/update-paths
theflockers: Different from what I said before, it not even added the `\`. It just broke yq:
```
++ jq -cr '.replacements[0].replacement'
+ replacement='|registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-cluster-api-provider-kubevirt-rhel8@sha256:8d7f489b7040e2c528e51e46b290074fc161a42cdf3120c2cc0544b3ca97dde9|{{ .components[] | select(.name=="devfile-sample-python-basic") | .containerImage }}|'
+ grep -q '{{.*}}'
++ sed 's|.*{{\(.*\)}}.*|\1|'
+ yqExp=' .components[] | select(.name=="devfile-sample-python-basic") | .containerImage '
++ yq -oy -r ' .components[] | select(.name=="devfile-sample-python-basic") | .containerImage ' /workspace/data/snapshot_spec.json
+ yqResult=quay.io/hacbs-release-tests/managed-release-team-tenant/simple-python/devfile-sample-python-basic@sha256:085ec47e4b1c07ee5a9f947c6964b8dfaed3659b6a748523ddf650b2d95eeeb6
++ sed 's|{{.*}}|quay.io/hacbs-release-tests/managed-release-team-tenant/simple-python/devfile-sample-python-basic@sha256:085ec47e4b1c07ee5a9f947c6964b8dfaed3659b6a748523ddf650b2d95eeeb6|g'
+ newReplacement='|registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-cluster-api-provider-kubevirt-rhel8@sha256:8d7f489b7040e2c528e51e46b290074fc161a42cdf3120c2cc0544b3ca97dde9|quay.io/hacbs-release-tests/managed-release-team-tenant/simple-python/devfile-sample-python-basic@sha256:085ec47e4b1c07ee5a9f947c6964b8dfaed3659b6a748523ddf650b2d95eeeb6|'
++ jq -cr '.replacements += [|registry-proxy.engineering.redhat.com/rh-osbs/multicluster-engine-cluster-api-provider-kubevirt-rhel8@sha256:8d7f489b7040e2c528e51e46b290074fc161a42cdf3120c2cc0544b3ca97dde9|quay.io/hacbs-release-tests/managed-release-team-tenant/simple-python/devfile-sample-python-basic@sha256:085ec47e4b1c07ee5a9f947c6964b8dfaed3659b6a748523ddf650b2d95eeeb6|]'
jq: error: syntax error, unexpected '|' (Unix shell quoting issues?) at <top-level>, line 1:
```

mmalina: I would need to see the input and the exact version of your script to debug. But it seems you changed the script to only work with the `replacement` bit, not `{"key": ..., "replacement"}`, but you're then trying to add just that `replacement` bit as an item to the whole `.replacements`. That won't work.

theflockers: I delete the comment above because the error was unrelated. The actual problem (removing the `xargs` is this one:

```
+ yqExp=' .components[] | select(.name==\"devfile-sample-python-basic\") | .containerImage '
++ yq -oy -r ' .components[] | select(.name==\"devfile-sample-python-basic\") | .containerImage ' /workspace/data/snapshot_spec.json
Error: 1:32: invalid input text "\\\"devfile-sample..."
```
That "\" should not be there. To be very honest, I already change this so much that I can't remember all reasons I changed this or that.

theflockers: changes done as discussed in slack

utils/update-paths
theflockers: changes done as discussed in slack

utils/update-paths
```diff
@@ -0,0 +1,74 @@
...
+    esac
+    shift
+done
+
+if [ -z "${PATHS}" ] || [ -z "${FILE}" ]; then
+    print_help
+    exit
+fi
+
+newPaths='{"paths": []}'
```
mmalina, line 49: This is no longer used and should be removed.

utils/update-paths
```diff
@@ -0,0 +1,74 @@
...
+pathsLength=$(jq -cr '. |length' <<< "${PATHS}")
+for(( p=0; p<pathsLength; p++)); do
+
+  path=$(jq ".[${p}]" <<< "${PATHS}")
+  newPath=$(jq '{"path": .path}' <<< "${path}")
```
mmalina, line 54: This is no longer used and should be removed.

utils/update-paths
mmalina: ok, cool, so the comment on line 62 about xargs should be removed.

utils/update-paths
```diff
@@ -0,0 +1,74 @@
...
+
+  newReplacements='{"replacements": []}'
+  replacementsLength=$(jq '.replacements | length' <<< "${path}")
+  for(( r=0; r<replacementsLength; r++ )); do
+
+      replacement=$(jq -cr ".replacements[${r}].replacement" <<< "${path}" |jq -R )
```
mmalina, line 60: The second `jq` is redundant. It doesn't make sense to explicitly request raw output in the first `jq` and then run a second `jq` to convert back to json output. So instead, just remove the `-r` in the first one.

utils/update-paths
```diff
@@ -0,0 +1,74 @@
...
+      if grep -q '{{.*}}' <<< "${replacement}"; then
+          # get the yq expression and use xargs to remove the trailling whitespaces
+          yqExp=$(sed 's|.*{{\(.*\)}}.*|\1|' <<< "${replacement}")
+
+          # execute the yq expression and replace the "replacement string" with the result
+          yqResult=$(yq -oy -r "${yqExp}" "${FILE}")
+          newReplacement=$(sed "s|{{.*}}|"${yqResult}"|g" <<< "${replacement}")
+
+          # appends to the newReplacements array
+          PATHS=$(jq -cr ".[${p}].replacements[$r].replacement = ${newReplacement}" <<< "${PATHS}")
```
mmalina, line 70: `replacement` and by extension `newReplacement` as well are raw strings. That's what we wanted in order to be able to process the `{{}}` section with `yq`. But the issue now is that here a json string is expected.

With raw string I get this error:
```
jq: error: syntax error, unexpected '|' (Unix shell quoting issues?) at <top-level>, line 1:
.[0].replacements[0].replacement = |registry...
```

That's because when assigning a string to a field, the string needs to be a proper json string, including `""`. Of course we could just add those. But that is not safe, because there could be `"` characters inside the string (maybe not in our use cases, but still). So we'll need to convert the raw string back to json string first. Like this:

```
          newReplacementJson=$(jq -R <<< ${newReplacement})
          PATHS=$(jq -cr ".[${p}].replacements[$r].replacement = ${newReplacementJson}" <<< "${PATHS}")
```

utils/update-paths
```diff
@@ -0,0 +1,74 @@
...
+      replacement=$(jq -cr ".replacements[${r}].replacement" <<< "${path}" |jq -R )
+      if grep -q '{{.*}}' <<< "${replacement}"; then
+          # get the yq expression and use xargs to remove the trailling whitespaces
+          yqExp=$(sed 's|.*{{\(.*\)}}.*|\1|' <<< "${replacement}")
+
+          # execute the yq expression and replace the "replacement string" with the result
+          yqResult=$(yq -oy -r "${yqExp}" "${FILE}")
+          newReplacement=$(sed "s|{{.*}}|"${yqResult}"|g" <<< "${replacement}")
+
+          # appends to the newReplacements array
```
mmalina, line 69: This no longer applies and should be removed.

utils/update-paths
theflockers: adjusted as discussed in slack.

utils/update-paths
theflockers: removed.

utils/update-paths
theflockers: done.

utils/update-paths
theflockers: removed.

PR #56, opened 2023-09-06
feat(HACBS-2590): adds new parameters to gitlab_create_mr

theflockers:
This PR adds the new parameters `--upstream-repo` and `--target-branch` to the `gitlab_create_mr` function.

### DIFF
--- a/utils/gitlab-functions
+++ b/utils/gitlab-functions
@@ -17,13 +17,21 @@ gitlab_auth() {
 }
 
 gitlab_create_mr() {
-    OPTIONS=$(getopt -l "head:,title:,description:" -o "h:t:d:" -a -- "$@")
+    OPTIONS=$(getopt -l "head:,title:,target-branch:,description:,upstream-repo:" -o "h:t:b:d:u:" -a -- "$@")
     eval set -- "$OPTIONS"
     while true; do
         case "$1" in
             -h|--head)
                 shift
                 HEAD="$1";
+                ;;
+            -u|--upstream-repo)
+                shift
+                UPSTREAM_REPO="$1";
+                ;;
+            -b|--target-branch)
+                shift
+                TARGET_BRANCH="$1";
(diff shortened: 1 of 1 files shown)


utils/gitlab-functions
```diff
@@ -40,7 +48,9 @@ gitlab_create_mr() {
         esac
         shift
     done
-    glab mr create --title "${TITLE}" -b ${HEAD} --description "${DESCRIPTION}" | \
+    git remote add glab-base "${UPSTREAM_REPO}"
```
davidmogar, line 51: Is `glab` GitLab? If so, can we give it a better name? `gitlab-base` or just `upstream` in a more generic way. Clients can change their name :upside_down_face:

theflockers: It is a default name used by the tool. I had to define it otherwise it gives a prompt instead of just creating the MR.

utils/gitlab-functions
```diff
@@ -40,7 +48,9 @@ gitlab_create_mr() {
...
+    glab mr create --title "${TITLE}" --source-branch ${HEAD} --target-branch ${TARGET_BRANCH} \
+        --description "${DESCRIPTION}" -R "${UPSTREAM_REPO}" | \
```
johnbieren, line 53: Does this work? the option is lowercase r, here it is uppercase. I also kind of think it should be u not r, but

theflockers: it does.
```
$ glab mr  --help |grep repo
  -R, --repo OWNER/REPO   Select another repository using the OWNER/REPO or `GROUP/NAMESPACE/REPO` format or full URL or git URL
```

johnbieren: ah, I had the functions confused. Thanks!

johnbieren (approved):

happybhati (approved):

PR #57, opened 2023-09-08
fix: fork should be rebased with upstream branch before MR

theflockers:
This PR adds the git_rebase function to git-functions so tasks that uses it can rebase the local (a fork, in case of file-updates) with the upstream with a given branch.

Also some small code adjusts.

### DIFF
--- a/utils/git-functions
+++ b/utils/git-functions
@@ -11,8 +11,8 @@ git_functions_usage() {
 }
 
 git_config_ident() {
-    git config user.name "${GIT_AUTHOR_NAME}" 
-    git config user.email "${GIT_AUTHOR_EMAIL}"
+    git config --global user.name "${GIT_AUTHOR_NAME}"
+    git config --global user.email "${GIT_AUTHOR_EMAIL}"
 }
 
 git_commit_and_push() {
@@ -35,10 +35,9 @@ git_commit_and_push() {
         esac
         shift
     done
-    git checkout -b $BRANCH
-    git_config_ident
-    git commit -a -m "$COMMIT_MSG"
-    git push origin $BRANCH
+    git checkout -b "${BRANCH}"
+    git commit -a -m "${COMMIT_MSG}"
+    git push origin "${BRANCH}"
 }
 
 git_clone_and_checkout() {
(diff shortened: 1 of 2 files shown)


utils/git-functions
```diff
@@ -64,19 +63,56 @@ git_clone_and_checkout() {
...
     CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPOSITORY#*://}"
-    git clone --depth 1 $CLONE_URL
-    cd $REPO_DIR
-    git checkout $REVISION
+    git clone --depth 1 "${CLONE_URL}"
+    cd "${REPO_DIR}"
+    git checkout "${REVISION}"
+}
+
+git_rebase() {
```
scoheb, line 75: where is this called?

utils/git-functions
```diff
@@ -11,8 +11,8 @@ git_functions_usage() {
 }
 
 git_config_ident() {
-    git config user.name "${GIT_AUTHOR_NAME}" 
-    git config user.email "${GIT_AUTHOR_EMAIL}"
+    git config --global user.name "${GIT_AUTHOR_NAME}"
```
scoheb, line 14: why is this needed?

utils/git-functions
theflockers: in the task. I will create a draft PR to show it.

scoheb (commented):
some questions

utils/git-functions
theflockers: two reasons: gitlab asks for the User e-mai when creating the MR, so if it is not passed it gives a prompt.

The second is because `git` asks  for the user as well to run the `git fetch`. I moved to `--global` because in some places `git_config_ident` is called there is no repo cloned yet, to it fails.

utils/git-functions
theflockers: @scoheb  https://github.com/hacbs-release/app-interface-deployments/pull/46/files

scoheb (approved):

utils/git-functions
johnbieren: I almost think we should set up global git configs in the container itself, but this is probably fine for the time being

PR #58, opened 2023-09-12
fix(RHTAPBUGS-756): set digest in repo when creating an image in Pyxis

mmalina:
Based on the mediaType returned by `skopeo inspect --raw`, the create_container_image will determine if the image is single arch or multi arch and based on that
it will populate either manifest_schema2_digest
or manifest_list_digest in the embedded
ContainerImageRepo object when creating a new
image in Pyxis.

This also removes the docker_image_digest and
image_id fields when creating the image. This is to align with metaxor which does not populate these either.

Note that currently we do not really support multi arch images in RHTAP and other changes will likely be needed to fully support them.

Once this is merged, we will update the create-pyxis-image tekton task to supply the new parameter when calling this script: https://github.com/redhat-appstudio/release-service-bundles/pull/195

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -11,6 +11,12 @@ import pyxis
 
 LOGGER = logging.getLogger("create_container_image")
 
+# Media types that are used for multi arch images
+MANIFEST_LIST_TYPES = [
+    "application/vnd.oci.image.index.v1+json",
+    "application/vnd.docker.distribution.manifest.list.v2+json",
+]
+
 
 def setup_argparser() -> Any:  # pragma: no cover
     """Setup argument parser
@@ -42,6 +48,12 @@ def setup_argparser() -> Any:  # pragma: no cover
         help="Should the `latest` tag of the ContainerImage be overwritten?",
         required=True,
     )
+    parser.add_argument(
+        "--media-type",
+        help="The mediaType string returned by `skopeo inspect --raw`. "
+        "Used to determine if it's a single arch or multiarch image.",
+        required=True,
+    )
     parser.add_argument("--verbose", action="store_true", help="Verbose output")
     return parser
 
@@ -52,9 +64,10 @@ def image_already_exists(args, digest: str) -> bool:
 
     :return: True if one exists, else false
     """
+    digest_field = get_digest_field(args.media_type)
 
     # quote is needed to urlparse the quotation marks
-    filter_str = quote(f'docker_image_digest=="{digest}";' f"not(deleted==true)")
+    filter_str = quote(f'repositories.{digest_field}=="{digest}";not(deleted==true)')
 
     check_url = urljoin(args.pyxis_url, f"v1/images?page_size=1&filter={filter_str}")
 
@@ -131,8 +144,6 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
             }
         ],
         "certified": json.loads(args.certified.lower()),
-        "docker_image_digest": docker_image_digest,
-        "image_id": docker_image_digest,
         "architecture": parsed_data["architecture"],
         "parsed_data": parsed_data,
     }
@@ -145,6 +156,9 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
             }
         )
 
+    digest_field = get_digest_field(args.media_type)
+    container_image_payload["repositories"][0][digest_field] = docker_image_digest
+
     rsp = pyxis.post(upload_url, container_image_payload).json()
 
     # Make sure container metadata was successfully added to Pyxis
@@ -152,6 +166,20 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
         LOGGER.info(f"The image id is: {rsp['_id']}")
     else:
         raise Exception("Image metadata was not successfully added to Pyxis.")
+
+
+def get_digest_field(media_type: str) -> str:
+    """This will return one of the two possible digest fields
+    to use in the repository object which is embedded in the
+    ContainerImage object.
+
+    manifest_schema2_digest is used for single arch images,
+    manifest_list_digest is used for multi arch images
+    """
+    if media_type in MANIFEST_LIST_TYPES:
+        return "manifest_list_digest"
+    else:
+        return "manifest_schema2_digest"
 
 
 def main():  # pragma: no cover
--- a/pyxis/test_create_container_image.py
+++ b/pyxis/test_create_container_image.py
@@ -6,47 +6,67 @@ from create_container_image import (
     image_already_exists,
     create_container_image,
     prepare_parsed_data,
+    get_digest_field,
 )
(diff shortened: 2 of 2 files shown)


mmalina:
@ralphbean PTAL

pyxis/test_create_container_image.py
```diff
@@ -105,6 +106,8 @@ def test_create_container_image_latest(mock_datetime: MagicMock, mock_post: Magi
     args.tag = "some_version"
     args.certified = "false"
     args.is_latest = "true"
+    # single arch image
+    args.media_type = "application/vnd.docker.distribution.manifest.v2+json"
```
kasemAlem, line 110: what is the point of doing this assignment here while you already ran it in line 61?

mmalina: Line 61 is a different test. This one is `test_create_container_image_latest` and the other one is `test_create_container_image`. They are fairly similar, the main difference is that this one is testing the scenario with "latest" tag.

kasemAlem: I see, 
thanks

pyxis/create_container_image.py
```diff
@@ -42,6 +48,12 @@ def setup_argparser() -> Any:  # pragma: no cover
         help="Should the `latest` tag of the ContainerImage be overwritten?",
         required=True,
     )
+    parser.add_argument(
+        "--media-type",
```
johnbieren, line 52: we already pass the entire `skopeo inspect` output. We can't just extract it from the file instead of having to pass another argument?

johnbieren: I see. It sucks we need to skopeo inspect twice and neither way of calling it gives us all we need. But it looks like you and others already investigated that on the JIRA

johnbieren (approved):

johnbieren (changes requested):
The commit message prefix is wrong. Guess we don't have CI for that in this repo

mmalina:
> The commit message prefix is wrong. Guess we don't have CI for that in this repo

Good catch. I fixed the commit message. But you're right the linting workflow doesn't work - in fact the workflow file is missing. It was supposed to be enabled here: https://github.com/redhat-appstudio/release-service-utils/pull/55 I will create a bug.

johnbieren (approved):

scoheb (dismissed):

kasemAlem (approved):

pyxis/create_container_image.py
```diff
@@ -42,6 +48,12 @@ def setup_argparser() -> Any:  # pragma: no cover
...
+        help="The mediaType string returned by `skopeo inspect --raw`. "
```
davidmogar, line 53:
```suggestion
        help="The mediaType string returned by `skopeo inspect --raw`."
```

pyxis/create_container_image.py
```diff
@@ -145,6 +155,17 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
             }
         )
 
+    # manifest_schema2_digest is set for single arch images and
+    # manifest_list_digest is set for multi arch images
+    if args.media_type in MANIFEST_LIST_TYPES:
+        container_image_payload["repositories"][0][
```
davidmogar, line 161: Cleaner as a one-liner.

pyxis/create_container_image.py
```diff
@@ -145,6 +155,17 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
...
+            "manifest_list_digest"
+        ] = docker_image_digest
+    else:
+        container_image_payload["repositories"][0][
```
davidmogar, line 165: Cleaner as a one-liner. Not that long isn't it?

pyxis/test_create_container_image.py
```diff
@@ -57,6 +57,8 @@ def test_create_container_image(mock_datetime: MagicMock, mock_post: MagicMock):
     args.pyxis_url = mock_pyxis_url
     args.tag = "some_version"
     args.certified = "false"
+    # single arch image,
```
davidmogar, line 60: I don't get this comment and it has an extra comma.

pyxis/test_create_container_image.py
```diff
@@ -105,6 +106,8 @@ def test_create_container_image_latest(mock_datetime: MagicMock, mock_post: Magi
     args.tag = "some_version"
     args.certified = "false"
     args.is_latest = "true"
+    # single arch image
```
davidmogar, line 109: I don't get this comment.

pyxis/create_container_image.py
mmalina: @davidmogar I disagree. The string continues on the next line. So the result will be those two strings joined. I want a space between the two sentences.

pyxis/create_container_image.py
mmalina: This was done by Black formatter, because the line was too long.

pyxis/create_container_image.py
mmalina: I think it is. I didn't do this myself, the Black formatter did it.

pyxis/test_create_container_image.py
mmalina: Removed the extra comma, thanks.

The mediaType value should/will be one of these:
```
application/vnd.oci.image.manifest.v1+json
application/vnd.docker.distribution.manifest.v1+json
application/vnd.docker.distribution.manifest.v2+json
application/vnd.oci.image.index.v1+json
application/vnd.docker.distribution.manifest.list.v2+json
```

The first three imply it's a single arch image, the last two imply it's a multi arch image. Based on this value we will populate one of the two fields.

Since we're really only deciding between the two possibilities, I decided that we just need to define the "manifest list" / "oci index" strings and otherwise assume it's a manifest. Should I maybe add the other strings as a comment near where I define `MANIFEST_LIST_TYPES` in the create_container_image script? Or what else would you suggest?

ralphbean (dismissed):

mmalina:
@davidmogar @johnbieren can you check again please? I had to refactor this a bit - I realized that the original `image_already_exists()` function won't work either since it was relying on the `docker_image_digest` field which we will no longer populate. Now it will check the right field in the embedded repository object. (The commit message is updated to include this.)

davidmogar (approved):

pyxis/create_container_image.py
```diff
@@ -52,9 +64,11 @@ def image_already_exists(args, digest: str) -> bool:
 
     :return: True if one exists, else false
     """
+    # return False
```
johnbieren, line 67: I am confused by this comment

johnbieren (approved):

pyxis/create_container_image.py
mmalina: Busted! I needed to skip the test for testing ;)

johnbieren (approved):

PR #59, opened 2023-09-19
chore(HACBS-2593): internal-request script should fail on IR failure

theflockers:
This PR fixes the script to make it check if the InternalRequest
has Succeeded or not and exit accordingly.

### DIFF
--- a/utils/internal-request
+++ b/utils/internal-request
@@ -2,8 +2,17 @@
 
 # This script creates an InternalRequest resource in a Kubernetes cluster
 # using the 'kubectl' command line tool. The resource is created with
-# parameters passed to the script, and optionally waits for the resource to
-# reach a 'completed' status.
+# parameters passed to the script.
+#
+# In sync mode (default) the script waits for the InternalRequest to reach a 'completed'
+# status and will provide an exit code based on the InternalRequest status:
+#
+# Succeeded: error code 0
+# Failed: error code 21
+# Rejected: error code 22
+#
+# In async mode it creates an InternalRequest and exits with code 0 without waiting
+# for status updates.
 #
 # Usage:
 #   ./internal-request.sh -r request [-p <key=value> ...] [-s sync] [-t timeout]
@@ -113,15 +122,32 @@ if [ "$SYNC" = "true" ]; then
   # Wait until status is set or timeout
   END_TIME=$(date -ud "$TIMEOUT seconds" +%s)
   while true; do
-    STATUS=$(kubectl get internalrequest "$INTERNAL_REQUEST_NAME" -o json | jq -r '.status')
+    STATUS=$(kubectl get internalrequest "$INTERNAL_REQUEST_NAME" -o json | jq '.status')
 
     if [ "$STATUS" != "null" ]; then
-      CONDITIONS=$(kubectl get internalrequest "$INTERNAL_REQUEST_NAME" -o json | jq -r '.status.conditions')
+      CONDITIONS=$(kubectl get internalrequest "$INTERNAL_REQUEST_NAME" -o json | jq '.status.conditions')
       if [ "$(echo "$CONDITIONS" | jq 'length')" -eq 1 ]; then
         CONDITION_REASON=$(echo "$CONDITIONS" | jq -r '.[0].reason')
-        if [ "$CONDITION_REASON" != "Running" ]; then
-          echo "InternalRequest completed."
-          exit 0
+        if [ "$CONDITION_REASON" == "Running" ]; then
+          continue
+        else
+            case $CONDITION_REASON in
+                "Succeeded")
+                    echo "InternalRequest succeeded"
+                    EXIT_CODE=0
+                    ;;
+                "Failed")
+                    echo "InternalRequest failed"
+                    EXIT_CODE=21
+                    ;;
+                "Rejected")
+                    echo "InternalRequest rejected"
+                    EXIT_CODE=22
+                    ;;
+            esac
+            echo "Conditions:"
+            echo $CONDITIONS
+            exit $EXIT_CODE
         fi
       fi
     fi


johnbieren (dismissed):

utils/internal-request
```diff
@@ -121,7 +121,7 @@ if [ "$SYNC" = "true" ]; then
         CONDITION_REASON=$(echo "$CONDITIONS" | jq -r '.[0].reason')
         if [ "$CONDITION_REASON" != "Running" ]; then
           echo "InternalRequest completed."
-          exit 0
+          [ "${CONDITION_REASON}" == "Succeeded" ]; exit $?
```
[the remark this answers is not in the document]
theflockers: Cool. I changed but added more output only for the failed case.

davidmogar:
That was actually on purpose as you can see in the script header:
```
# This script creates an InternalRequest resource in a Kubernetes cluster
# using the 'kubectl' command line tool. The resource is created with
# parameters passed to the script, and optionally waits for the resource to
# reach a 'completed' status.
```

I delegated the check for the caller as a succeeded or failed condition would probably still require the resource to be loaded to check the results or the condition message respectively.

I'm not saying I oppose to the change. Just saying that it wasn't the goal.

In addition, the Rejected status is not taken into account in this PR. Failed and Rejected are grouped together, which again would require to load the resource to see what happened.

utils/internal-request
```diff
@@ -119,9 +119,15 @@ if [ "$SYNC" = "true" ]; then
...
-          echo "InternalRequest completed."
-          exit 0
+        if [ "$CONDITION_REASON" == "Running" ]; then
+          continue
+        elif [ "$CONDITION_REASON" == "Succeeded" ]; then
+            echo "InternalRequest succeeded"
+            exit 0
+        else
+            echo "InternalRequest failed"
+            echo $CONDITIONS
```
mmalina, line 129: Nit: I would suggest: 
```sh
echo Conditions: 
echo $CONDITIONS
```

utils/internal-request
```diff
@@ -119,9 +119,15 @@ if [ "$SYNC" = "true" ]; then
       CONDITIONS=$(kubectl get internalrequest "$INTERNAL_REQUEST_NAME" -o json | jq -r '.status.conditions')
```
mmalina, line 119: The `-r` in `jq -r` is redundant here. The result is a json array. `--raw` only has effect if the result is a string:

> With this option, if the filter's result is a string then it will be written directly to standard output rather than being formatted as a JSON string with quotes.

utils/internal-request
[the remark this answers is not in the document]
mmalina: Nice, thanks. Added a few more small comments.

mmalina:
> That was actually on purpose

I was actually wondering if this was the case. And it's a valid argument - depending on what the internal request was, the caller might want to do different things to check why it failed. But on the other hand the calling script can _catch_ this failure and still check what actually happened on its own. So I'm fine either way. But in both cases, this should be clearly documented in the script description (currently it's not clear what happens).

theflockers:
> That was actually on purpose as you can see in the script header:
> 
> ```
> # This script creates an InternalRequest resource in a Kubernetes cluster
> # using the 'kubectl' command line tool. The resource is created with
> # parameters passed to the script, and optionally waits for the resource to
> # reach a 'completed' status.
> ```
> 
> I delegated the check for the caller as a succeeded or failed condition would probably still require the resource to be loaded to check the results or the condition message respectively.
> 
> I'm not saying I oppose to the change. Just saying that it wasn't the goal.
> 
> In addition, the Rejected status is not taken into account in this PR. Failed and Rejected are grouped together, which again would require to load the resource to see what happened.

I understand but I can't see a reason (so far) where the IR would fail and we should still continue.  
I see this like any other script or command that return non zero in case of non-success (like `grep` for instance). 

Knowing that the IR did not succeeded by the exit code IMHO should be default behavior as we can still test the exit code of the script to avoid premature exit.

davidmogar:
> > That was actually on purpose as you can see in the script header:
> > ```
> > # This script creates an InternalRequest resource in a Kubernetes cluster
> > # using the 'kubectl' command line tool. The resource is created with
> > # parameters passed to the script, and optionally waits for the resource to
> > # reach a 'completed' status.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I delegated the check for the caller as a succeeded or failed condition would probably still require the resource to be loaded to check the results or the condition message respectively.
> > I'm not saying I oppose to the change. Just saying that it wasn't the goal.
> > In addition, the Rejected status is not taken into account in this PR. Failed and Rejected are grouped together, which again would require to load the resource to see what happened.
> 
> I understand but I can't see a reason (so far) where the IR would fail and we should still continue. I see this like any other script or command that return non zero in case of non-success (like `grep` for instance).
> 
> Knowing that the IR did not succeeded by the exit code IMHO should be default behavior as we can still test the exit code of the script to avoid premature exit.

Yes, as stated, the script was not created that way. It was meant to create IR and, optionally, wait for them to finish (independently of the result). In this context, the script should exclusively error in the case of a problem creating the resource. With the changes, the script could fail this way or could fail because the IR failed. If you want to go that way, do it. But if we go that way, proper documentation and proper error codes should be put into place. If not we just simply let the called to check the outcome of the request as it would still have to load it in any case.

Regarding an example for an IR failing and still continue, let's say that we need to call a service as soon as we did something and not wait for another PipelineRun to be re-run to fix the error. I don't know. I agree with this being not needed today, but I don't see it either as a problem.

In any case, leaving the decision to the team ;)

theflockers:
Any other opinions @johnbieren @happybhati @scoheb?

scoheb:
> Any other opinions @johnbieren @happybhati @scoheb?

My only requirement or opinion is that as a user, something that fails in a script but shows as passed is not good. Either the script should fail or the calling task/pipeline should fail.

theflockers:
Also, the default behavior of the script is watch the internal request. So, if we are waiting it to finish, it makes sense IMHO that you want to know the result of it.

In the other hand, for an _async_ call (`-s false`) it exists exits with 0 and you need to watch the IR in the task itself.

theflockers:
@johnbieren could you give your thoughts here? See David's comment, please.

johnbieren:
> @johnbieren could you give your thoughts here? See David's comment, please.

I really do not know enough about this script to decide its behavior

happybhati:
> > Any other opinions @johnbieren @happybhati @scoheb?
> 
> My only requirement or opinion is that as a user, something that fails in a script but shows as passed is not good. Either the script should fail or the calling task/pipeline should fail.

I would +1 what Scott said; either way, this should be documented in the script description(as Martin suggested).

theflockers:
Ok. I will then adjust the doc a bit and move forward with the change.

utils/internal-request
```diff
@@ -2,8 +2,16 @@
 
 # This script creates an InternalRequest resource in a Kubernetes cluster
 # using the 'kubectl' command line tool. The resource is created with
-# parameters passed to the script, and optionally waits for the resource to
-# reach a 'completed' status.
+# parameters passed to the script.
+#
+# In sync mode (default) it waits for the InternalRequest to reach a 'completed'
```
davidmogar, line 7:
```suggestion
# In sync mode (default) the script waits for the InternalRequest to reach a 'completed'
```

davidmogar (changes requested):

utils/internal-request
```diff
@@ -113,15 +122,32 @@ if [ "$SYNC" = "true" ]; then
...
+                    EXIT_CODE=21
+                    ;;
+                "Rejected")
+                    echo "InternalRequest rejected"
+                    EXIT_CODE=22
+                    ;;
+            esac
+            echo "Conditions:"
+            echo $CONDITIONS
+            exit $EXIT_CODE
```
davidmogar, line 150: Is this tested? Does it work? Unsure about variable scope in bash.

theflockers: Yes, it does. Tested of all 3 scenarios.

theflockers: # case Success
```
+ CONDITION_REASON=Succeeded
+ '[' Succeeded == Running ']'
+ case $CONDITION_REASON in
InternalRequest succeeded
+ echo 'InternalRequest succeeded'
+ EXIT_CODE=0
+ echo Conditions:
Conditions:
[ { "lastTransitionTime": "2023-09-25T12:29:31Z", "message": "", "reason": "Succeeded", "status": "True", "type": "Succeeded" } ]
+ echo '[' '{' '"lastTransitionTime":' '"2023-09-25T12:29:31Z",' '"message":' '"",' '"reason":' '"Succeeded",' '"status":' '"True",' '"type":' '"Succeeded"' '}' ']'
+ exit 0
```

# case Failed
```
+ CONDITION_REASON=Failed
+ '[' Failed == Running ']'
+ case $CONDITION_REASON in
InternalRequest failed
+ echo 'InternalRequest failed'
+ EXIT_CODE=21
Conditions:
+ echo Conditions:
+ echo '[' '{' '"lastTransitionTime":' '"2023-09-25T11:07:06Z",' '"message":' '"Tasks' Completed: 1 '(Failed:' 1, Cancelled '0),' Skipped: '0",' '"reason":' '"Failed",' '"status":' '"False",' '"type":' '"Succeeded"' '}' ']'
[ { "lastTransitionTime": "2023-09-25T11:07:06Z", "message": "Tasks Completed: 1 (Failed: 1, Cancelled 0), Skipped: 0", "reason": "Failed", "status": "False", "type": "Succeeded" } ]
+ exit 21
```

# case Rejected
```
v+ CONDITION_REASON=Rejected
+ '[' Rejected == Running ']'
+ case $CONDITION_REASON in
+ echo 'InternalRequest rejected'
+ EXIT_CODE=22
+ echo Conditions:
+ echo '[' '{' '"lastTransitionTime":' '"2023-09-25T11:09:57Z",' '"message":' '"the' internal request namespace '(managed-release-team-tenant)' is not in the allow 'list",' '"reason":' '"Rejected",' '"status":' '"False",' '"type":' '"Succeeded"' '}' ']'
+ exit 22
```

davidmogar:
These PR changes the behavior of, at least, a release service bundle task. Do we need a change in there?

theflockers:
It is not required. With the change it will now behave as expected. - Fail on error.

davidmogar (approved):

PR #63, opened 2023-10-03
feat(RHTAPREL-392): add jinja2 and templating to the image

johnbieren:
As part of RHTAPREL-392, we need to create advisories. The easiest way to do this is from a template using ansible, so this commit adds ansible and a template to the repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,16 +20,21 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     jq \
     python39-devel \
     diffutils \
+    python39-pip \
     python39-requests \
     skopeo \
     krb5-workstation \
     && dnf clean all
+
+RUN pip3 install jinja2 \
+    jinja2-ansible-filters
 
 ADD data/certs/2015-IT-Root-CA.pem data/certs/2022-IT-Root-CA.pem /etc/pki/ca-trust/source/anchors/
 RUN update-ca-trust
 
 COPY pyxis /home/pyxis
 COPY utils /home/utils
+COPY templates /home/templates
 
 # Set HOME variable to something else than `/` to avoid 'permission denied' problems when writing files.
 ENV HOME=/tekton/home
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,4 @@
 pytest
 requests
+jinja2
+jinja2-ansible-filters
--- a/utils/test_apply_template.py
+++ b/utils/test_apply_template.py
@@ -1,0 +1,44 @@
+import pytest
+from unittest.mock import patch, MagicMock
+from apply_template import setup_argparser, main
+
+
+@patch(
+    "argparse._sys.argv",
+    ["apply_template", "--data", "{}", "--template", "somefile", "-o", "newfile"],
+)
+def test_setup_argparser_proper_args():
+    args_out = setup_argparser()
+    assert args_out.data == "{}"
+    assert args_out.template == "somefile"
+    assert args_out.output == "newfile"
+
+
+def test_setup_argparser_improper_args():
+    with pytest.raises(SystemExit) as e:
+        setup_argparser()
+    assert e.value.code == 2
+
+
+@patch("builtins.open")
+@patch("apply_template.Template.render")
+@patch("apply_template.setup_argparser")
+def test_apply_template_advisory_template(
+    mock_argparser: MagicMock, mock_render: MagicMock, mock_open: MagicMock
+):
+    args = MagicMock()
+    args.template = "templates/advisory.yaml.jinja"
+    args.data = "{}"
+    args.output = "somefile"
+    mock_argparser.return_value = args
+    mock_render.return_value = "applied template file"
+    mock_open1 = MagicMock()
+    mock_open2 = MagicMock()
+    mock_open.side_effect = [mock_open1, mock_open2]
+    mock_open1.__enter__.return_value.read.return_value = "foo: bar"
+    file = mock_open2.__enter__.return_value
+
+    # Act
+    main()
+
+    file.write.assert_called_once_with("applied template file")
--- a/templates/advisory.yaml.jinja
+++ b/templates/advisory.yaml.jinja
@@ -1,0 +1,23 @@
+apiVersion: rhtap.redhat.com/v1alpha1
+kind: Advisory
+metadata:
+  name: {{ advisory_name }}
+spec:
+  product_id: {{ advisory.spec.product_id }}
+  cpe: {{ advisory.spec.cpe }}
+  type: {{ advisory.spec.type }}
+{%- if 'issues' in advisory.spec %}
+  issues:
+    {{ advisory.spec.issues | to_nice_yaml(indent=2) | indent(4) | trim }}
+{%- endif %}
+  content:
+    {{ advisory.spec.content | to_nice_yaml(indent=2) | indent(4) | trim }}
+  synopsis: {{ advisory.spec.synopsis }}
+  topic: >-
+    {{ advisory.spec.topic | wordwrap(76) | indent(4) }}
+  description: >-
+    {{ advisory.spec.description | wordwrap(76) | indent(4) }}
+  solution: >-
+    {{ advisory.spec.solution | wordwrap(76) | indent(4) }}
+  references:
+    {{ advisory.spec.references | to_nice_yaml | indent(4) }}
--- a/utils/apply_template.py
+++ b/utils/apply_template.py
@@ -1,0 +1,51 @@
+#!/usr/bin/env python3
+from jinja2 import Template
+from jinja2_ansible_filters import AnsibleCoreFiltersExtension
+import argparse
+import json
+
+
+def setup_argparser() -> argparse.Namespace:  # pragma: no cover
+    """Setup argument parser
+
+    :return: Initialized argument parser
+    """
+
+    parser = argparse.ArgumentParser(description="Applies a template.")
+
+    parser.add_argument(
+        "--data",
+        help="JSON string containing data to use in the template.",
+        required=True,
+    )
+    parser.add_argument(
+        "--template",
+        help="Path to the template file to use.",
+        required=True,
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        help="The desired filename of the result.",
+        required=True,
+    )
+    return parser.parse_args()
+
+
+def main():  # pragma: no cover
+    """Main func"""
+
+    args = setup_argparser()
+
+    with open(args.template) as t:
+        template = Template(t.read(), extensions=[AnsibleCoreFiltersExtension])
+    content = template.render(json.loads(args.data))
+
+    filename = args.output
+    with open(filename, mode="w", encoding="utf-8") as advisory:
+        advisory.write(content)
+        print(f"Wrote {filename}")
+
+
+if __name__ == "__main__":  # pragma: no cover
+    main()


ansible/templates/advisory.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: rhtap.redhat.com/v1alpha1
```
scoheb, line 1: why is this file in release-service-utils?

are you planning on referring to it in a script in the release-utils image?

johnbieren: Because my task that creates the advisory does it through this
```
          # Create advisory file
          ansible localhost -m template -a "src=/home/ansible/templates/advisory.yaml dest=$ADVISORY_FILEPATH" -e advisory_name=$ADVISORY_NAME \
             -e advisory_type=$(get_advisory_type) -e '{"advisory":$(params.advisory_json)}'
```

davidmogar: You already have python. You could use Jinja direcly. I don't oppose to Ansible, but Ansible uses Jinja.

mmalina: Yes, installing `ansible` just to render a jinja template seems like an overkill. Although it is true that to use Jinja directly, you will need a piece of Python code. (There seems to be a cli tool for that, but it doesn't seem popular or maintained, so I wouldn't trust it.) So I would prefer to use Jinja directly via Python. But if you want to avoid writing a Python script, I would be fine with using ansible. But in that case we should install ansible via dnf, unless it's not available.

davidmogar: Pip seems to be the recommended way of installing Ansible.

mmalina: Also, I'm not sure if the template should live here. But we probably don't have a better place for it right now... And having it as a file here still seems much better compared to creating it inline in some task (which is the only other option I can think of ATM).

johnbieren: `ansible` is not available in ubi8 via dnf. We really are against using ansible? I'm a little confused... you both said using ansible is fine but opened a discussion about not using it. Which is it?

johnbieren: It is a template that is used as a helper for a release service function. I'm not sure why it doesn't belong here. Isn't the entire point of release-service-utils to hold things we need for other functions/tasks?

mmalina: (We could also create the yaml file directly using the yaml library in Python.)

davidmogar: I'm just saying that if I want to use Jinja, I install Jinja and not something that uses Jinja. I'm also saying that I'm ok with you using Ansible if you want. The image is going to be bigger for sure, but I don't think that's much of a problem at this point. Leaving the decision to you.

johnbieren: I think there are many ways we could create the yaml. I'd like to use ansible, as shown by using it here. It is a Red Hat product and has fantastic documentation. If there is a strong argument against it, please say so, otherwise let's resolve this discussion

davidmogar: > It is a template that is used as a helper for a release service function. I'm not sure why it doesn't belong here. Isn't the entire point of release-service-utils to hold things we need for other functions/tasks?

If the template is going to be used in a single place, it should not be here. In this case, putting the yaml in the task itself seems reasonable tbh and you could just construct it using bash variables maybe. I don't know what every body else things about this.

johnbieren: I don't think `going to be used in a single place, it should not be here` is a pattern we have followed. Take the pyxis stuff for example. The create_container_image script is used in one place. It lives here. The push_sbom script is used in one place. It lives here. I really think constructing it with ansible is much nicer and easier to read than using bash

davidmogar: Using Jinja directly would require to do something like:
```
import jinja2
environment = jinja2.Environment()
template = environment.from_string(you_template)
template.render(your_variables)
```
This script could live in here and that would save you from installing Ansible. So, if you have an strong opinion about using Ansible and you think that is better or that is going to be used in other cases, go ahead.

mmalina: As David says, in principle it seems wrong to install quite a big binary just to use something that you can use directly. But same as David, I'm fine with it if it makes your life easier.

Regarding where to store the template. I said it's probably fine. I shouldn't have mentioned it. It's just that the name release-service-utils (as in utilities) sounds more like scripts and functions and cli tools to be used by the tekton tasks. A template is more like a data file - a resource to be used by our tasks. But as I said, I don't think we have a better place for it.

davidmogar: Btw, `pip install ansible` comes, if I'm not mistaken, with ansible_collections and that's more than 500M. Is `ansible-core` enough for you?

davidmogar: Ansible itself is 14MB (not sure if this is core itself) plus other packages that it uses. Jinja2 is less than 2MB.

johnbieren: I'm looking into using jinja2 without ansible. The only hurdle so far is that some filters I am using are showing up as nonexistent (but were when I was using ansible) so I am looking into it. Assuming I fix it, I will update the PR to not include ansible

johnbieren: Updated to remove ansible. I know I am missing a test, but otherwise it is open for review

templates/advisory.yaml
```diff
@@ -0,0 +1,23 @@
...
+kind: Advisory
+metadata:
+  name: {{ advisory_name }}
+spec:
+  product_id: {{ advisory.spec.product_id }}
+  cpe: {{ advisory.spec.cpe }}
+  type: {{ advisory_type }}
+{%- if 'issues' in advisory.spec %}
+  issues:
+    {{ advisory.spec.issues | to_nice_yaml(indent=2) | comment(decoration='    ') | trim }}
```
mmalina, line 11: Can you explain the `comment` filter? According to the docs, it's used for adding comments in a file:

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_filters.html#adding-comments-to-files

But by setting custom `decoration` you're basically just using it to add indentation? Can you use the `indent()` filter instead? And why is the `trim` needed?

templates/advisory.yaml
```diff
@@ -0,0 +1,23 @@
...
+{%- endif %}
+  content:
+    {{ advisory.spec.content | to_nice_yaml(indent=2) | comment(decoration='    ') | trim }}
+  synopsis: {{ advisory.spec.synopsis }}
+  topic: >-
+    {{ advisory.spec.topic | wordwrap(76) | replace('\n', '\n    ') }}
+  description: >-
+    {{ advisory.spec.description | wordwrap(76) | replace('\n', '\n    ') }}
+  solution: >-
+    {{ advisory.spec.solution | wordwrap(76) | replace('\n', '\n    ') }}
```
mmalina, line 21: Seeing the `replace` and also the `to_nice_yaml ` in other places, I think we would be better off just generating the yaml from Python directly. But if you prefer this, I guess it's fine.

But wouldn't `indent` work in this case? At least we could stay consistent.

utils/apply-template.py
```diff
@@ -0,0 +1,53 @@
...
+    args = parser.parse_args()
+
+    with open(args.template) as t:
+        template = Template(t.read(), extensions=[AnsibleCoreFiltersExtension])
+    content = template.render(json.loads(args.data))
+
+    filename = args.output
+    with open(filename, mode="w", encoding="utf-8") as advisory:
+        advisory.write(content)
+        print(f"Wrote {filename}")
```
mmalina, line 49: This should probably be `println` to also print a new line at the end.

templates/advisory.yaml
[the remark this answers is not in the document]
johnbieren: Sure, I'll change it

utils/apply-template.py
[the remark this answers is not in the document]
johnbieren: Yeah.. I am not a big fan of them either so I wanted to see if I could get it through review without doing it haha

utils/apply-template.py
johnbieren: will change it

templates/advisory.yaml
johnbieren: I'm far from a jinja2 expert, so I just played around with things until it worked. Basically, my issue was that nested objects did not have proper indentation. The comment with the indent decoration fixed that.

templates/advisory.yaml
johnbieren: I can try it

utils/apply-template.py
mmalina: Sorry! That's what happens when you don't use a language for a while. And I wasn't sure, so I did a quick google and it seemed you needed println, but it turns out there is no such thing as println? Anyway, disregard this.

johnbieren: Agreed :joy: I went right ahead with the change and failed for no println

templates/advisory.yaml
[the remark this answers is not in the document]
johnbieren: fixed

templates/advisory.yaml
johnbieren: changed to indent, good call

templates/advisory.yaml
johnbieren: Replied offline

Dockerfile
```diff
@@ -20,16 +20,21 @@ RUN dnf -y --setopt=tsflags=nodocs install \
...
     python39-devel \
     diffutils \
+    python39-pip \
     python39-requests \
     skopeo \
     krb5-workstation \
     && dnf clean all
 
+RUN pip3 install Jinja2 \
+    jinja2-ansible-filters
```
mmalina, line 30: Can you add these also to `requirements.txt` please?

johnbieren: I don't see a `requirements.txt` you mean create a new one and add them?

mmalina: Oh, my bad. I have one locally with just `requests` in it. I think it's a good practice to provide one in a repo so that it's easier to get started - usually that's what you expect in a Python project.

So I would propose to add one with:
```
requests
Jinja2
jinja2-ansible-filters.
```
But if you disagree, I can continue to live with a local one.

johnbieren: added

mmalina: Thanks.

requirements.txt
mmalina, line 1: Thanks for adding the file. But do we need the versions? Because if we want to lock down the versions, we should first of all do it in the Dockerfile . Either way, I don't think it makes sense to define the versions here, when they won't be used in the actual image.

johnbieren: I like no versions. Removed

mmalina: Fine with me, thanks.

utils/apply-template.py
```diff
@@ -0,0 +1,53 @@
...
+
+def main():  # pragma: no cover
+    """Main func"""
+
+    parser = setup_argparser()
+    args = parser.parse_args()
+
+    with open(args.template) as t:
+        template = Template(t.read(), extensions=[AnsibleCoreFiltersExtension])
+    content = template.render(json.loads(args.data))
```
johnbieren, line 44: @davidmogar @mmalina I started to write a test for this, but do you guys think we really need one? I can move this line to a separate function and write a test for it, but it is really just this line. I wouldn't be testing the `open` or `write` parts of `main`, so this is really all that is left, which is just calling one function from another library

mmalina: I understand that this is a simple script. But I think it makes sense to add a test to stay consistent. We should also include tests for the setup_argparser function (I would recommend to move the `parser.parse_args()` call inside that function to be really able to test that various CLI arguments return the expected vars (or a failure).

johnbieren: The test fails for not having jinja... do I need to update the github workflow to pip install that before running `pytest`? Seems a little messy if so

mmalina: I guess that's where the requirements.txt comes in handy? We should just install all of that before running the tests: `python3 -m pip install -r requirements.txt`

johnbieren: Okay let me split that out into a separate PR

johnbieren: https://github.com/redhat-appstudio/release-service-utils/pull/64

mmalina: Cool, thanks, approved.

davidmogar (dismissed):

utils/test_apply_template.py
```diff
@@ -0,0 +1,34 @@
...
+    args.template = "templates/advisory.yaml.jinja"
+    args.data = "{}"
+    args.output = "somefile"
+    mock_argparser.return_value = args
+    mock_render.return_value = "applied template file"
+
+    # Act
+    main()
+
+    assert os.path.exists("somefile")
```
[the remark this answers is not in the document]
johnbieren: Thanks, updated

utils/test_apply_template.py
```diff
@@ -0,0 +1,44 @@
...
+    args = MagicMock()
+    args.template = "templates/advisory.yaml.jinja"
+    args.data = "{}"
+    args.output = "somefile"
+    mock_argparser.return_value = args
+    mock_render.return_value = "applied template file"
+    mock_open1 = MagicMock()
+    mock_open2 = MagicMock()
+    mock_open.side_effect = [mock_open1, mock_open2]
+    mock_open1.__enter__.return_value.read.side_effect = "foo: bar"
```
mmalina, line 38: Why use `side_effect` here? `side_effect` is normally used either if you want to throw an exception, or if you need to return different values for subsequent calls. So I think this should be: `mock_open1.__enter__.return_value.read.return_value = "foo: bar"`

utils/test_apply_template.py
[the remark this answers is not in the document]
mmalina: Thanks. One last small comment.

utils/test_apply_template.py
johnbieren: good point, fixed

mmalina (approved):

PR #64, opened 2023-10-10
chore: add requirements.txt file

johnbieren:


mmalina (approved):
For reference, this came up from discussion here: https://github.com/redhat-appstudio/release-service-utils/pull/63#discussion_r1348651047

And here: https://github.com/redhat-appstudio/release-service-utils/pull/63#discussion_r1349157940

PR #66, opened 2023-10-11
chore(RHTAPREL-435): pin base image and update labels

scoheb:
- pin base image in Dockerfile
- specify labels that we should not be overriding
- upgrade OPM version due to
https://redhat-internal.slack.com/archives/C3VS0LV41/p1697051133176919

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,18 +1,16 @@
-FROM registry.access.redhat.com/ubi8/ubi
+FROM registry.access.redhat.com/ubi8/ubi:8.8-1067.1696517599
 
 ARG COSIGN_VERSION=2.1.1
 ARG KUBECTL_VERSION=1.27.2
-ARG OCP_VERSION=4.13.3
-ARG JQ_VERSION=1.6
+ARG OCP_VERSION=4.14.0-rc.5
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
 
-RUN curl -L https://github.com/stedolan/jq/releases/download/jq-{JQ_VERSION}/jq-linux64 -o /usr/bin/jq &&\
-    curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
+RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
     curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl &&\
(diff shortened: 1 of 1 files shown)


johnbieren (approved):

Dockerfile
```diff
@@ -7,13 +7,14 @@ ARG JQ_VERSION=1.6
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
 
-RUN curl -L https://github.com/stedolan/jq/releases/download/jq-{JQ_VERSION}/jq-linux64 -o /usr/bin/jq &&\
```
johnbieren, line 10: Why the changes to the `RUN`?

scoheb: This is a test to determine which binary is causing an image layer to have high vulnerability.

johnbieren (changes requested):
We use jq extensively in our tasks. We can't remove it from the image

scoheb:
> We use jq extensively in our tasks. We can't remove it from the image

We install it immediately after via `dnf`:

https://github.com/redhat-appstudio/release-service-utils/blob/6caa5f1c252f17276465b4d2dcf2242b66356bea/Dockerfile#L18

johnbieren:
> > We use jq extensively in our tasks. We can't remove it from the image
> 
> We install it immediately after via `dnf`:
> 
> https://github.com/redhat-appstudio/release-service-utils/blob/6caa5f1c252f17276465b4d2dcf2242b66356bea/Dockerfile#L18

Oh wow. I was the one who approved that addition installing it via curl when we already dnf installed it too. Oops. Hopefully no version diff issues once we merge this...

johnbieren (approved):

scoheb:
> > > We use jq extensively in our tasks. We can't remove it from the image
> > 
> > 
> > We install it immediately after via `dnf`:
> > https://github.com/redhat-appstudio/release-service-utils/blob/6caa5f1c252f17276465b4d2dcf2242b66356bea/Dockerfile#L18
> 
> Oh wow. I was the one who approved that addition installing it via curl when we already dnf installed it too. Oops. Hopefully no version diff issues once we merge this...

I can put it back to reduce the risk...but it seems safer to get it from `dnf`

johnbieren:
> > > > We use jq extensively in our tasks. We can't remove it from the image
> > > 
> > > 
> > > We install it immediately after via `dnf`:
> > > https://github.com/redhat-appstudio/release-service-utils/blob/6caa5f1c252f17276465b4d2dcf2242b66356bea/Dockerfile#L18
> > 
> > 
> > Oh wow. I was the one who approved that addition installing it via curl when we already dnf installed it too. Oops. Hopefully no version diff issues once we merge this...
> 
> I can put it back to reduce the risk...but it seems safer to get it from `dnf`

Yeah I say get it only from dnf and if we have issues we can deal with them

PR #68, opened 2023-10-17
feat: get-resource returns {} if json key not found

johnbieren:
This commit changes the behavior of the get-resource util to return {} instead of the empty string if a jsonpath is passed and the key is not found.

### DIFF
--- a/utils/get-resource
+++ b/utils/get-resource
@@ -37,7 +37,7 @@ IFS='/' read -r namespace name <<< "$namespaced_name"
 
 # If a jsonpath is provided, append it to the kubectl command
 if [ ! -z "$jsonpath" ]; then
-  kubectl get $resource_type -n $namespace $name -o jsonpath="$jsonpath"
+  kubectl get $resource_type -n $namespace $name -o jsonpath="$jsonpath" --allow-missing-template-keys=false || echo "{}"
 else
   kubectl get $resource_type -n $namespace $name -o json
 fi


scoheb (approved):

davidmogar (approved):

mmalina:
Sorry to be late here. I can imagine situations where this wouldn't make sense, e.g. if you're looking for a string field and it's not there - then being given `{}` seems wrong. But since this is only used in the collect-data task, I guess it's ok.

What's not ideal is that it will now return `{}` and not fail even if the resource doesn't exist at all or if `kubectl` can't even access the cluster.

Also, I would have liked to have this behavior documented in the description.

johnbieren:
> Sorry to be late here. I can imagine situations where this wouldn't make sense, e.g. if you're looking for a string field and it's not there - then being given `{}` seems wrong. But since this is only used in the collect-data task, I guess it's ok.
> 
> What's not ideal is that it will now return `{}` and not fail even if the resource doesn't exist at all or if `kubectl` can't even access the cluster.
> 
> Also, I would have liked to have this behavior documented in the description.

I can open another PR to improve the documentation. I do agree about the second point, I planned to bring it up on the PR updating collect-data

mmalina:
> I can open another PR to improve the documentation. I do agree about the second point, I planned to bring it up on the PR updating collect-data

Probably not necessary to open a separate PR for the documentation. But if you're gonna address the second point, you can change it then. Also, I can do it :) Let me know.

johnbieren:
> > I can open another PR to improve the documentation. I do agree about the second point, I planned to bring it up on the PR updating collect-data
> 
> Probably not necessary to open a separate PR for the documentation. But if you're gonna address the second point, you can change it then. Also, I can do it :) Let me know.

I opened a thread on the other PR and we can talk about what we want to do there

PR #72, opened 2023-10-23
chore(RHTAPREL-645): script to promote overlays

scoheb:


johnbieren (approved):
Seems you already use it and it works, lgtm

PR #76, opened 2023-10-27
feat(RHTAPREL-618): add rh-push option when creating pyxis image

mmalina:
If set to true, the registry and repository entries in the Pyxis Container Image object will be converted to use Red Hat's official registry. E.g. a mapped repository
of quay.io/redhat-pending/product---my-image will be converted to use registry registry.access.redhat.com and repository product/my-image.

Also, the image will be marked as published.

This will be used in the rh-push-to-registry-redhat-io pipeline.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -1,4 +1,39 @@
 #!/usr/bin/env python3
+"""
+Python script to create a Container Image object in Pyxis
+
+Note about releasing to registry.redhat.io (using `--rh-push true` CLI argument):
+
+Our goal is to be able to download images from registry.redhat.io. For that to happen,
+an image needs to be pushed to quay.io/redhat-prod/$PRODUCT----$IMAGE, e.g.
+quay.io/redhat-prod/rhtas-tech-preview----tuf-server-rhel9. When creating
+the Container Image object in Pyxis, the registry needs to be set to
+registry.access.redhat.com and the repository would be rhtas-tech-preview/tuf-server-rhel9
+in the example above ("----" converted to "/"). This also requires a corresponding
+Container Repository object to exist in Pyxis. This will typically be created as part
+of product onboarding to RHTAP.
+
+For stage, if you want to be able to pull an image from registry.stage.redhat.io,
+the image is pushed to quay.io/redhat-pending, the Container Image is created
+in stage Pyxis, but the registry value in Pyxis is still set to registry.access.redhat.com.
+
+Why is the registry set to registry.access.redhat.com and not registry.redhat.io?
+Mostly for historical reasons.
+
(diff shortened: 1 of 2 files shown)


pyxis/create_container_image.py
```diff
@@ -54,6 +54,15 @@ def setup_argparser() -> Any:  # pragma: no cover
         "Used to determine if it's a single arch or multiarch image.",
         required=True,
     )
+    parser.add_argument(
+        "--rh-push",
+        help="If set to true, the registry and repository entries in the Pyxis Container "
```
scoheb, line 59: Missing a word?

pyxis/create_container_image.py
```diff
@@ -54,6 +54,15 @@ def setup_argparser() -> Any:  # pragma: no cover
...
+        "Image object will be converted to use Red Hat's official registry. E.g. a mapped "
+        "repository of quay.io/redhat-pending/product---my-image will be converted to use "
+        "registry registry.access.redhat.com and repository product/my-image. Also, "
```
scoheb, line 62: I am always confused by this. When do you specify `access` and when do you specify ` registry.redhat.io`?

scoheb (changes requested):

pyxis/create_container_image.py
mmalina: No, the string is split into multiple lines to avoid a line too long. See the next lines.

pyxis/create_container_image.py
mmalina: AFAIK, it's always `registry.access.redhat.com`. `registry.access.redhat.com` was here first. Then `registry.redhat.io` came along. Which one is used is controlled by the `requires_terms` flag in [the Container Repository object](https://pyxis.engineering.redhat.com/docs/objects/ContainerRepository.html?tab=Fields). (It is/was also possible to use both, but I'm not sure how that is achieved.) Nowadays everybody should use `requires_terms=true` by default. To use the public repo, you would need an exception. So I assume in RHTAP we currently only plan to use the authenticated one (registry.redhat.io).

pyxis/create_container_image.py
scoheb: yep. sorry

pyxis/create_container_image.py
scoheb: can we summarise this and add to help? or as a comment? I think this is golden info

mmalina: I added a lengthy note about this in the script description (at the top). Please check.

scoheb (approved):

pyxis/create_container_image.py
scoheb: thanks!

johnbieren (approved):

PR #78, opened 2023-10-30
feat(RHTAPREL-643): add git function to push with retries

johnbieren:
Add a function to the git-functions utility function to push with retries, rebasing if the push fails.


utils/git-functions
```diff
@@ -106,6 +106,43 @@ git_rebase() {
...
+
+    attempt=0
+    until [ "$attempt" -gt "$RETRIES" ] ; do
+      git push "$URL" && break
+      attempt=$((attempt+1))
+      git pull --rebase origin $BRANCH
+    done
+    if [ "$attempt" -gt "$RETRIES" ] ; then
+      echo "Max retries exceeded."
+      exit 1
```
mmalina, line 142: This should probably be `return 1`. `exit` will stop the whole script execution immediately, which may be ok, but it should be up to the calling script to decide this. 

For example, if you do `git_push_with_retries || true` you would normally expect to carry on even if the function fails. But when `exit` is used in the function, it will just exit the whole script execution.

davidmogar: Return looks better to me as well.

johnbieren: fixed

davidmogar (approved):

utils/git-functions
mmalina: cool

mmalina (approved):

PR #81, opened 2023-11-01
feat(RHTAPREL-643): keep undefined variables in templates

johnbieren:
Change the apply_template.py util to keep the undefined variables in the template instead of removing them.

### DIFF
--- a/utils/apply_template.py
+++ b/utils/apply_template.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-from jinja2 import Template
+from jinja2 import Template, DebugUndefined
 from jinja2_ansible_filters import AnsibleCoreFiltersExtension
 import argparse
 import json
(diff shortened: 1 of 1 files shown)


utils/apply_template.py
mmalina, line 1: Can you explain why we want to keep them? What's the use case? Maybe add a brief note to the PR/commit description?

johnbieren: Because I want to be able to modify the ones I didn't pass in later steps e.g. gitlab ci pipelines. This is focused mainly on `{{ advisory_name }}`

johnbieren: I didn't mention it in the commit because I generally just think it is better to leave what is not provided instead of replace with empty string

mmalina: Yeah, ok, thanks, makes sense.

mmalina (approved):

happybhati (approved):
LGTM 👍

PR #86, opened 2023-11-03
feat(RHTAPBUGS-946): support multiple tags in create_pyxis_image

mmalina:
The tag cli argument was renamed to tags and now it supports multiple tags using a single string with tags separated by space.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -69,8 +69,9 @@ def setup_argparser() -> Any:  # pragma: no cover
     )
     parser.add_argument("--certified", help="Is the ContainerImage certified?", required=True)
     parser.add_argument(
-        "--tag",
-        help="The ContainerImage tag name to upload",
+        "--tags",
+        help="Tags to include in the ContainerImage object. It can be a single tag "
+        "or multiple tags separated by space",
         required=True,
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


pyxis/create_container_image.py
```diff
@@ -176,19 +177,20 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
     del parsed_data["name"]
 
     upload_url = urljoin(args.pyxis_url, "v1/images")
+
+    pyxis_tags = [{
+        "added_date": date_now,
+        "name": tag,
+    } for tag in args.tags.split(" ")]
```
ralphbean, line 184: `args.tags.split()` would be more forgiving in the case the input includes two spaces:

```python
# two trailing spaces
>>> "foo  ".split(" ")
['foo', '', '']
>>> "foo  ".split()
['foo']
# two spaces inbetween foo and bar
>>> "foo  bar".split(" ")
['foo', '', 'bar']
>>> "foo  bar".split()
['foo', 'bar']

```

ralphbean (dismissed):

pyxis/create_container_image.py
mmalina: Good idea, fixed. Also, streamlined the way "latest" tag is added. Please check now.

johnbieren (approved):

ralphbean (approved):

happybhati (approved):

PR #88, opened 2023-11-03
chore: add gh CLI to image

ernesgonzalez33:
Needed for the create-github-release task in the release-service-catalog. To be added in this PR: https://github.com/redhat-appstudio/release-service-catalog/pull/281

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,12 +5,14 @@ ARG KUBECTL_VERSION=1.27.2
 ARG OCP_VERSION=4.14.0-rc.5
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
+ARG GH_VERSION=2.32.1
 
 RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
     curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl &&\
     curl -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_VERSION}/opm-linux.tar.gz |tar -C /usr/bin -xzf - &&\
     curl -L https://gitlab.com/gitlab-org/cli/-/releases/v${GLAB_VERSION}/downloads/glab_${GLAB_VERSION}_Linux_x86_64.tar.gz | tar -C /usr -xzf - bin/glab &&\
-    chmod +x /usr/bin/{yq,kubectl,opm,glab} &&\
+    curl -L https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_amd64.tar.gz  | tar -C /usr -xzf - --strip=1 gh_${GH_VERSION}_linux_amd64/bin/gh &&\
+    chmod +x /usr/bin/{yq,kubectl,opm,glab,gh} &&\
     rpm -ivh https://github.com/sigstore/cosign/releases/download/v${COSIGN_VERSION}/cosign-${COSIGN_VERSION}.x86_64.rpm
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -5,12 +5,14 @@ ARG KUBECTL_VERSION=1.27.2
...
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
+ARG GH_VERSION=2.32.1
 
 RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
     curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl &&\
     curl -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_VERSION}/opm-linux.tar.gz |tar -C /usr/bin -xzf - &&\
     curl -L https://gitlab.com/gitlab-org/cli/-/releases/v${GLAB_VERSION}/downloads/glab_${GLAB_VERSION}_Linux_x86_64.tar.gz | tar -C /usr -xzf - bin/glab &&\
-    chmod +x /usr/bin/{yq,kubectl,opm,glab} &&\
+    curl -L https://github.com/cli/cli/releases/download/v2.32.1/gh_${GH_VERSION}_linux_amd64.tar.gz -o gh_${GH_VERSION}_linux_amd64.tar.gz && tar -xzf gh_${GH_VERSION}_linux_amd64.tar.gz && mv gh_${GH_VERSION}_linux_amd64/bin/gh /usr/bin/gh &&\
```
mmalina, line 14: Why not directly pipe the tgz through tar like we do with ocp and glab? At the very least the tar file would need to be removed.

ernesgonzalez33: Because it is a directory. So piping wasn't working. But I can remove the tar file

mmalina: I think for glab it's also a directory. That's why we specify to only extract one file from it - `bin/glab`. Is there a reason why it wouldn't work like this in this case?

ernesgonzalez33: Changed now. Adding the `strip=1` removes the top dir that was causing me problems

mmalina: Cool!

Dockerfile
```diff
@@ -5,12 +5,14 @@ ARG KUBECTL_VERSION=1.27.2
...
 ARG YQ_VERSION=4.34.1
 ARG GLAB_VERSION=1.31.0
+ARG GH_VERSION=2.32.1
 
 RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_linux_amd64 -o /usr/bin/yq &&\
     curl -L https://dl.k8s.io/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl &&\
     curl -L https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OCP_VERSION}/opm-linux.tar.gz |tar -C /usr/bin -xzf - &&\
     curl -L https://gitlab.com/gitlab-org/cli/-/releases/v${GLAB_VERSION}/downloads/glab_${GLAB_VERSION}_Linux_x86_64.tar.gz | tar -C /usr -xzf - bin/glab &&\
-    chmod +x /usr/bin/{yq,kubectl,opm,glab} &&\
+    curl -L https://github.com/cli/cli/releases/download/v2.32.1/gh_${GH_VERSION}_linux_amd64.tar.gz  | tar -C /usr -xzf - --strip=1 gh_${GH_VERSION}_linux_amd64/bin/gh &&\
```
mmalina, line 14: Can we also use the variable for the `v2.32.1` dir here? Like `/v${GH_VERSION}/`?

mmalina (approved):

johnbieren (approved):

PR #91, opened 2023-11-09
refactor(RHTAPBUGS-919): reuse session in pyxis api calls

mmalina:
In my testing, when running this script against qa Pyxis with an sbom that has 113 components, this will yield around 27 % of script runtime reduction.

Note that I'm also working on batching of the component creation in Pyxis which will bring even more performance increase, but that will be another PR.

### DIFF
--- a/pyxis/pyxis.py
+++ b/pyxis/pyxis.py
@@ -9,14 +9,15 @@ import requests
 
 LOGGER = logging.getLogger("pyxis")
 
+session = None
 
-def _get_session(pyxis_url: str, auth_required: bool = True) -> requests.Session:
+
+def _get_session(auth_required: bool = True) -> requests.Session:
     """Create a Pyxis http session with auth based on env variables.
 
     Auth is optional and can be set to use either API key or certificate + key.
 
     Args:
-        url (str): Pyxis API URL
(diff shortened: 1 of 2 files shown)


pyxis/pyxis.py
```diff
@@ -9,14 +9,15 @@
 
 LOGGER = logging.getLogger("pyxis")
 
+session = None
 
-def _get_session(pyxis_url: str, auth_required: bool = True) -> requests.Session:
```
lslebodn, line 13: have you considered `@functools.cache`
* https://docs.python.org/3/library/functools.html#functools.cache

mmalina: Yeah, I thought of that. Do you think it's nicer?

lslebodn: Change would be one-liner :-) instead of many `global`
Just not sure how it would affect tests. So feel free to ignore if it would make testing more complicated.

mmalina: I guess since I already implemented it this way, it's probably fine for now. You're right that testing a function that has the cache annotation could be a bit tricky, but probably doable. Anyway, thanks for the suggestion!

davidmogar (approved):

PR #92, opened 2023-11-09
refactor(RHTAPBUGS-919): create components in batches

mmalina:
By creating sbom components in Pyxis in batches
of 5, the overall script runtime is reduced by around
47 % in my testing with an sbom with 113 components.

The batch size is configurable, but currently set to 5.

Note that this is a follow up of https://github.com/redhat-appstudio/release-service-utils/pull/91 which is merged now.

### DIFF
--- a/pyxis/upload_sbom.py
+++ b/pyxis/upload_sbom.py
@@ -241,9 +241,8 @@ def create_content_manifest_components(
         LOGGER.info(f"Adding component {i+1} to {i+len(batch)}")
         mutation = template.render(components=batch)
 
-        variables = {"id": content_manifest_id}
-        for j, component in enumerate(batch):
-            variables[f"input{j}"] = component
+        variables = {f"input{j}": component for j, component in enumerate(batch)}
+        variables["id"] = content_manifest_id
 
         body = {"query": mutation, "variables": variables}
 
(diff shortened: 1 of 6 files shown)


pyxis/upload_sbom.py
```diff
@@ -211,40 +214,48 @@ def create_content_manifest(graphql_api: str, image_id: str) -> str:
...
+    LOGGER.info(f"Creating {len(components)} components in Pyxis...")
+    template = get_template()
+    for i in range(0, len(components), batch_size):
+        batch = components[i : i + batch_size]
+        LOGGER.info(f"Adding component {i+1} to {i+len(batch)}")
+        mutation = template.render(components=batch)
+
+        variables = {"id": content_manifest_id}
+        for j, component in enumerate(batch):
+            variables[f"input{j}"] = component
```
lslebodn, line 246: Dictionary Comprehension would be simpler to understand.
and you could append "id" after that.

davidmogar: Agree.

davidmogar (approved):
Just a minor comment. Looks good.

pyxis/upload_sbom.py
mmalina: OK, changed.

davidmogar (approved):

PR #93, opened 2023-11-13
fix(RHTAPBUGS-972): add option source container enabled option for pyxis

johnbieren:
Add the ability to pass whether or not the source container image is enabled when creating ContainerImage entries in Pyxis. The default is false.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -86,6 +86,11 @@ def setup_argparser() -> Any:  # pragma: no cover
         required=True,
     )
     parser.add_argument(
+        "--source-container-enabled",
+        help="Were source container images published alongside the image?",
+        default="false",
+    )
+    parser.add_argument(
         "--media-type",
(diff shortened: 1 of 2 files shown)


pyxis/create_container_image.py
```diff
@@ -197,6 +202,7 @@ def create_container_image(args, parsed_data: Dict[str, Any]):
                 "repository": image_repo,
                 "push_date": date_now,
                 "tags": pyxis_tags,
+                "source_container_image_enabled": (args.source_container_enabled == "true"),
```
mmalina, line 205: This is wrong. There is no such field in the object: https://pyxis.engineering.redhat.com/docs/objects/ContainerImageRepo.html?tab=Fields

Note that this is different from [the Container Repository object](https://pyxis.engineering.redhat.com/docs/objects/ContainerRepository.html?tab=Fields) which is updated in the publish-pyxis-repository task.

johnbieren: Thanks, I was looking at ContainerRepository

PR #95, opened 2023-11-21
chore: add apache 2 license to repo

mmalina:
License was missing in this repo. Add Apache 2 license. This was brought up by mbasti, thanks!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
davidmogar, line 189: 🙃

LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      License. However, in accepting such obligations, You may act only
+      on Your own behalf and on Your sole responsibility, not on behalf
+      of any other Contributor, and only if You agree to indemnify,
+      defend, and hold each Contributor harmless for any liability
+      incurred by, or claims asserted against, such Contributor by reason
+      of your accepting any such warranty or additional liability.
+
+   END OF TERMS AND CONDITIONS
+
+   APPENDIX: How to apply the Apache License to your work.
```
davidmogar, line 178: 🙃

davidmogar (changes requested):

LICENSE
mmalina: I didn't realize you need to change something in the template. But FWIW, we have the same issue here :) https://github.com/redhat-appstudio/release-service-catalog/blob/main/LICENSE and here https://github.com/redhat-appstudio/release-service/blob/main/LICENSE

Fixed here.

LICENSE
mmalina: Hmm, what shall we do about this? I guess you're supposed to add that snippet to all your files ideally. But it's probably generally accepted to just have the license file in the repo and then it's implied that it's all covered by it. And here we also have that snippet left there: https://github.com/redhat-appstudio/infra-deployments/blob/main/LICENSE

So I'd say leave it there?

davidmogar: Can be omitted, but fair enough. Let's keep it.

davidmogar (approved):

PR #97, opened 2023-11-28
feat: Add sonar cloud properties

sawood14012:
This commit Adds sonar cloud properties for better control.

Signed-off-by msawood@redhat.com

https://issues.redhat.com/browse/RHTAPREL-713


.sonarcloud.properties
```diff
@@ -0,0 +1,7 @@
+# Path to sources
+sonar.sources=.
+sonar.exclusions=**/*test*, **/*.yaml,
```
johnbieren, line 3: do we want to exclude the zz_generated files like we do in internal-services?

johnbieren (approved):

PR #98, opened 2023-11-28
feat(RHTAPBUGS-959): add wait-for-ir script to enable parallel runs

mmalina:
The new wait-for-ir script will wait for one or more InternalRequests to complete. The IR(s) to wait for can either be specified by a name or using a label selector.

The internal-request script was modified to use this new script if used in sync mode (-s).

Note that this is needed for https://github.com/redhat-appstudio/release-service-catalog/pull/319 and is tested there.

### DIFF
--- a/utils/internal-request
+++ b/utils/internal-request
@@ -24,6 +24,9 @@
 #       InternalRequest resource. The value of the parameter is treated as a string,
 #       and it can be a valid JSON object or array. When passing complex parameter
 #       values, make sure to enclose them in quotes.
+#   -i  identifier: if set, a label will be added to the IR with key $LABEL_KEY
+#       (see below in the script) and value set to this string. Optional.
+#       E.g. a TaskRun UID.
 #   -s  Sync: a flag that indicates whether the script has to finish to complete
 #       the tasks or exit immediately after creating the resource. Default is true.
 #   -t  Timeout: Defaults to 600 seconds.
@@ -41,6 +44,8 @@
 set -e
 
 TIMEOUT=600
+LABEL_KEY="internal-services.appstudio.openshift.io/group-id"
+LABEL_VALUE=""
 SYNC=true
 PARAMS=""
 
@@ -51,6 +56,9 @@ function usage {
     echo "  -p  Params: can be specified multiple times. Each '-p' flag represents a"
     echo "      parameter that will be added to the 'parameters' field in the"
     echo "      InternalRequest resource."
+    echo "  -i  identifier: if set, a label will be added to the IR with key $LABEL_KEY"
+    echo "      and value set to this string. Optional."
+    echo "      E.g. a TaskRun UID."
     echo "  -s  Sync: a flag that indicates whether the script has to finish to complete the tasks or exit immediately after creating the resource. Default is true."
     echo "  -t  Timeout: Defaults to 600 seconds."
     echo "  -h  Display this help message."
@@ -60,11 +68,12 @@ function usage {
 
 # Parsing arguments
 PARAMS=() # initialize PARAMS as an empty array
-while getopts r:p:s:t:h flag
+while getopts r:p:i:s:t:h flag
 do
     case "${flag}" in
         r) REQUEST=${OPTARG};;
         p) PARAMS+=("${OPTARG}");; # append each parameter to the PARAMS array
+        i) LABEL_VALUE=${OPTARG};;
         s) SYNC=${OPTARG};;
         t) TIMEOUT=${OPTARG};;
         h) usage;;
@@ -109,6 +118,12 @@ PAYLOAD=$(jq -n \
       }
     }'
 )
+if [ -n $LABEL_VALUE ]; then
+    PAYLOAD=$(jq \
+        --arg label_key "$LABEL_KEY" \
+        --arg label_value "$LABEL_VALUE" \
+        '.metadata.labels += {($label_key): $label_value}' <<< $PAYLOAD)
+fi
 
 # Create InternalRequest using kubectl
 RESOURCE=$(echo "$PAYLOAD" | kubectl create -f - -o json)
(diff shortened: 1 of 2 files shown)


johnbieren (approved):
Seems fine to me. Maybe @theflockers can take a look too as the IR expert

utils/internal-request
```diff
@@ -41,6 +43,8 @@
 set -e
 
 TIMEOUT=600
+LABEL_KEY="appstudio.openshift.io/task-id"
```
davidmogar, line 46: This could be a bit generic. I'm afraid that any other team sets this label. Could we add a prefix maybe? Either `release` or `internal` would be just fine.

utils/internal-request
```diff
@@ -51,6 +55,8 @@ function usage {
     echo "  -p  Params: can be specified multiple times. Each '-p' flag represents a"
     echo "      parameter that will be added to the 'parameters' field in the"
     echo "      InternalRequest resource."
+    echo "  -l  label: Label value to use for selecting IRs to watch,"
```
davidmogar, line 58: Given that we actually force the label key, wouldn't it make more sense to call this `identifier`?

utils/internal-request
davidmogar, line 46: In the `-l` option you say that, for example, the value could be the `TaskRun UID` but here you actually say that this is a task id. I would just call it `id` or `identifier`.

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
+#!/bin/bash
+
+# This script waits for one or more InternalRequests to complete.
+#
+# The IRs to watch can be specified either by name or a label selector.
```
davidmogar, line 5: Not sure about this but shouldn't be:
```suggestion
# The IRs to watch can be specified either by name or by a label selector.
```

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+#
+# The script waits for the InternalRequests to reach a 'completed'
+# status and will provide an exit code as follows:
+# All succeeded: exit code 0
```
davidmogar, line 9: Can you use a list (like in Parameters) to increase readability?

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+# At least one IR failed or rejected: exit code 21
```
davidmogar, line 10: What's the reason behind using `21`?

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+# Timeout reached: exit code 1
+#
+# Usage:
+#   ./wait-for-ir [-n name] [-l label] [-t timeout]
+#
+# Parameters:
+#   -n  name: The name of the InternalRequest to watch.
+#   -l  label: Label value to use for selecting IRs to watch,
```
davidmogar, line 18: I'd call it `id` or `identifier`.

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+#       e.g. the TaskRun UID.
+#   -t  Timeout: Defaults to 600 seconds.
+#   -h  Display this help message.
+#
+#   One of -n or -l has to be specified, but not both.
```
davidmogar, line 23:
```suggestion
#  One of -n or -l has to be specified, but not both.
```

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+    usage
+fi
+
+# Wait until status is set for all IRs or timeout is reached
+END_TIME=$(date -ud "$TIMEOUT seconds" +%s)
+SUCCESS=true
+while true; do
+    echo Checking IR statuses...
+    if [ -n "$NAME" ]; then
+        # Wrap the single IR in an array
```
theflockers, line 89: Maybe change to `a JSON array`. At the first read i was looking for a bash array.

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+#   - jq: This script uses jq to parse JSON. It must be installed on the system
+#     running the script.
+#
+# Note: This script is intended to be used with a specific Kubernetes API
+# that includes the 'InternalRequest' resource type.
+
+set -e
+
+NAME=""
+LABEL_KEY="appstudio.openshift.io/task-id"
```
davidmogar, line 37: Same comment as before.

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+    echo "Usage: $0 [-n name] [-l label] [-t timeout]"
+    echo
+    echo "  -n  name: The name of the InternalRequest to watch."
+    echo "  -l  label: Label value to use for selecting IRs to watch,"
+    echo "      e.g. the TaskRun UID."
+    echo "  -t  Timeout: Defaults to 600 seconds."
+    echo "  -h  Display this help message."
+    echo
+    echo "  One of -n or -l has to be specified, but not both."
+    exit 1
```
davidmogar, line 51: This shouldn't be using the same error code than the timeout. Confusing.

utils/wait-for-ir
mmalina: ok, done

utils/wait-for-ir
```diff
@@ -0,0 +1,139 @@
...
+        if [ $SUCCESS = true ]; then
+            print_conditions "$IRS"
+            echo Result: success
+            exit 0
+        else
+            echo ERROR: At least one InternalRequest failed
+            print_conditions "$IRS"
+            echo Result: failure
+            exit 21
+        fi
```
theflockers, line 128: I guess the `Rejected` condition is missing.

utils/internal-request
mmalina: What do you mean "we force the label key"? I guess I could change it to `-l / identifier`. But it's not mandatory and it pairs with the `-l` in the other script where I think it makes sense - you choose IRs either by name or by label.

mmalina: Oh I see what you mean - that the label key is forced. That's true. Originally I thought this would be the full selector (key=value), but then I decided to set the key in the script. I don't mind either way. But I still think `identifier` is not a great name. It's the label value.

theflockers (commented):
How are we planning to control the amount of IRs created/monitored at once? Should it be controlled by the task itself?

utils/internal-request
mmalina: Yeah, I say task id. And one example of a task id is the uid ;) id is short for identification. I'm fine with changing it to `"release.appstudio.openshift.io/id"`. Would you prefer that? In reality it's actually an id for the taskrun - that's the scope that we want here. So I'd say it should actually be `"release.appstudio.openshift.io/task-run-id"` WDYT?

utils/wait-for-ir
mmalina: I don't think the preposition needs repeating if it's the same, but ok, changed.

utils/internal-request
mmalina: ok, added `release.` to the beginning.

utils/wait-for-ir
mmalina: This was used in the `internal-request` script, so I used the same. I guess the idea is to have something specific in case the user of this script wants to check for it - other exit codes could be anything else blowing up in the script. But I don't have a strong opinion on this, I'm fine to change it to whatever.

utils/wait-for-ir
mmalina: Are you sure? This note is related to parameters, that's why it's indented the same as the parameters.

utils/wait-for-ir
mmalina: I changed the timeout code to 124 as that's what's also used by the `timeout` command for timeout.

utils/wait-for-ir
mmalina: Done.

mmalina:
> How are we planning to control the amount of IRs created/monitored at once? Should it be controlled by the task itself?

@theflockers  Yes, it's handled in the task. See the related PR: https://github.com/redhat-appstudio/release-service-catalog/pull/319

utils/wait-for-ir
mmalina: Rejected is a failure. Anything that is not a success is a failure. And the script will print all the conditions at the end, so you can easily see what the actual condition was.

theflockers: I recall the first idea was matching the `InternalRequest` status with the exit code but looking now at all the tasks that uses the `internal-request` script we did not use this at all. So ok then.

mmalina: OK, cool, thanks for checking.

utils/internal-request
davidmogar: It's the label value, which would be the id you use. So you are actually passing an id, not a label.

utils/internal-request
mmalina: OK, I changed it to `-i / identifier`, please check. Although I think calling it "label" would make it more obvious.

mmalina: I changed this to `internal-services.appstudio.openshift.io/group-id` as suggested in another thread.

davidmogar: Feel free to ask others. This is just my opinion. I can be wrong :wink:

johnbieren: I am fine with label since it is relevant for the labels on a k8s resource. I am also fine with identifier... identifier and label are so similar in English it does not make much difference to me :slightly_smiling_face: It isn't reveal to end users anyways, just us, and we will probably just copy from whatever task uses this first when we add it to others, whether we call it --label or --identifier or --bigfoot

mmalina: OK, thanks. I'm fine with "-i" since I already changed it.

scoheb: > --bigfoot
classic

theflockers: Any of them are good to me too.

theflockers (approved):
LGTM

davidmogar (approved):

PR #109, opened 2024-01-02
chore: enable source builds

scoheb:
- EC policy now requires them

### DIFF
--- a/.tekton/release-service-utils-pull-request.yaml
+++ b/.tekton/release-service-utils-pull-request.yaml
@@ -31,8 +31,6 @@ spec:
     value: .
   - name: revision
     value: '{{revision}}'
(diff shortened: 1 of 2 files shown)


johnbieren (approved):
It's confusing that this is in the definition twice (like L34 in the pull-request pipeline def)

scoheb:
> It's confusing that this is in the definition twice (like L34 in the pull-request pipeline def)

That is a leftover. I am removed the reference on L34

johnbieren:
> > It's confusing that this is in the definition twice (like L34 in the pull-request pipeline def)
> 
> That is a leftover. I am removed the reference on L34

awesome, thank you

johnbieren (approved):

PR #113, opened 2024-01-11
feat(RHTAPBUGS-1077): allow setting multiple IR labels

johnbieren:
This changes the -i identifier flag to -l labels and allows for passing multiple labels instead of just one single identifier.


utils/internal-request
```diff
@@ -87,20 +86,30 @@ then
...
 for param in "${PARAMS[@]}"
 do
     KEY=$(echo "$param" | cut -d'=' -f1)
     VALUE=$(echo "$param" | cut -d'=' -f2-)
     PARAM_JSON_ARRAY+=("$(jq -n --arg key "$KEY" --arg value "$VALUE" '{($key): $value}')")
 done
 
+for label in "${LABELS[@]}"
+do
+    LABELS_ADDED="true"
```
mmalina, line 104: I think you don't need this and you can just use `[[ -n ${LABELS[@]} ]]` to check that it's non-empty.

johnbieren: Thanks, updated. As I wrote it I actually thought to myself hmm I wonder if Martin is going to have a more clever way to do this

davidmogar (approved):

PR #114, opened 2024-01-15
refactor: wait-for-ir accepts labels

johnbieren:
This commit refactors the wait-for-ir script to take a list of labels instead of accepting an identifier value.

### DIFF
--- a/utils/wait-for-ir
+++ b/utils/wait-for-ir
@@ -11,17 +11,17 @@
 #   - Timeout reached: exit code 124
 #
 # Usage:
-#   ./wait-for-ir [-n name] [-i identifier] [-t timeout]
+#   ./wait-for-ir [-n name] [-l labels] [-t timeout]
 #
 # Parameters:
 #   -n  name: The name of the InternalRequest to watch.
-#   -i  identifier: If set, the IRs to watch will be determined by a label
-#       of this value and key $LABEL_KEY (see below in the script). Optional.
-#       E.g. a TaskRun UID.
+#   -l  labels: Can be specified multiple times. If set, the
+#       IRs to watch will be determined by a label of this key
+#       and value. Optional. E.g. a TaskRun UID
 #   -t  Timeout: Defaults to 600 seconds.
 #   -h  Display this help message.
 #
-#   One of -n or -i has to be specified, but not both.
+#   One of -n or -l has to be specified, but not both.
 #
 # Prerequisites:
 #   - kubectl: The Kubernetes command line tool must be installed and properly
@@ -35,17 +35,15 @@
 set -e
 
 NAME=""
-LABEL_KEY="internal-services.appstudio.openshift.io/group-id"
-LABEL_VALUE=""
 TIMEOUT=600
 
 function usage {
-    echo "Usage: $0 [-n name] [-i identifier] [-t timeout]"
+    echo "Usage: $0 [-n name] [-l labels] [-t timeout]"
     echo
     echo "  -n  name: The name of the InternalRequest to watch."
-    echo "  -i  identifier: If set, the IRs to watch will be determined by a label"
-    echo "      of this value and key $LABEL_KEY. Optional."
-    echo "      E.g. a TaskRun UID."
+    echo "  -l  labels: Can be specified multiple times. If set, the"
+    echo "      IRs to watch will be determined by a label of this key"
+    echo "      and value. Optional. E.g. a TaskRun UID"
     echo "  -t  Timeout: Defaults to 600 seconds."
     echo "  -h  Display this help message."
     echo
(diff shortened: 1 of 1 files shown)


utils/wait-for-ir
```diff
@@ -65,23 +63,34 @@ function print_conditions {
...
+if [[ -z "$NAME" && -z ${LABELS[@]} || -n $NAME && -n ${LABELS[@]} ]]
 then
     usage
 fi
 
+# Create a label string to pass to the kubectl get command
+LABEL_STRING=""
+for label in "${LABELS[@]}"
+do
+    KEY=$(echo "$label" | cut -d'=' -f1)
```
mmalina, line 88: Do we need this parsing? Couldn't we just concatenate all the items in `LABELS` using `,` as a separator? In fact, if the argument parsing had this, all we would need to do is remove the final `,`:
```sh
LABELS=""
...
l) LABELS+="${OPTARG},";; 
```

The upside is that it's much simpler. But an argument could be made that it makes sense to copy&paste what's in the other script. So up to you, just an idea.

johnbieren: With that change, would I still pass like `-l key=value -l secondkey=secondvalue`? If we are going to change `wait-for-ir` to be like `internal-request`, they should be passed the same way, which is that way. If it is just changing the manipulation code inside the script, that is fine

mmalina: Yes. Of course the easiest to implement would be to support `-l key=value,secondkey=secondvalue` and then we wouldn't need any manipulation at all.

But what I'm suggesting is to stay consistent with `internal-request`, so `-l key=value -l secondkey=secondvalue`. In `internal-request` you need to construct the json from those keys and values. But here we only need a comma-separated list of `key=value` strings. So no need to parse the keys and values parts separately.

johnbieren: Got it. That is much simpler. Updated, thank you, bash master

utils/wait-for-ir
```diff
@@ -65,23 +63,26 @@ function print_conditions {
...
 done
 
 # Check that name or label is set, but not both
-if [[ -z "$NAME" && -z $LABEL_VALUE || -n $NAME && -n $LABEL_VALUE ]]
+if [[ -z "$NAME" && -z ${LABELS[@]} || -n $NAME && -n ${LABELS[@]} ]]
 then
     usage
 fi
 
+LABELS=$(echo "${LABELS}" | rev | cut -c2- | rev) # remove final ,
```
mmalina, line 84: I'd suggest using `LABELS=${LABELS%,}`

```
${string%substring}
Deletes shortest match of $substring from back of $string.
```

utils/wait-for-ir
```diff
@@ -65,23 +63,26 @@ function print_conditions {
...
+        l) LABELS+="${OPTARG},";; # append each label to the LABELS array
         t) TIMEOUT=${OPTARG};;
         h) usage;;
         *) usage;;
     esac
 done
 
 # Check that name or label is set, but not both
-if [[ -z "$NAME" && -z $LABEL_VALUE || -n $NAME && -n $LABEL_VALUE ]]
+if [[ -z "$NAME" && -z ${LABELS[@]} || -n $NAME && -n ${LABELS[@]} ]]
```
mmalina, line 79: `$LABELS` will be just a normal string now, so we want `-z "$LABELS"` and `-n "$LABELS"` here.

utils/wait-for-ir
```diff
@@ -65,23 +63,26 @@ function print_conditions {
 }
 
 # Parsing arguments
-while getopts n:i:t:h flag
+LABELS=() # initialize LABELS as an empty array
```
mmalina, line 66: This should be `LABELS="" # initialize LABELS as an empty string` (not sure if we need the comment though)

utils/wait-for-ir
johnbieren: I'm surprised google didn't have that pop up. Much nicer. Updated

utils/wait-for-ir
johnbieren: Good catch

utils/wait-for-ir
johnbieren: Good catch

utils/wait-for-ir
mmalina: Added a few more comments.

But playing with this, I realized that using multiple label selectors means all of them need to be satisfied. So it's not very likely we will be watching more than one at once, right? Meaning that it would be sufficient to support just one. But then again supporting multiple is consistent with the other scripts. And it's implemented already. So I guess it's fine. Just thinking out loud.

johnbieren: Yeah, it does a logical and, so I would be very surprised if we use this often with more than one label specified. However, as you allude to, I think it is nice to be in sync with the other script

mmalina (approved):

PR #119, opened 2024-02-06
fix(RHTAPREL-816): don't fail translate-delivery-repo

johnbieren:
Some tests call the translation even when they aren't using a quay.io/redhat-[prod,pending] repo. As the old behavior did not fail in these cases, this commit changes translate-delivery-repo to also not fail in these scenarios.

### DIFF
--- a/utils/translate-delivery-repo
+++ b/utils/translate-delivery-repo
@@ -14,13 +14,13 @@ REPO=$1
 
 if [ -z "${REPO}" ]; then
     echo -e "Please pass a repo to translate like 'quay.io/redhat-prod/product----repo'"
-    exit
+    exit 1
 fi
 
 if [[ "${REPO}" != "quay.io/redhat-prod"* && "${REPO}" != "quay.io/redhat-pending"* ]] ||
     [[ "${REPO}" != *"----"* ]] ; then
-    echo Malformatted repo to translate. Expected format is quay.io/redhat-[prod,pending]/product----repo
-    exit 1
+    echo -n "Warning: Repo to translate is not in expected format. Expected format is " >&2
+    echo "quay.io/redhat-[prod,pending]/product----repo" >&2
 fi
 
 REPO=${REPO/quay.io\/redhat-prod/registry.redhat.io}


utils/translate-delivery-repo
```diff
@@ -17,12 +17,6 @@ if [ -z "${REPO}" ]; then
     exit
```
mmalina, line 17: Here it should most likely be `exit 1`, especially since otherwise you'd have the string above as the result of this script, which would most definitely cause things to break.

utils/translate-delivery-repo
```diff
@@ -17,12 +17,6 @@ if [ -z "${REPO}" ]; then
...
 fi
 
-if [[ "${REPO}" != "quay.io/redhat-prod"* && "${REPO}" != "quay.io/redhat-pending"* ]] ||
-    [[ "${REPO}" != *"----"* ]] ; then
-    echo Malformatted repo to translate. Expected format is quay.io/redhat-[prod,pending]/product----repo
```
mmalina, line 22: Maybe we could still leave this as a warning? `echo Warning: Repo to translate is not in expected format ...  >&2`

But I'm fine with removing it altogether as well.

utils/translate-delivery-repo
johnbieren: Fixed

utils/translate-delivery-repo
johnbieren: I think that is a good idea. Added

utils/translate-delivery-repo
```diff
@@ -14,13 +14,13 @@ REPO=$1
...
-    exit
+    exit 1
 fi
 
 if [[ "${REPO}" != "quay.io/redhat-prod"* && "${REPO}" != "quay.io/redhat-pending"* ]] ||
     [[ "${REPO}" != *"----"* ]] ; then
-    echo Malformatted repo to translate. Expected format is quay.io/redhat-[prod,pending]/product----repo
-    exit 1
+    echo -n Warning: Repo to translate is not in expected format. Expected format is >&2
+    echo quay.io/redhat-[prod-pending]/product----repo >&2
```
mmalina, line 23: Can we wrap the string in `""`? `[]` have special meaning in bash and by default bash will leave the string as is if there is no match, but it's still not nice I think.

utils/translate-delivery-repo
mmalina: Thanks.

utils/translate-delivery-repo
mmalina: Also, originally you had `[prod,pending]`, not `[prod-pending]` - the comma is probably more correct.

johnbieren: Yeah, nice catch. Thanks

johnbieren: Fixed

mmalina (approved):

PR #121, opened 2024-02-08
feat(RHTAPREL-818): add product_[name,version] to advisory template

johnbieren:


johnbieren:
Moving to draft as there is an unresolved comment on the Jira about a second new field

happybhati (approved):

PR #127, opened 2024-02-28
fix(RHTAPREL-860): git_clone_and_checkout works on other branches

johnbieren:


utils/git-functions
```diff
@@ -67,9 +67,8 @@ git_clone_and_checkout() {
 
     # rewrites the repo url with the auth token
     CLONE_URL="https://oauth2:${ACCESS_TOKEN}@${REPOSITORY#*://}"
-    git clone --depth 1 "${CLONE_URL}"
+    git clone --branch "${REVISION}" --single-branch "${CLONE_URL}"
```
mmalina, line 70: Why not keep `--depth 1` instead of `--single-branch`?

From `man git-clone`:
```
       --depth <depth>
           Create a shallow clone with a history truncated to the specified number of commits.
           Implies --single-branch unless --no-single-branch is given to fetch the histories near
           the tips of all branches. [..]
```
I tried it and it seems to work.
```suggestion
    git clone --depth 1 --branch "${REVISION}" "${CLONE_URL}"
```

johnbieren: updated

mmalina (approved):

PR #129, opened 2024-03-06
feat(RHTAPREL-799): add ship_date parameter to advisory template

mmalina:
This will be populated when creating the advisory.


mmalina:
> LGTM Can you update the Jira to `s/shipDate/ship_date/'? It makes more sense in this format IMO because the other fields are like this, but the Jira uses camelCase. Just in case the consuming teams are following the Jira

Good point. I edited the Jira to use ship_date and updated_date and I also edited my question in SECDATA-428 .

I'll probably wait for their response before merging this.

PR #133, opened 2024-03-25
fix: handle null Env from skopeo

ralphbean:
The pyxis schema can only accept an array here, not None, however skopeo can return null/None. Here, if the value is falsy, then use an empty array rather than the falsy value.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -151,7 +151,7 @@ def prepare_parsed_data(skopeo_result: Dict[str, Any]) -> Dict[str, Any]:
         "layers": skopeo_result.get("Layers", []),
         "name": skopeo_result.get("Name", ""),
         "architecture": skopeo_result.get("Architecture", ""),
-        "env_variables": skopeo_result.get("Env", []),
+        "env_variables": skopeo_result.get("Env", []) or [],
     }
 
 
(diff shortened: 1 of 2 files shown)


ralphbean:
You can see an example null value with:

```
skopeo inspect --no-tags docker://quay.io/redhat-user-workloads/centos-bootc-tenant/centos-bootc/centos-bootc:on-pr-8238f083a86b60632f23e111f080b940457016b1 | jq .Env
```

scoheb (approved):

pyxis/create_container_image.py
```diff
@@ -151,7 +151,7 @@ def prepare_parsed_data(skopeo_result: Dict[str, Any]) -> Dict[str, Any]:
         "layers": skopeo_result.get("Layers", []),
         "name": skopeo_result.get("Name", ""),
         "architecture": skopeo_result.get("Architecture", ""),
-        "env_variables": skopeo_result.get("Env", []),
+        "env_variables": skopeo_result.get("Env", []) or [],
```
davidmogar, line 154: I see that the PR is already merged but, for my own education, why does this work? `get` was already setting the default to `[]` if the key was not found.

davidmogar: Ok, I'm answering myself. `None` is not the same that `not found` so whenever `Env` is `None`, no default was set.

PR #134, opened 2024-03-25
fix: accept both index and manifest digests when creating pyxis records

ralphbean:
We need both when creating a multi-arch image. Both manifest_schema2_digest and manifest_list_digest should be supplied.

### DIFF
--- a/utils/get-image-architectures
+++ b/utils/get-image-architectures
@@ -2,8 +2,8 @@
 #
 # script:      get-image-architectures
 # 
-# description: This script fetches all architectures for a passed image.
-#              It will output each arch on its own line.
+# description: This script fetches all architectures and their digests for a passed image.
+#              It will output each arch on its own line, as json.
 #
 # example command:  
 #              get-image-architectures IMAGE where IMAGE is quay.io/org/repo@sha256:abcde for example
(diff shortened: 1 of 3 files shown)


ralphbean:
When merged, this change needs to be promoted alongside https://github.com/redhat-appstudio/release-service-catalog/pull/388

utils/get-image-architectures
```diff
@@ -2,8 +2,8 @@
 #
 # script:      get-image-architectures
 # 
-# description: This script fetches all architectures for a passed image.
-#              It will output each arch on its own line.
+# description: This script fetches all architectures and their digests for a passed image.
+#              It will output each arch on its own line, base64 encoded.
```
johnbieren, line 6: Why are we base64 encoding digests?

davidmogar: Not sure that I understand correctly, but you can for sure iterate over lines with bash. You could do something like the following:
```
~ % cat lines 
{"line": 1, "foo": "bar", "baz": "qux"}
{"line": 2, "foo": "bar", "baz": "qux"}
{"line": 3, "foo": "bar", "baz": "qux"}
{"line": 4, "foo": "bar", "baz": "qux"}
{"line": 5, "foo": "bar", "baz": "qux"}
~ % cat script.sh 
#!/bin/bash

cat lines | while IFS= read -r line; do
    echo "Processing line $(echo $line | jq '.line')"
done
~ % ./script.sh 
Processing line 1
Processing line 2
Processing line 3
Processing line 4
Processing line 5
```

ralphbean: Ack, will refactor to use that model.

ralphbean: Done in [09cfc9c](https://github.com/redhat-appstudio/release-service-utils/pull/134/commits/09cfc9cdd0d27422ec78d0fb7d70bca7c2cb078e) and https://github.com/redhat-appstudio/release-service-catalog/pull/388

johnbieren (approved):
needs rebase but lgtm

ralphbean:
@johnbieren rebased on main.

johnbieren (approved):

PR #143, opened 2024-05-01
feat(KFLUXBUGS-1115): add upload_rpm_manifest python util

johnbieren:
This commit adds a utility to the pyxis directory for uploading RPM
Manifests via graphql queries.

### DIFF
--- a/pyxis/upload_rpm_manifest
+++ b/pyxis/upload_rpm_manifest
@@ -1,0 +1,1 @@
+upload_rpm_manifest.py
--- a/pyxis/upload_rpm_manifest.py
+++ b/pyxis/upload_rpm_manifest.py
@@ -1,0 +1,180 @@
+#!/usr/bin/env python3
+"""Upload rpm manifest to Pyxis
+
+This script will take Pyxis image ID and an sbom cyclonedx file
+on the input. It will inspect the sbom for the rpms and then push
+data into Pyxis. If an RPM Manifest already exists for the container
+image, nothing is done as we assume it was already pushed by this
+script.
+
+Required env vars:
+PYXIS_KEY_PATH
+PYXIS_CERT_PATH
+
+Optional env vars:
+PYXIS_GRAPHQL_API
+"""
+import logging
+import string
+import os
+from pathlib import Path
+import time
+from packageurl import PackageURL
+from upload_sbom import load_sbom_components, parse_arguments
+
+import pyxis
+
+LOGGER = logging.getLogger("upload_rpm_manifest")
+
+
+def upload_container_rpm_manifest_with_retry(
+    graphql_api: str,
+    image_id: str,
+    sbom_path: str,
+    retries: int = 3,
+    backoff_factor: float = 5.0,
+):
+    """Call the upload_container_rpm_manifest function with retries"""
+    last_err = RuntimeError()
+    for attempt in range(retries):
+        try:
+            time.sleep(backoff_factor * attempt)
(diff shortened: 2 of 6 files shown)


mmalina:
PR is missing description. Can you add some more details? Maybe copying the commit message would be enough?

pyxis/upload_rpm_manifest.py
```diff
@@ -0,0 +1,178 @@
...
+    one RpmsItems per rpm component. A list is then formed of them
+    and returned to be used in a containerImageRPMManifest.
+    """
+
+    rpms_items = []
+    for component in components:
+        if "purl" in component:
+            purl_dict = PackageURL.from_string(component["purl"]).to_dict()
+            if purl_dict["type"] == "rpm":
+                rpm_item = {"name": purl_dict["name"]}
```
mmalina, line 137: What about the other fields of [RpmsItems](https://pyxis.engineering.redhat.com/docs/objects/RpmsItems.html?tab=Fields)? `nvra` could be composed I guess. `summary` could be taken from the component description maybe? But I guess you consulted the Pyxis team on this? So probably ok.

pyxis/upload_rpm_manifest.py
```diff
@@ -0,0 +1,178 @@
...
+"""
+    variables = {"id": "konflux-" + image_id, "input": {"image_id": image_id, "rpms": rpms}}
+    body = {"query": mutation, "variables": variables}
+
+    data = pyxis.graphql_query(graphql_api, body)
+
+    return data["create_image_rpm_manifest"]["data"]["_id"]
+
+
+def create_rpm_items(components: list[dict]) -> list[dict]:
```
mmalina, line 124: I wonder if it's just me, but when I see `create_` in this context I automatically expect it will create something in Pyxis - like `create_image_rpm_manifest`. So I'd go with something like `prepare_` or `construct_`. But it's fine to keep it as is if you don't feel this way.

pyxis/upload_rpm_manifest.py
johnbieren: I asked on slack about the fields and never got any response. None are required, so I guess what I added is sufficient. I don't see any summary in the component section in the sbom :man_shrugging:

pyxis/upload_rpm_manifest.py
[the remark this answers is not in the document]
johnbieren: Yeah I don't like it at all, but I added it for consistency

pyxis/upload_rpm_manifest.py
johnbieren: Changed to construct_rpm_items

pyxis/upload_rpm_manifest.py
mmalina: Hmm, that's unfortunate. I think we should really get some validation from them before merging this, because otherwise there's a high risk that it won't properly and will need fixing later. I pinged a few of them in your slack thread.

pyxis/upload_rpm_manifest
mmalina, line 1: This got me confused. I was like: Is this a text file? When I select ... -> View File, it does say symlink, but in the diff view it's not clear at all. Anyway, all is good, just wanted to share my struggles :D

mmalina (approved):

pyxis/upload_rpm_manifest
johnbieren: More reason to stop with the sym links :slightly_smiling_face:

PR #152, opened 2024-05-29
fix(KFLUXBUGS-1294): add cleanup_tags script for cleaning up tags

mmalina:
This script will take a Pyxis Container Image ID as input and it will remove all its tags from all previous images that use the Red Hat registry and the same repository and architecture.

### DIFF
--- a/pyxis/cleanup_tags.py
+++ b/pyxis/cleanup_tags.py
@@ -1,0 +1,360 @@
+#!/usr/bin/env python3
+"""
+Python script to clean up tags from previous Container Image objects in Pyxis
+
+It takes a Pyxis image id for input and it will check the tags this image has
+and then it will look for all other images with the same arch and repo
+and it will remove all the tags that are present in the image provided as input.
+
+Required env vars:
+PYXIS_KEY_PATH
+PYXIS_CERT_PATH
+
+Optional env vars:
+PYXIS_GRAPHQL_API
+"""
+import argparse
+import logging
+import string
+import os
+import time
+from typing import List, Dict, Any
+
+import pyxis
+
+LOGGER = logging.getLogger("cleanup_tags")
+
+
+def parse_arguments() -> argparse.Namespace:  # pragma: no cover
+    """Parse CLI arguments
+
+    :return: Dictionary of parsed arguments
+    """
+
+    parser = argparse.ArgumentParser(description="Clean up tags from previous images in Pyxis")
+
+    parser.add_argument(
+        "--pyxis-graphql-api",
+        default=os.environ.get(
+            "PYXIS_GRAPHQL_API", "https://graphql-pyxis.api.redhat.com/graphql/"
+        ),
+        help="Pyxis Graphql endpoint",
+    )
+    parser.add_argument(
+        "image_id",
+        help="Pyxis Container Image ID",
+    )
+    parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
+    parser.add_argument(
+        "--retry",
+        "-r",
+        action="store_true",
+        help="If set, retry the upload in case it fails",
+    )
+
+    return parser.parse_args()
+
+
+def cleanup_tags_with_retry(
+    graphql_api: str,
+    image_id: str,
+    retries: int = 3,
+    backoff_factor: float = 5.0,
+):
+    last_err = RuntimeError()
+    for attempt in range(retries):
+        try:
+            time.sleep(backoff_factor * attempt)
+            cleanup_tags(graphql_api, image_id)
+            return
+        except RuntimeError as e:
+            LOGGER.warning(f"Attempt {attempt+1} failed.")
+            last_err = e
+    LOGGER.error("Out of attempts. Raising the error.")
+    raise last_err
+
+
+def cleanup_tags(graphql_api, image_id: str):
+    image = get_image(graphql_api, image_id)
+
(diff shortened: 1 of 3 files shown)


pyxis/cleanup_tags.py
```diff
@@ -0,0 +1,359 @@
+#!/usr/bin/env python3
+"""
+Python script to clean up tags from previous Container Image objects in Pyxis
+
+It takes a Pyxis image id for input and it will check the tags this image has
+and then it will look for all other images with the same arch and repo
+and it will remove all the tags that are present in the image provide as input.
```
johnbieren, line 7: `s/provide/provided/`

pyxis/cleanup_tags.py
```diff
@@ -0,0 +1,359 @@
...
+
+    parser.add_argument(
+        "--pyxis-graphql-api",
+        default=os.environ.get(
+            "PYXIS_GRAPHQL_API", "https://graphql-pyxis.api.redhat.com/graphql/"
+        ),
+        help="Pyxis Graphql endpoint.",
+    )
+    parser.add_argument(
+        "image_id",
```
johnbieren, line 44: was leaving off the `--` intentional?

pyxis/cleanup_tags.py
```diff
@@ -0,0 +1,359 @@
...
+    """
+
+    parser = argparse.ArgumentParser(description="Clean up tags from previous images in Pyxis")
+
+    parser.add_argument(
+        "--pyxis-graphql-api",
+        default=os.environ.get(
+            "PYXIS_GRAPHQL_API", "https://graphql-pyxis.api.redhat.com/graphql/"
+        ),
+        help="Pyxis Graphql endpoint.",
```
johnbieren, line 41: no other help has a period, let's remove it here

pyxis/cleanup_tags.py
```diff
@@ -0,0 +1,359 @@
...
+    variables = {"id": image_id}
+    body = {"query": query, "variables": variables}
+
+    data = pyxis.graphql_query(graphql_api, body)
+    image = data["get_image"]["data"]
+
+    return image
+
+
+def get_image_properties(image: Dict):
```
johnbieren, line 140: I think it would be nice to rename to something like `get_registry_access_image_properties` or something that says this is only for registry.access.redhat.com

pyxis/cleanup_tags.py
```diff
@@ -0,0 +1,359 @@
...
+    For each image in `images` it will remove all `tags`
+    from its repositories and then it will update the image
+    using graphql mutation update_image.
+    """
+    for image in images.values():
+        LOGGER.info(f"Updating image {image['_id']} with architecture {image['architecture']}")
+        LOGGER.info("Repositories and tags before update:")
+        for repository in image["repositories"]:
+            repo_tags = [tag["name"] for tag in repository["tags"]]
+            LOGGER.info(f"  {repository['registry']}/{repository['repository']}: {repo_tags}")
```
johnbieren, line 257: love this

pyxis/cleanup_tags.py
mmalina: Fixed, thanks.

pyxis/cleanup_tags.py
mmalina: Yes, arguments prefixed with `-` are optional, while this one is mandatory. See here: https://docs.python.org/3/library/argparse.html#name-or-flags 
> When [parse_args()](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.parse_args) is called, optional arguments will be identified by the - prefix, and the remaining arguments will be assumed to be positional:

pyxis/cleanup_tags.py
mmalina: Not sure how that happened :) Removed.

pyxis/cleanup_tags.py
mmalina: Good point. I renamed it to `get_rh_registry_image_properties` and then say "Get the registry.access.redhat.com repository properties of the image" in the description which is seemingly inconsistent - registry vs repository - but it kind of makes sense to me. Let me know if you don't like it.

johnbieren: I am good with it (and feel free to resolve this if it is the case) as long as you don't foresee having to do the same for `registry.redhat.io`

johnbieren (approved):
Approving pending the one open conversation.
I didn't test it, but I am sure you did and the code looks good to me reading it over

pyxis/cleanup_tags.py
mmalina: Cool. There is no `registry.redhat.com` repository in the container image objects. See this note for more details on how this works: https://github.com/konflux-ci/release-service-utils/blob/main/pyxis/create_container_image.py#L5
Basically, the `registry.access.redhat.com` entry is the one that makes the image available from `registry.redhat.io`.

mmalina:
> Approving pending the one open conversation. I didn't test it, but I am sure you did and the code looks good to me reading it over

Thanks for the review. Yeah, I tested with some Container Image objects that I created in stage for testing. But ideally I'd like to confirm with a real world scenario. I should get a chance soon when they release a new bootc image here: https://issues.redhat.com/browse/RHEL-38544 But I will probably merge this on Monday regardless of their timing. And if there is a bug, I will fix it afterwards.

PR #163, opened 2024-06-24
chore: include oras in utils image

ralphbean:
I'd like to use it to run `oras resolve` to get the top-level digest of OCI artifacts that are not images.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,4 @@
+FROM quay.io/konflux-ci/oras:latest@sha256:04cbd5efbcf1a38944b050e3adbc3071218643f6aa92e95d5d25b173574bca5e as oras
 FROM registry.access.redhat.com/ubi8/ubi:8.10-901.1717584420
 
 ARG COSIGN_VERSION=2.1.1
@@ -14,6 +15,9 @@ RUN curl -L https://github.com/mikefarah/yq/releases/download/v${YQ_VERSION}/yq_
     curl -L https://github.com/cli/cli/releases/download/v${GH_VERSION}/gh_${GH_VERSION}_linux_amd64.tar.gz  | tar -C /usr -xzf - --strip=1 gh_${GH_VERSION}_linux_amd64/bin/gh &&\
     chmod +x /usr/bin/{yq,kubectl,opm,glab,gh} &&\
     rpm -ivh https://github.com/sigstore/cosign/releases/download/v${COSIGN_VERSION}/cosign-${COSIGN_VERSION}.x86_64.rpm
+
+COPY --from=oras /usr/bin/oras /usr/bin/oras
+COPY --from=oras /usr/local/bin/select-oci-auth /usr/local/bin/select-oci-auth
 
 RUN dnf -y --setopt=tsflags=nodocs install \
     git \
--- a/utils/get-image-architectures
+++ b/utils/get-image-architectures
@@ -23,8 +23,9 @@ if [ "$ARTIFACT_TYPE" != "null" ] ; then
     # OCI artifact (not a normal image)
     digest=$(echo $IMAGE | awk -F '@' '{print $2}')
     if [[ ! "$digest" =~ ^sha256: ]] ; then
-        echo "Could not determine digest for $IMAGE of artifactType ${ARTIFACT_TYPE}"
-        exit 1;
+        AUTH_FILE=$(mktemp)
+        select-oci-auth $IMAGE > $AUTH_FILE
+        digest=$(oras resolve --registry-config "$AUTH_FILE" "$IMAGE")
     fi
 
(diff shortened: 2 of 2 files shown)


ralphbean:
In particular, I need to update a bunch of `skopeo inspect` calls in the release-service-catalog to use this.

Dockerfile
```diff
@@ -1,20 +1,24 @@
+FROM quay.io/konflux-ci/oras:latest@sha256:04cbd5efbcf1a38944b050e3adbc3071218643f6aa92e95d5d25b173574bca5e as oras
+FROM quay.io/konflux-ci/yq:latest@sha256:974dea6375ee9df561ffd3baf994db2b61777a71f3bcf0050c5dca91ac9b3430 as yq
```
mmalina, line 2: Question: Why do these specify the `latest` tag if they also include a digest reference? Isn't that redundant?

utils/get-image-architectures
```diff
@@ -23,8 +23,8 @@ if [ "$ARTIFACT_TYPE" != "null" ] ; then
     # OCI artifact (not a normal image)
     digest=$(echo $IMAGE | awk -F '@' '{print $2}')
     if [[ ! "$digest" =~ ^sha256: ]] ; then
-        echo "Could not determine digest for $IMAGE of artifactType ${ARTIFACT_TYPE}"
-        exit 1;
+        select-oci-auth $IMAGE > auth.json
+        digest=$(oras resolve --registry-config auth.json $IMAGE)
     fi
 
     # Just report that the image is for linxu/amd64, which is not exactly true - but,
```
mmalina, line 30: Can you please fix the typo here? linxu -> linux

utils/get-image-architectures
```diff
@@ -23,8 +23,8 @@ if [ "$ARTIFACT_TYPE" != "null" ] ; then
     # OCI artifact (not a normal image)
     digest=$(echo $IMAGE | awk -F '@' '{print $2}')
     if [[ ! "$digest" =~ ^sha256: ]] ; then
-        echo "Could not determine digest for $IMAGE of artifactType ${ARTIFACT_TYPE}"
-        exit 1;
+        select-oci-auth $IMAGE > auth.json
```
mmalina, line 26: We had an issue in the past where we couldn't write a file somewhere: https://issues.redhat.com/browse/KFLUXBUGS-1291

In this case this will really depend on the PWD at the time when this function is executed. It would probably be safer to use a temp file.
```sh
AUTH_FILE=$(mktemp)
select-oci-auth $IMAGE > $AUTH_FILE
digest=$(oras resolve --registry-config $AUTH_FILE $IMAGE)
```

johnbieren (commented):
The commit mentions oras, but not yq. Can you add some details about why we need this other version of yq?

Dockerfile
ralphbean: It is redundant. The digest takes precedence.

The tag is there as a hint to humans like us and to tools like renovatebot that it should follow the floating tag. If we ever want to update the digest to move forwards - we know what tag to check to find the newest, "latest" digest.

utils/get-image-architectures
ralphbean: Heh, yes, will do.

utils/get-image-architectures
ralphbean: Good call. Will do so.

ralphbean:
> The commit mentions oras, but not yq. Can you add some details about why we need this other version of yq?

Yeah, you know - I'll just remove the `yq` change. I think I got carried away.

Long-term, I'd like us to start producing our own builds of *all* of those tools and move all consumption to this `FROM ...` pattern. But, we can consider doing that separately, later, rather than cramming it into this change.

johnbieren (dismissed):

Dockerfile
mmalina: I see, thanks for the explanation.

mmalina (dismissed):

ralphbean:
I rebased this to resolve a conflict in the Dockerfile. I think this is ready for approval and merge whenever you're ready.

johnbieren (approved):

mmalina (approved):

ralphbean:
> The tasks after the build are timing out in the ITS. Not sure why

This may have been due to the pipeline service outage today. That resolved _just now_. Let's see.

johnbieren:
> /retest

It worked, but needs rebase now :slightly_smiling_face:

johnbieren:
> > /retest
> 
> It worked, but needs rebase now 🙂

I think we both rebased it at the same time and I may have messed things up.. I see two commits now

johnbieren:
> > > /retest
> > 
> > 
> > It worked, but needs rebase now 🙂
> 
> I think we both rebased it at the same time and I may have messed things up.. I see two commits now

I will just squash when I merge

PR #165, opened 2024-06-25
feat: adapt pyxis image creation for OCI artifacts

ralphbean:
The driving detail here is that "skopeo inspect" without --raw is not implemented for non-image OCI artifacts, forcing us to move to `oras`.

This also includes changes from https://github.com/konflux-ci/release-service-utils/pull/163.


mmalina:
@ralphbean can you also provide a corresponding PR for the release-service-catalog, so I can see how those new parameters will be supplied?

pyxis/create_container_image.py
```diff
@@ -144,19 +160,20 @@ def image_already_exists(args, digest: str) -> bool:
...
+    with open(args.oras_manifest_fetch) as json_file:
+        oras_manifest_fetch = json.load(json_file)
+
     return {
-        "digest": skopeo_result.get("Digest", ""),
-        "docker_version": skopeo_result.get("DockerVersion", ""),
-        "layers": skopeo_result.get("Layers", []),
-        "name": skopeo_result.get("Name", ""),
-        "architecture": skopeo_result.get("Architecture", ""),
-        "env_variables": skopeo_result.get("Env", []) or [],
```
mmalina, line 159: I'm not really sure what the purpose of the parsed_data in [the Container Image object](https://pyxis.engineering.redhat.com/docs/objects/ParsedData.html?tab=Fields) is. Is it ok to no longer include `env_variables` or `docker_version`? We could check with the Pyxis team.

pyxis/create_container_image.py
mmalina, line 1: I'm curious about why we need to support non-image OCI artifacts here at all. Can you talk about the use case? For example we'll have OCI artifacts for disk images that we will be releasing. But I don't think we will be creating Container Image objects in Pyxis for those. Or will we?

ralphbean:
> @ralphbean can you also provide a corresponding PR for the release-service-catalog, so I can see how those new parameters will be supplied?

Yes - I started working on it last night but ran into some issues. I'll mark this as draft until that is ready.

pyxis/create_container_image.py
ralphbean: That's a good point! I just assumed that of course we'd want to see them in catalog.redhat.com.

There is some behavior in the proxy at registry.redhat.io that is driven by data in pyxis, but I don't have the details on hand. We'll need _that_ to work, but perhaps only that.

pyxis/create_container_image.py
ralphbean: You know, I looked at the catalog UI last night and decided that they weren't used. I expect they were _once_ used, but are no longer displayed. 

In any case, we can table that until we determine if we need to create containerImage entities for these OCI artifacts or not at all.

pyxis/create_container_image.py
mmalina: But the disk images will be available for download via exodus cdn, not via registry.redhat.com. In the release pipeline, we will fetch the images from OCI images in quay using oras. But then we will push them using `pubtools-pulp push`. Or is there some other use case? Right now I don't have a reason to believe that Pyxis Container Images should be used for something other than actual container images.

ralphbean: Oh, yes - it's the independent models. There's a requirement to deliver machine learning models independently via the registry, not on cdn.

See also https://github.com/ralphbean/merlinite-poc and https://github.com/konflux-ci/build-definitions/blob/main/task/oci-copy/0.1/oci-copy.yaml and https://gitlab.cee.redhat.com/releng/konflux-release-data/-/blob/main/config/stone-prod-p01.wcfb.p1/product/ReleasePlanAdmission/rhel-ai/models-staging.yaml

pyxis/create_container_image.py
mmalina: I guess if you're confident this can be removed it's fine with me :)

mmalina (approved):

mmalina (approved):

johnbieren:
Failing black formatter

johnbieren (approved):

PR #172, opened 2024-07-01
fix(RELEASE-1000): prevent failure if no components key found

davidmogar:
It seems that bundles could no contain the components key in the sbom. This causes the upload_sbom script to fail whenever the components are loaded.


theflockers (approved):
I would at least create a ticket to add a test case for this. Other than that, LGTM.

theflockers (approved):
LGTM

PR #179, opened 2024-07-09
feat(RHELDST-24850): add pubtools-pulp-push wrapper script

rbikar:
This change adds and new script stored in dedicated directory.
This script will be used for pushing content for RELEASE-971.

Necessary updates to Dockerfile and requirements.txt were done
as well to allow the same execution conventions for new script
as there are already for `pyxis` scripts.

Signed-off-by: Robert Bikar <rbikar@redhat.com>

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -10,6 +10,11 @@ jobs:
   python:
     runs-on: ubuntu-latest
     steps:
+      - name: Install OS packages
+        # libkrb5-dev is required for gssapi dependency installation (krb5-config command)
+        run: |
+          sudo apt-get -y update
+          sudo apt-get install -y libkrb5-dev
       - name: Check out repo
         uses: actions/checkout@v3
       - name: Black Lint
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     krb5-devel \
     krb5-workstation \
     rsync \
+    gcc \
     && dnf clean all
 
 RUN curl -LO https://github.com/release-engineering/exodus-rsync/releases/latest/download/exodus-rsync && \
@@ -41,7 +42,12 @@ RUN curl -LO https://github.com/release-engineering/exodus-rsync/releases/latest
 RUN pip3 install jinja2 \
     jinja2-ansible-filters \
     packageurl-python \
-    pubtools-content-gateway 
+    pubtools-content-gateway \
+    pubtools-pulp \
+    pubtools-exodus
+
+# remove gcc, required only for compiling gssapi indirect dependency of pubtools-pulp via pushsource
+RUN dnf -y remove gcc
 
 ADD data/certs/2015-IT-Root-CA.pem data/certs/2022-IT-Root-CA.pem /etc/pki/ca-trust/source/anchors/
 RUN update-ca-trust
@@ -49,6 +55,7 @@ RUN update-ca-trust
 COPY pyxis /home/pyxis
 COPY utils /home/utils
 COPY templates /home/templates
+COPY pubtools-pulp-wrapper /home/pubtools-pulp-wrapper
 
 # It is mandatory to set these labels
 LABEL name="Konflux Release Service Utils"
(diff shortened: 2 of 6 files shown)


mmalina:
@rbikar just a small detail: The actual jira for adding the script is RHELDST-24850, right? So could you reference that one in the PR/commit title?

requirements.txt
```diff
@@ -3,3 +3,5 @@ requests
 jinja2
 jinja2-ansible-filters
 packageurl-python
+pubtools-pulp
+pubtools-exodus
```
mmalina, line 7: It seems an empty line at the end is missing.

Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     krb5-devel \
     krb5-workstation \
     rsync \
+    gcc \
```
mmalina, line 35: It gcc really needed? What needs it? It's quite big:
```
Total download size: 56 M
Installed size: 150 M
```

mmalina (commented):
Added a few comments.

Also, the python tests are failing.

Also, can you add a Signed-off-by: line in the commit message? Usually all you need for that is `--signoff` when doing `git commit`. And you will also need to sign the commit. Documentation is here: https://github.com/konflux-ci/release-service-catalog/blob/development/CONTRIBUTING.md#signing-commits

requirements.txt
rbikar: line added

Dockerfile
rbikar: withouth gcc local image build fails with
```
  error: command 'gcc' failed: No such file or directory
  ----------------------------------------
  ERROR: Failed building wheel for gssapi
ERROR: Could not build wheels for gssapi which use PEP 517 and cannot be installed directly
Failed to build gssapi

```

I guess right now best I can do is to remove gcc after installation, as it's not needed for running it.

rbikar:
@mmalina  
can you please re-run tests? I guess  it can't be re-run automatically because I am 1st time contributor.

Dockerfile
rbikar: Also if you want to save more space you may want to use `minimal` variant of ubi9 image.

requirements.txt
mmalina: Thanks.

Dockerfile
mmalina: Yeah, that is not a bad idea. I think it's worth considering in the future. I created a jira here: https://issues.redhat.com/browse/RELEASE-1008

mmalina:
> @mmalina can you please re-run tests? I guess it can't be re-run automatically because I am 1st time contributor.

I guess it was already approved by Johnny or somebody else, because the tests are running without me doing anything :)

mmalina:
> I don't know what that last check that is pending approval even is

You mean "Red Hat Konflux / release-service-utils-on-pull-request"? That's the build in Konflux of the image and also integration tests, no?

johnbieren:
> > I don't know what that last check that is pending approval even is
> 
> You mean "Red Hat Konflux / release-service-utils-on-pull-request"? That's the build in Konflux of the image and also integration tests, no?

Oh, before my ok-to-test it just said Red Hat Konflux

.github/workflows/python.yaml
```diff
@@ -10,6 +10,10 @@ jobs:
   python:
     runs-on: ubuntu-latest
     steps:
+      - name: Install OS packages
+        run: |
+          sudo apt-get -y update
+          sudo apt-get install -y libkrb5-dev
```
mmalina, line 16: (I thought I commented here already, but I guess not?)

Why is this needed? Can you maybe add a comment saying why it's needed? Interestingly I can run `pytest` on my system (macOS) without installing this - and I would be surprised if I had a similar thing installed. But this makes me worried that for some users `pytest` won't work without installing that package first.

I'm not saying we need to deal with this now, but just wanted to raise this.

Dockerfile
mmalina: I guess we'll have to live with this for now.

mmalina (dismissed):

.github/workflows/python.yaml
rbikar: It's because of the gssapi dependency - but this time the ubuntu workers miss this package so we need to install it manually (we also do the same in our repos). It's likely that libkrb5 pkgs are integral part or are pre-installed on different linux distros so it works out-of-the-box.

Error for ref: https://github.com/konflux-ci/release-service-utils/actions/runs/9854405125/job/27212388432

I'll add a little comment there.

mmalina: ok, thanks.

johnbieren (approved):

PR #188, opened 2024-07-18
feat(RELEASE-971): add plr timeouts to InternalRequests

johnbieren:
This commit modifies the internal-request script in the utils directory to accept timeout values for the pipeline, task, and finally blocks that will be passed to tekton when created the pipelineRun.


johnbieren:
Tested this locally and it works

happybhati (approved):

PR #204, opened 2024-07-30
feat(RELEASE-238): Update translate-delivery-repo script to handle new repository format

happybhati:
Changes:
- Updated the `translate-delivery-repo` script to translate repository references to the appropriate public registry references.

- The script now handles repositories in the `registry.redhat.io/product/repo` and `registry.stage.redhat.io/product/repo` formats.

- Translates production repos from `registry.redhat.io/product/repo` to `registry.access.redhat.com/product/repo`.

- Translates stage repos from `registry.stage.redhat.io/product/repo` to `registry.access.stage.redhat.com/product/repo`.

- Outputs a JSON string with two entries: one for `redhat.io` and one for `access.redhat.com`.

- Added validation to ensure the passed repo is in the expected format.

- Provided example usage in the script comments for clarity.

### DIFF
--- a/utils/translate-delivery-repo
+++ b/utils/translate-delivery-repo
@@ -3,40 +3,33 @@
 # script:      translate-delivery-repo
 # 
 # description: This script translates repo references to the proper public registry references.
-#              It turns quay.io/redhat-prod/product----repo into registry.redhat.io/product/repo for production repos and
-#              quay.io/redhat-pending/product----repo -> registry.stage.redhat.io/product/repo for stage repos. It outputs a
+#              It turns registry.redhat.io/product/repo into registry.access.redhat.com/product/repo for production repos and
+#              registry.stage.redhat.io/product/repo -> registry.access.stage.redhat.com/product/repo for stage repos. It outputs a
 #              json string with two entries. Each entry is of the form repo: [redhat.io, access.redhat.com],
 #              url: <translated-delivery-repo>. 
 #
 # example command:  
-#              translate-delivery-repo REPO where REPO is quay.io/redhat-prod/product----repo for example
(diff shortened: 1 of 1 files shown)


utils/translate-delivery-repo
mmalina, line 1: I think we should re-evaluate if we need registry.access.redhat.com entries at all. If not, we could get rid of this script completely now. If we still need those, we could either rename this to make it more in line with the new purpose - that it just gives you the registry.access repo url from the registry.redhat.io repo url.

But even if we still need it, wouldn't it make more sense to just also prepare this string in apply-mapping along with rh-registry-repo? It could be just another key for that. And then all the tasks could just use those two fields directly.

The reason I'm suggesting we might not need registry.access is that that repo is unauthenticated and nowadays all images that are pushed require auth by default. You need to get an exception from prodsec if you want to release something for registry.access. AFAIK, Konflux was only intending to release stuff for the authenticated registry. But there is still stuff like ubi images that will always be available from registry.access.redhat.com and maybe eventually we will release those in Konflux as well. So we would need to check with the architects.

So for now, I would probably suggest to keep it, but calculate the string directly in apply-mapping.

johnbieren: I agree with Martin here about just having the logic in apply-mapping and add a second new key to each component like you are already adding one. I do not think we should drop registry.access just yet, but the part about just adding this in apply-mapping is what I agree with. Once your work on this ticket is done @happybhati IIUC, all this translation will happen in apply-mapping and the other tasks will just use the proper key for each component and not need this script, so I think we will be able to remove this script eventually (I hope)

happybhati: That makes sense, thanks; I will make the necessary changes as mentioned above.

happybhati:
Closing as this is no longer needed.

PR #212, opened 2024-08-13
fix(KFLUXBUGS-1480): retry upload_container_rpm_manifest on HTTPErrors

johnbieren:
This commit modifies the upload_container_rpm_manifest_with_retry function to retry if the exception is an HTTPError with code 504. Other HTTPErrors will fail immediately, as is currently the case for all HTTPErrors.


mmalina (approved):
Perfect! LGTM!

PR #215, opened 2024-08-15
fix(MMENG-4169): update cosign to 2.4.0

ligangty:
We're trying to use release-service-utils in our release pipeline, which needs a more recent cosign support due to some verification errors. The related PR in release-service-catalog is here: https://github.com/konflux-ci/release-service-catalog/pull/510


ligangty:
@johnbieren I just found that the cosign git repo changed its assets url for the rpms, so fixed in new commit.

mmalina (approved):

mmalina:
I can see that it was built successfully. Merging.

PR #218, opened 2024-08-15
feat(RELEASE-1108): new dev portal wrapper

scoheb:
- removed python code from Tekton task and placed as separate script with tests

### DIFF
--- a/developer-portal-wrapper/developer_portal_wrapper.py
+++ b/developer-portal-wrapper/developer_portal_wrapper.py
@@ -1,0 +1,213 @@
+#!/usr/bin/env python3
+"""
+Run push-cgw-metadata command to push metadata to CGW:
+1. Extract all components under contentGateway key from dataPath
+2. Find all the files in contentDir that starts with the component name
+4. Generate necessary metadata for each file
+5. Dump the metadata to a YAML file
+6. Run push-cgw-metadata to push the metadata
+"""
+import os
+import yaml
+import hashlib
+import subprocess
+import argparse
+import logging
+
+LOG = logging.getLogger("developer-portal-wrapper")
+DEFAULT_LOG_FMT = "%(asctime)s [%(levelname)-8s] %(message)s"
+DEFAULT_DATE_FMT = "%Y-%m-%d %H:%M:%S %z"
+CGW_ENV_VARS_STRICT = (
+    "CGW_USERNAME",
+    "CGW_TOKEN",
+)
+
+WORKSPACE_DIR = "/tmp"
+METADATA_FILE_PATH = f"{WORKSPACE_DIR}/cgw_metadata.yaml"
+RESULT_FILE_JSON_PATH = f"{WORKSPACE_DIR}/results.json"
+
+os.makedirs(WORKSPACE_DIR, exist_ok=True)
+
+
+def generate_download_url(content_dir, file_name):
+    """
+    Generate a download URL in this format:
+    /content/origin/files/sha256/{checksum[:2]}{checksum}/{file_name}
+    """
+    prefix = "/content/origin/files/sha256"
+    sha256_hash = hashlib.sha256()
+    with open(content_dir + "/" + file_name, "rb") as f:
+        for byte_block in iter(lambda: f.read(4096), b""):
+            sha256_hash.update(byte_block)
+    checksum = sha256_hash.hexdigest()
+    return f"{prefix}/{checksum[:2]}/{checksum}/{file_name}"
+
+
+def generate_metadata(
+    product_name, product_code, product_version_name, content_dir, content_list, file_prefix
+):
+    """
+    Generate metadata for each file in
+    content_list that starts with the component name
+    """
+    metadata = []
+
+    default_values_per_component = {
+        "type": "FILE",
+        "shortURL": f"/cgw/{product_code}",
+        "hidden": False,
+        "invisible": False,
+    }
+
+    for file in content_list:
+        matching_component = file.startswith(file_prefix)
+
+        if matching_component:
+            LOG.info("processing file: %s", file)
+            component_item = {
+                "productName": product_name,
+                "productCode": product_code,
+                "productVersionName": product_version_name,
+                "downloadURL": generate_download_url(content_dir, file),
+                "label": file,
+            }
+            default_values_per_component["shortURL"] += f"/{file}"
+            metadata.append(
+                {
+                    "type": "file",
+                    "action": "create",
+                    "metadata": {**default_values_per_component, **component_item},
+                }
+            )
+        else:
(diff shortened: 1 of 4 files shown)


johnbieren (dismissed):
I assume this was tested and works so I didn't go too in depth reviewing it

developer-portal-wrapper/developer_portal_wrapper.py
mmalina, line 1: We have a convention of adding a symlink without the .py extension, so it probably makes sense to add it here as well. That way you will be able to just call `developer-portal-wrapper` in the task.

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+            )
+        else:
+            LOG.warning(f"Skipping file: {file} as it does not start with any component name")
+
+    return metadata
+
+
+def validate_args(args):
+    assert all([os.getenv(item) for item in CGW_ENV_VARS_STRICT]), (
+        f"Provide all required CGW environment variables: " f"{', '.join(CGW_ENV_VARS_STRICT)}"
```
mmalina, line 90: Why split the string in two if it fits on one line?
```suggestion
        f"Provide all required CGW environment variables: {', '.join(CGW_ENV_VARS_STRICT)}"
```

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+                    "metadata": {**default_values_per_component, **component_item},
+                }
+            )
+        else:
+            LOG.warning(f"Skipping file: {file} as it does not start with any component name")
+
+    return metadata
+
+
+def validate_args(args):
```
mmalina, line 88: This function doesn't do anything with the args at all. Maybe it originally did? So with the current functionality, I'd suggest to drop the arg and the return and call it `validate_env_vars`.

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+
+    metadata = generate_metadata(
+        product_name,
+        product_code,
+        product_version_name,
+        content_dir,
+        content_list,
+        file_prefix,
+    )
+    LOG.info("%s Files will be published to CGW", len(metadata))
```
mmalina, line 193: Files should be files (lowercase)?

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+            subprocess.run(command, check=True)
+        except subprocess.CalledProcessError:
+            LOG.exception("Command %s failed, check exception for details", cmd_str)
+            raise
+        except Exception as exc:
+            LOG.exception("Unknown error occurred")
+            raise RuntimeError from exc
+
+
+def entrypoint():
```
mmalina, line 229: This seems redundant. Why not just call `main()` in the `if` below? That's a pretty common practice.

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+
+    cmd_args = settings_to_args(args)
+    command = [main_command] + common_args + cmd_args
+    cmd_str = " ".join(command)
+
+    if args.dry_run:
+        LOG.info("Would have run: %s", cmd_str)
+    else:
+        try:
+            LOG.info("Running %s", cmd_str)
```
mmalina, line 219: This will leak the credentials to standard output. Would be great if it supported env vars, but I don't see anything here: https://release-engineering.github.io/pubtools-content-gateway/entrypoints_reference.html#cgw-environment

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: added

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: This was copy-paste...I fixed it per your suggestion.

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: fixed

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: fixed

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: done

scoheb:
> I assume this was tested and works so I didn't go too in depth reviewing it

Fully tested locally this morning :)

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -0,0 +1,211 @@
...
+    ]
+
+    command = [main_command] + common_args
+    cmd_str = " ".join(command)
+
+    if args.dry_run:
+        LOG.info("Would have run: %s", cmd_str)
+    else:
+        try:
+            LOG.info("Running %s", cmd_str)
```
mmalina, line 200:
```suggestion
    common_args = [
        "--CGW_filepath",
        METADATA_FILE_PATH,
        "--CGW_hostname",
        cgw_hostname,
    ]
    cred_args = [
        "--CGW_username",
        cgw_username,
        "--CGW_password",
        cgw_token,
    ]
    command = [main_command] + common_args
    cmd_str = " ".join(command)
    command += cred_args

    if args.dry_run:
        LOG.info("Would have run: %s", cmd_str)
    else:
        try:
            LOG.info("Running %s", cmd_str)
```

developer-portal-wrapper/developer_portal_wrapper.py
mmalina: Well you can still drop it from the log :) I understand that if this was already used directly in the task definition and it ran somewhere, it would have already leaked it. But I think it's still worth fixing it. I suggested a code change.

johnbieren: Should we just remove the log statement if it will leak creds?

johnbieren (dismissed):
commented on one open thread but otherwise looks fine to me

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: using @mmalina 's suggestion

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: fixed

johnbieren (approved):

mmalina (approved):

PR #226, opened 2024-08-19
fix(RELEASE-1108): logging should go to stdout

scoheb:
By default, logging goes to stderr in Python. This will change it to go to stdout. That makes it in line with our other Python scripts and also, it will help with debugging when this script runs in the Tekton task, because we suppress stderr there (and only print it if the script fails).

### DIFF
--- a/developer-portal-wrapper/developer_portal_wrapper.py
+++ b/developer-portal-wrapper/developer_portal_wrapper.py
@@ -8,6 +8,7 @@ Run push-cgw-metadata command to push metadata to CGW:
 6. Run push-cgw-metadata to push the metadata
 """
 import os
+import sys
 import yaml
 import hashlib
 import subprocess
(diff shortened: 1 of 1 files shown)


developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -142,13 +143,19 @@ def parse_args():
...
     args = parse_args()
-    validate_env_vars()
+
+    loglevel = logging.DEBUG if args.debug else logging.INFO
+
+    stream_handler = logging.StreamHandler(sys.stdout)
+    stream_handler.setLevel(loglevel)
 
     logging.basicConfig(
         level=logging.DEBUG if args.debug else logging.INFO,
```
mmalina, line 153: You missed this one from my PR ;)
```suggestion
        level=loglevel,
```

scoheb: fixed. thanks

mmalina (approved):

PR #232, opened 2024-08-27
fix(RELEASE-1125): correct shortUrl for dev portal

scoheb:
- correct value used for shortUrl in dev portal

### DIFF
--- a/developer-portal-wrapper/developer_portal_wrapper.py
+++ b/developer-portal-wrapper/developer_portal_wrapper.py
@@ -53,13 +53,7 @@ def generate_metadata(
     """
     metadata = []
 
-    default_values_per_component = {
-        "type": "FILE",
-        "shortURL": f"/cgw/{product_code}",
-        "hidden": False,
-        "invisible": False,
-    }
-
+    short_url_prefix = f"/cgw/{product_code}/{product_version_name}"
     for file in content_list:
         matching_component = file.startswith(file_prefix)
 
@@ -70,14 +64,17 @@ def generate_metadata(
                 "productCode": product_code,
                 "productVersionName": product_version_name,
                 "downloadURL": generate_download_url(content_dir, file),
+                "shortURL": f"{short_url_prefix}/{file}",
                 "label": file,
+                "type": "FILE",
+                "hidden": False,
+                "invisible": False,
             }
-            default_values_per_component["shortURL"] += f"/{file}"
             metadata.append(
                 {
                     "type": "file",
                     "action": "create",
-                    "metadata": {**default_values_per_component, **component_item},
+                    "metadata": component_item,
                 }
             )
         else:


happybhati (approved):

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -70,9 +70,9 @@ def generate_metadata(
                 "productCode": product_code,
                 "productVersionName": product_version_name,
                 "downloadURL": generate_download_url(content_dir, file),
+                "shortURL": default_values_per_component["shortURL"] + f"/{file}",
```
mmalina, line 73: You will end up with `shortURL` included in both `default_values_per_component` and `component_item` and then you combine them together (on line 80), so the latter one will one. So it will work, but it seems funky.

How about just setting something like `shortURLPrefix` outside of the dictionary and then using it here? Also, you could use 1 f-string:
```
"shortURL": f"{shortURLPrefix}/{file}",
```

scoheb: you know me when it comes to python...I am "funky". I'll updaste it

mmalina: (Also, it seems like many of the other items here (`productName`, `productCode`, `productVersionName`) could be moved to `default_values_per_component` as well. Or we could just ditch `default_values_per_component` and defined everything here. But I'm aware you're probably not the author of this so you don't have to do anything with that.)

scoheb: updated. please take a look

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -70,14 +64,17 @@ def generate_metadata(
...
+                "hidden": False,
+                "invisible": False,
             }
-            default_values_per_component["shortURL"] += f"/{file}"
             metadata.append(
                 {
                     "type": "file",
                     "action": "create",
-                    "metadata": {**default_values_per_component, **component_item},
+                    "metadata": {**component_item},
```
mmalina, line 77: I could have said that... I never used anything like this personally, but AFAICS, the ** was used to unpack the dictionaries to be able to combine the two together. But now that it's just one, we can simply do:
```suggestion
                    "metadata": component_item,
```

developer-portal-wrapper/developer_portal_wrapper.py
mmalina: Thanks. Added another small comment.

developer-portal-wrapper/developer_portal_wrapper.py
scoheb: updated. thanks

mmalina (approved):

PR #241, opened 2024-09-06
fix(KONFLUX-4292): upload a fixed gpg key id to pyxis

ralphbean:
To fix KONFLUX-4175, we need to upload a gpg key id to pyxis.

The right way to do this is for the SBOM itself to contain the gpg key id. Our task here would simply copy it to pyxis.  However, to do that fully, we'll need to get a change into both syft and cachi2, which may take some time.

As a stopgap, this change inspects the "publisher" field of the sbom. If it is set to Red Hat, we'll publish the Red Hat release2 key id to pyxis.

You might ask, what if the rpms are not actually signed by Red Hat keys? We have a separate layer of protection against that in KONFLUX-3149, but the data about which rpms are signed by which keys will not be exposed by that work in the way that the release process can use here.

* https://issues.redhat.com/browse/KONFLUX-4292
* https://issues.redhat.com/browse/KONFLUX-4175

This change should be reverted as a part of completing KONFLUX-4175.

### DIFF
--- a/pyxis/test_upload_rpm_manifest.py
+++ b/pyxis/test_upload_rpm_manifest.py
@@ -28,6 +28,14 @@ COMPONENTS = [
     {  # no upstream
         "purl": "pkg:rpm/rhel/pkg4@1-2.el8?arch=x86_64&distro=rhel-8.0",
     },
+    {  # with RH publisher
+        "purl": "pkg:rpm/rhel/pkg5?arch=noarch&upstream=pkg5-1-2.el8.src.rpm&distro=rhel-8.0",
+        "publisher": "Red Hat, inc.",
+    },
+    {  # with other publisher
+        "purl": "pkg:rpm/rhel/pkg6?arch=noarch&upstream=pkg6-1-2.el8.src.rpm&distro=rhel-8.0",
+        "publisher": "Blue Shoe, inc.",
+    },
     {  # not an rpm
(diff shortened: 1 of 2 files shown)


johnbieren (approved):
Would you like me to merge this @ralphbean or hold off for now?

mmalina (approved):

ralphbean:
Yes, please. Merge when you're ready. I'll rebase now.

lslebodn:
@johnbieren @ralphbean more than half  year passed and hotfix is still used.
https://github.com/konflux-ci/release-service-utils/blob/19e9f9950316ab4c997a7b46e949b885bc517607/pyxis/upload_rpm_data.py#L270-L276

Is there a plan to get rid of woarkaroudn?

And one more question. Is there a chance it might hide situation where unsigned rpms are shipped?

mmalina:
> @johnbieren @ralphbean more than half year passed and hotfix is still used.
> 
> https://github.com/konflux-ci/release-service-utils/blob/19e9f9950316ab4c997a7b46e949b885bc517607/pyxis/upload_rpm_data.py#L270-L276
> 
> Is there a plan to get rid of woarkaroudn?

Why not raise this in the referenced Jira rather than in a closed/merged PR? https://issues.redhat.com/browse/KONFLUX-4175
 
> And one more question. Is there a chance it might hide situation where unsigned rpms are shipped?

Checking if something is signed is not the goal of this task. RPM signatures are checked by Conforma.

PR #244, opened 2024-09-09
feat(CLDX-134): add mac-signing-wrapper script

pkhander:
This commit adds a wrapper script to sign mac binaries. The script will be called inside a tekton task (step) to sign mac binaries. Does the following:

1. Establish SSH connection to a mac vm.
2. Pull unsigned binaries using oras from a given unsigned digest
3. Sign and notarize binaries inside.
4. Push signed content to oras and return the signed digest. 

JIRA: CLDX-134

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,7 +43,8 @@ RUN pip3 install jinja2 \
     packageurl-python \
     pubtools-content-gateway \
     pubtools-pulp \
-    pubtools-exodus
+    pubtools-exodus \
+    paramiko
 
 # remove gcc, required only for compiling gssapi indirect dependency of pubtools-pulp via pushsource
 RUN dnf -y remove gcc
@@ -56,6 +57,7 @@ COPY utils /home/utils
 COPY templates /home/templates
 COPY pubtools-pulp-wrapper /home/pubtools-pulp-wrapper
 COPY developer-portal-wrapper /home/developer-portal-wrapper
+COPY mac-signing-wrapper /home/mac-signing-wrapper
 
 # It is mandatory to set these labels
 LABEL name="Konflux Release Service Utils"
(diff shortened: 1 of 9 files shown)


pkhander:
Hey team, May I please ask for a initial review on this ? I am going to start working on the unit tests in the meantime, thanks!

Dockerfile
mmalina, line 1: I wonder if having these wrappers has any benefit compared to calling the commands directly. Especially since this is already being done that way for Windows: https://github.com/konflux-ci/release-service-catalog/pull/541
It seems pretty straightforward. And unless there are reasons for this, having another layer of abstraction just makes it harder to debug or update in the future.

pkhander: @mmalina I thought it would be somewhat cleaner to call the script inside the step instead of putting the logic there, we will also have checksum signing coming soon (CLDX-82). It can make use of the modules as well. 
Personally, I found it easier to test the functionality and unit tests as a separate script and then insert it in tekton.

I wrote the logic for [publish-to-cgw](https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/publish-to-cgw/publish-to-cgw.yaml#L57) task which was later moved to utils as a [wrapper](https://github.com/konflux-ci/release-service-utils/tree/main/developer-portal-wrapper) to be used for Internal Requests, so I thought to start here this time :) 

If you still prefer this to be a part of the task instead of a separate wrapper, I can get working on that, np!

mmalina: OK, if you think it makes it easier, it's fine with me. I wonder what @johnbieren thinks as well.

mac-signing-wrapper/mac-signing-wrapper.py
```diff
@@ -0,0 +1,89 @@
...
+
+def main():
+    args = parse_arguments()
+    config = load_config(args.config_file)
+
+    try:
+        with SSHConnection(
+            config["mac_host"], config["mac_user"], config["mac_password"]
+        ) as ssh:
+            # with SSHConnection(
```
johnbieren, line 41: Did you mean to have this left over?

mac-signing-wrapper/macos_commands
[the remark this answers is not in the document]
pkhander: Removed unnecessary links and I am working on test files at the moment, will move this to ready once I submit those.

Dockerfile
pkhander: turns our we will not make much use of this. Limitations in paramiko library affected setting env variables and we thought it would be easier to just implement in bash. Closing, ty!

PR #248, opened 2024-09-12
feat(RELEASE-1144): push content sets to Pyxis

mmalina:
The `upload_rpm_manifest` script was renamed to
`upload_rpm_data`.

On top of the rpm manifest, it will now also update the ContainerImage.content_sets field in Pyxis.

The content sets is a list of unique strings
taken from the repository_id qualifier of purl
entries.

I tested the script by running it against several images in stage Pyxis. I hit different scenarios (no content sets in pyxis, then saved and skipped, and also no content sets in the sbom) and it all seemed to work.


pyxis/upload_rpm_data.py
```diff
@@ -55,29 +60,44 @@ def upload_container_rpm_manifest_with_retry(
     raise last_err
 
 
-def upload_container_rpm_manifest(graphql_api: str, image_id: str, sbom_path: str):
+def upload_container_rpm_data(graphql_api: str, image_id: str, sbom_path: str):
     """Upload a Container Image RPM Manifest to Pyxis"""
```
johnbieren, line 64: This should probably be updated

johnbieren (commented):
Unfortunate that github can't render the test file diff, but I diff'ed it locally and it seems sane. One nit pick

mmalina:
> Unfortunate that github can't render the test file diff, but I diff'ed it locally and it seems sane. One nit pick

Yeah, it's weird. It did show the diff properly for the main py file, but not for the test file. Maybe too many changes?

pyxis/upload_rpm_data.py
mmalina: Good catch, fixed.

johnbieren (approved):

PR #267, opened 2024-09-27
feat(RELEASE-1192): support using pipelineruns

scoheb:
- in the case of private internal clusters, using internal-services adds an additional layer of complexity that is not needed.
- this PR add 2 scripts that work to create PipelineRuns and wait for them.
- add new soft link for utils/wait-for-internal-request for consistency


utils/wait-for-pipelinerun
```diff
@@ -0,0 +1,143 @@
+#!/usr/bin/env bash
+
+# This script waits for one or more PipelineRuns to complete.
+#
+# The PLRs to watch can be specified either by name or by a label selector.
+#
+# The script waits for the PipelineRuns to reach a 'completed'
+# status and will provide an exit code as follows:
+#   - All succeeded: exit code 0
+#   - At least one IR failed or rejected: exit code 21
```
mmalina, line 10: IR -> PLR, also, "rejected" is not really a good word for PLR status?

mmalina:
Can you reference the jira in the PR/commit title? It should be [RELEASE-1164](https://issues.redhat.com//browse/RELEASE-1164) , right? Also, can you mention that this will be used in https://github.com/konflux-ci/release-service-catalog/pull/585 ?

utils/wait-for-internal-request
johnbieren, line 1: Why do we want this? The commit message for consistency, but for consistency with what? Only our python scripts have sym links so far I think (which not even sure that was necessary, but that's a separate issue)

utils/pipelinerun
```diff
@@ -0,0 +1,289 @@
...
+# Failed: error code 21
+# Rejected: error code 22
+#
+# Usage:
+#   ./pipelinerun -r request [-p <key=value> ...] [-s sync] [-t timeout] [--service-account name]
+#   [--task-git-url url] [--task-git-revision revision]
+#   [--pipeline-timeout 1h0m0s] [--task-timeout 0h55m0s] [--finally-timeout 0h5m0s]
+#
+# Parameters:
+#   -r                  Request: the name of the request.
```
johnbieren, line 20: What is a request for a pipelineRun?

utils/pipelinerun
```diff
@@ -0,0 +1,289 @@
...
+fi
+
+# Make sure all timeouts are passing in XhYmZs format
+PIPELINE_VALUES=$(echo $PIPELINE_TIMEOUT | awk -F[h,m,s] '{print $3}')
+TASK_VALUES=$(echo $TASK_TIMEOUT | awk -F[h,m,s] '{print $3}')
+FINALLY_VALUES=$(echo $FINALLY_TIMEOUT | awk -F[h,m,s] '{print $3}')
+if [[ $PIPELINE_VALUES = "" ]] || [[ $TASK_VALUES = "" ]] || [[ $FINALLY_VALUES = "" ]] ; then
+    echo The pipeline, task, or finally timeout value was passed incorrectly.
+    echo The correct format is XhYmZs, where X, Y, and Z are integers.
+    echo "You cannot leave out any of the 3 (use value 0 if necessary)."
```
johnbieren, line 163: They all have defaults. Why can't we leave any of them out?

utils/pipelinerun
```diff
@@ -0,0 +1,289 @@
...
+if [[ $ALL_TASKS_TIMEOUT -gt $PIPELINE_TIMEOUT_SECS ]] ; then
+    echo The sum of the task and finally timeout cannot exceed the pipeline timeout.
+    echo Pipeline timeout is $PIPELINE_TIMEOUT_SECS and the sum of the others is $ALL_TASKS_TIMEOUT
+    echo This leads to tekton validation webhook errors. Exiting...
+    exit 1
+fi
+
+# Ensure pipeline timeout does not exceed PipelineRun timeout
+if [ $PIPELINE_TIMEOUT_SECS -gt $TIMEOUT ] ; then
+    echo WARNING: The passed pipeline timeout is greater than the PipelineRun timeout
```
johnbieren, line 181: These messages are confusing to me. It is a pipelineRun timeout vs a script timeout. It talks about them both as pipeline timeouts

utils/pipelinerun
```diff
@@ -0,0 +1,289 @@
...
+PARAM_JSON="${PARAM_JSON_ARRAY[@]}"
+PARAM_JSON="${PARAM_JSON%,}" # remove final ,
+LABEL_JSON=$(echo "${LABEL_JSON_ARRAY[@]}" | jq -s 'add')
+
+# Create JSON payload for the PipelineRun
+PAYLOAD=$(jq -n \
+    --arg request "$REQUEST" \
+    --arg taskgiturl "$TASKGITURL" \
+    --arg taskgitrevision "$TASKGITREVISION" \
+    --arg pipeline "internal/pipelines/${REQUEST}/${REQUEST}.yaml" \
```
johnbieren, line 217: I see now why you want a request parameter. If we are doing it like this, I think this script should be called like internal-request-pipelinerun or something... anything to show that this can be used only for the internal pipeliens

utils/wait-for-internal-request
mmalina: I think what Scott meant is: pipelinerun -> wait-for-pipelinerun. So for consistency with that, we also want: internal-request -> wait-for-internal-request . I'm fine with having it.

utils/pipelinerun
mmalina: It's the name of the pipeline. Which makes about the same amount of sense as in the internal-request case...

johnbieren: Yes, https://github.com/konflux-ci/release-service-utils/pull/267#discussion_r1781518866

utils/wait-for-internal-request
johnbieren: wait-for-pipelinerun isn't a sym link

scoheb: it is. See this page - https://github.com/konflux-ci/release-service-utils/pull/267/files - on left hand side in file tree...

scoheb: maybe GH shows the file is gets linked to here?

utils/pipelinerun
scoheb: we could rename it to be `Job` if you prefer. whatever vehicle we use to trigger the signing...

legacy: internal-requests
this PR: pipelineruns

utils/pipelinerun
scoheb: I copied this from the other script.

Maybe it is because we try to validate the format if a value is passed:

```
$ export PIPELINE_TIMEOUT=1h0m0s
$ echo $PIPELINE_TIMEOUT | awk -F[h,m,s] '{print $3}'
0
$ export PIPELINE_TIMEOUT=1h
$ echo $PIPELINE_TIMEOUT | awk -F[h,m,s] '{print $3}'
 
$ 
```

johnbieren: Sure, you can improperly pass them and face error, but you can leave any of them out due to the defaults that are set

utils/pipelinerun
johnbieren: Isn't this a step towards not having internal requests? I don't think we should keep the internal request naming in it. This relates to my other comment https://github.com/konflux-ci/release-service-utils/pull/267/files#r1781518866. This file should be clearly about internal request pipelines and named as such, or it should be generic

utils/wait-for-internal-request
johnbieren: I see a blue folder with an arrow for wait-for-internal-request and a white paper for wait-for-pipelinerun

utils/pipelinerun
scoheb: and the other legacy script should be called `internal-request-internal-request`?

I vote to keep as-is

johnbieren: huh? what other script?
Okay, so what are we going to call the script that creates any pipelinerun, not only ones in internal/pipelines?

utils/pipelinerun
scoheb: and if they are left out, they get the defaults and then this condition is false.

utils/pipelinerun
scoheb: I will rework

utils/pipelinerun
scoheb: `internal-request`

Unfortunately, I think we are always going to have to support both types - **IRs** and **PLRs**

for public clusters, they'll always need to go thru internal-services to sign.

utils/wait-for-internal-request
mmalina: yeah, I see the same as Johnny - the only symlink is wait-for-internal-request

utils/pipelinerun
johnbieren: If I pass an incorrect value, I am going to get an error message that says "you cannot leave out any of the 3." That is not true, there are defaults, you can leave out any of the 3 you want (and if I do, while the ones I do pass are properly formatted, I won't see the error message, agreed). You just can't pass an improper value. Why can't we have the error message updated to be what is actually true?

utils/pipelinerun
johnbieren: why would we rename internal-request to internal-request-internal-request? I am not following

mmalina: I hope we don't need to create any other pipelinerun. I see this as a quick drop in replacement for internal requests, but if we decide we can get rid of internal-requests completely, we should refactor stuff so that everything still runs in the same pipeline. So I'm with Scott here.

utils/pipelinerun
scoheb: so you would prefer this?

IR case
- RequestType = internal-request
- create,wait scripts = internal-request (wait-for-internal-request)

PLR case
- RequestType = internal-request-pipelinerun
- create,wait scripts = internal-request-pipelinerun (wait-for-internal-request-pipelinerun)

utils/pipelinerun
johnbieren: This script only creates pipelines in the internal dir. I am really not following why we can't name it as such. Again, what will we name the script (if we find a need to create one) that creates *any* pipelinerun?

scoheb: > why would we rename internal-request to internal-request-internal-request? I am not following

see this: https://github.com/konflux-ci/release-service-utils/pull/267#discussion_r1782872315

utils/wait-for-internal-request
scoheb: I am trying to make sure we are consistent here and still backwards compatible.

you want me to rename `wait-for-ir` to `wait-for-internal-request` and have no soft link?

utils/wait-for-pipelinerun
scoheb: fixed

utils/pipelinerun
scoheb: updated

utils/pipelinerun
scoheb: reworked log messages

utils/pipelinerun
johnbieren: internal-request and wait-for-ir can stay as they are. I don't see reason to change it. I don't like calling this pipelinerun because it isn't just a pipelinerun. It is a very specific kind. We cannot call any pipelinerun that isn't in the internal directory with it. So, it should IMO be internal-pipelinerun, and the wait script can be wait-for-internal-pipelinerun or wait-for-internal-plr or whatever you'd like. Or, the script can be generic, work for any pipelinerun, keep its current name, and when we call it the parameter would be `internal/pipelines/$foo.yaml` instead of the parameter just being foo

mmalina: `internal-pipelinerun` sounds good to me (`internal-request-pipelinerun` would be bad IMO, because no IR is involved, so it would be confusing).

scoheb: using `internal-pipelinerun` now

utils/pipelinerun
scoheb: see previous comment. I suggest we close this thread

utils/internal-pipelinerun
```diff
@@ -0,0 +1,293 @@
...
+if [[ $ALL_TASKS_TIMEOUT -gt $PIPELINE_TIMEOUT_SECS ]] ; then
+    echo The sum of the task and finally timeout cannot exceed the pipeline timeout.
+    echo Pipeline timeout is $PIPELINE_TIMEOUT_SECS and the sum of the others is $ALL_TASKS_TIMEOUT
+    echo This leads to tekton validation webhook errors. Exiting...
+    exit 1
+fi
+
+# Warn if PipelineRun timeout does not exceed Request timeout
+if [ $PIPELINE_TIMEOUT_SECS -gt $TIMEOUT ] ; then
+    echo "WARNING: The passed Request timeout ($TIMEOUT) is greater than PipelineRun timeout ($PIPELINE_TIMEOUT_SECS)"
```
johnbieren, line 183: Can we change `Request` to `script` on this line and the one below it?
And the error message has it backwards. The plr timeout is greater, which means the script can timeout and fail before the plr times out

utils/wait-for-internal-request
```diff
@@ -138,5 +139,4 @@ while true; do
         exit 124
     fi
 
-    sleep 5
```
johnbieren, line 141: If we remove this sleep, aren't we going to be printing significantly more output since we have echoes in the loop?

utils/wait-for-internal-pipelinerun
```diff
@@ -0,0 +1,144 @@
...
+# The PLRs to watch can be specified either by name or by a label selector.
+#
+# The script waits for the PipelineRuns to reach a 'completed'
+# status and will provide an exit code as follows:
+#   - All succeeded: exit code 0
+#   - At least one PLR failed: exit code 21
+#   - Timeout reached: exit code 124
+#
+# Usage:
+#   ./wait-for-pipelinerun [-n name] [-l labels] [-t timeout]
```
johnbieren, line 14: needs update per new name. However, I am fine with this script being called `wait-for-pipelinerun` *because* there is nothing in it specific to the internal pipelineruns. Unless I am missing something, this script can be used to watch any tekton pipelineRun, not just internal ones. It is more generic (which is good IMO)

johnbieren (commented):
It sounds like you have been testing this, so I assume it is already done, but can you make sure the new scripts are +x permission? I believe they will need to be, and the github ui doesn't tell me the chmod status of the files

utils/internal-request
```diff
@@ -168,8 +167,8 @@ fi
 
 # Ensure pipeline timeout does not exceed InternalRequest timeout
 if [ $PIPELINE_TIMEOUT_SECS -gt $TIMEOUT ] ; then
-    echo WARNING: The passed pipeline timeout is greater than the InternalRequest timeout
-    echo This means the InternalRequest can fail before the pipelineRun times out, should it take that long
+    echo WARNING: The passed InternalRequest timeout is greater than the script timeout
```
johnbieren, line 170: I think this one was fine because the argument passed is a pipeline timeout, not internal request timeout (see line 35)

johnbieren (dismissed):

scoheb:
> It sounds like you have been testing this, so I assume it is already done, but can you make sure the new scripts are +x permission? I believe they will need to be, and the github ui doesn't tell me the chmod status of the files

confirmed

utils/internal-request
scoheb: updated

johnbieren: think you forgot to push

johnbieren (dismissed):

utils/wait-for-internal-request
mmalina: OK, but that makes it a breaking change. Wouldn't it be safer/easier to add a wait-for-ir symlink and maybe remove it later? Unless we have a simple PR ready that just renames all the mentions in the catalog repo. I don't like the idea of having this just buried in that big PR and having to merge that PR to avoid a risk of breakage.

johnbieren: If a task has its utils ref updated, it should use the new script name. I don't see this as very risky. It won't pass e2e if it is done improperly in the catalog pr

mmalina: You're right it would be caught by e2e. OK then.

utils/wait-for-internal-pipelinerun
```diff
@@ -0,0 +1,145 @@
...
+          print_conditions "$PLRS"
+          echo Result: failure
+          exit 21
+      fi
+  fi
+
+  if [ "$(date +%s)" -gt "$END_TIME" ]; then
+      echo "ERROR: Timeout while waiting for the PipelineRuns to complete"
+      print_conditions "$PLRS"
+      echo result: timeout
```
mmalina, line 140: Result - capital R for consistency with the other two.

utils/internal-pipelinerun
```diff
@@ -0,0 +1,293 @@
...
+#
+# Note: Internal in this sense means it is used by the Release Service in a specific fashion
+#       to interact with internal services such as RADAS
+#
+# In sync mode (default) the script waits for the PipelineRun to reach a 'completed'
+# status and will provide an exit code based on the PipelineRun status:
+#
+# Succeeded: error code 0
+# Failed: error code 21
+# Rejected: error code 22
```
mmalina, line 15: There is no such thing as rejected here. But there is timeout - code 124.

scoheb: fixed

utils/wait-for-internal-pipelinerun
scoheb: fixed

johnbieren (dismissed):

mmalina (dismissed):

pyxis/find_signature.py
```diff
@@ -0,0 +1,137 @@
...
+            "repository": repository,
+            "manifest_digest": manifest_digest,
+            "index": currentIndex,
+        }
+        body = {"query": query, "variables": variables}
+
+        data = pyxis.graphql_query(graphql_api, body)
+        signatures = data["find_signature_data_by_index"]["data"]
+        LOGGER.info(f"Found {len(signatures)} signatures.")
+        if len(signatures) == 1:
```
mmalina, line 103: This seems like a huge waste of requests, right? You will have N signatures for each pair of repository and manifest_digest. Ideally you would only need to make N requests in total to determine which of these signatures already exist. But with the current approach you will make N*N/2 (because for each reference/tag, you will start searching from index 1). But it's definitely a good start.

What I think we should do is have the script only take the repository and manifest_digest as parameters and also an output file and save a json list of all the references that have signatures. And then in the rh-sign image task, we would just check the json before making each signing request. Also, I would switch the order of the loops slightly:
```
for each component
  for each manifest_digest
    for each of (rh_registry_repo, registry_access_repo)
      find_signatures repo manifest_digest --output signed_references.json
      for each tag
        if repo:tag in signed_references.json
          sign_image
```

scoheb: At this point, the current method in **rh-sign-image** using the individual call to `find-signature` seem to be fine performance-wise although I agree there is lots of room for improvement.

The biggest issue to still the case where all the images have never been signed...not the skipping use-case.

I'd prefer to leave this optimisation for another PR so we can get this PR into production.

mmalina: ok, I created https://issues.redhat.com/browse/RELEASE-1216

scoheb: Can we resolve this now?

mmalina: Yep.

scoheb: I actually went ahead and created `find_signatures` and integrated it into `rh-sign-image`...

pyxis/find_signatures.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+"""
+Python script to check if signature exists
```
mmalina, line 3: Needs updating as it will now produce a list of all signatures existing for a pair of (repository, manifest_digest).

pyxis/find_signatures.py
```diff
@@ -0,0 +1,156 @@
...
+Optional env vars:
+PYXIS_GRAPHQL_API
+"""
+import argparse
+import logging
+import os
+
+import pyxis
+
+LOGGER = logging.getLogger("find_signature")
```
mmalina, line 18: -> find_signatures

pyxis/find_signature.py
```diff
@@ -0,0 +1,158 @@
...
+        "--retry",
+        "-r",
+        action="store_true",
+        help="If set, retry the query in case it fails",
+    )
+
+    return parser.parse_args()
+
+
+def find_signature_using_reference(graphql_api, reference: str, manifest_digest: str) -> bool:
```
mmalina, line 60: It seems some more updates are needed now that it was change to look up all references. You no longer need the `--reference` cli argument, right? And you have `--repository`, so you don't need to derive repository from `reference` (which you will no longer have). And the name of this function should also be changed.

scoheb: removed this script since it is no longer used

pyxis/find_signatures.py
scoheb: fixed

pyxis/find_signatures.py
scoheb: fixed

pyxis/find_signature.py
mmalina: Oh, I thought I was commenting on find_signatures. Didn't realize there were both files.

pyxis/find_signatures.py
```diff
@@ -0,0 +1,156 @@
+#!/usr/bin/env python3
+"""
+Python script to find all references for a given repository and manifest_digest
```
mmalina, line 3: I'm missing the keyword "signature" in this description ;)

So something like this maybe?
```suggestion
Python script to find all signatures in Pyxis for a given repository and manifest_digest 
and save a list of references that are signed in a text file, one reference per line
```

scoheb: thanks. added your suggestion.

pyxis/find_signatures.py
```diff
@@ -0,0 +1,161 @@
...
+def parse_arguments() -> argparse.Namespace:  # pragma: no cover
+    """Parse CLI arguments
+
+    :return: Dictionary of parsed arguments
+    """
+
+    parser = argparse.ArgumentParser(
+        description="find all signatures in Pyxis for a given repository and manifest_digest"
+                    "and save a list of references that are signed in a text file, one reference"
+                    "per line"
```
mmalina, line 31: Black failed on the indentation here, in case you haven't noticed.
```suggestion
        "and save a list of references that are signed in a text file, one reference"
        "per line"
```
(I hope I removed the right amount of spaces :) )

pyxis/find_signatures.py
```diff
@@ -0,0 +1,161 @@
...
+
+    return parser.parse_args()
+
+
+def find_signatures_for_repository(graphql_api, repository: str, manifest_digest: str) -> set:
+    LOGGER.info(f"repository: {repository}")
+    LOGGER.info(f"manifest_digest: {manifest_digest}")
+
+    references = set()
+    query = """
```
mmalina, line 73: It turns out we can do everything with just one request, see my comment here: https://issues.redhat.com/browse/RELEASE-1216?focusedId=25777436&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25777436

That query has pagination, so you will set page_size to something like 50 which in most cases will give you all the signatures - but it will have to be in a loop and run again until returned size < page_size. I can help with the code, let me know.

pyxis/find_signatures.py
```diff
@@ -0,0 +1,161 @@
...
+            LOGGER.debug(f"signature_data: {signature_data}")
+            reference_from_signature = signature_data["get_signature"]["data"]["reference"]
+            LOGGER.debug(f"reference_from_signature: {reference_from_signature}")
+            references.add(reference_from_signature)
+            current_index += 1
+        else:
+            LOGGER.debug("no more signatures")
+            break
+    LOGGER.info(f"Found {len(references)} references.")
+    return references
```
mmalina, line 140:
```suggestion
def find_signatures_for_repository(graphql_api, repository: str, manifest_digest: str, page_size: int = 50) -> set:
    LOGGER.info(f"repository: {repository}")
    LOGGER.info(f"manifest_digest: {manifest_digest}")

    references = set()
    query = """
query ($repository: String!, $manifest_digest: String!, $page: Int!, $page_size: Int!) {
    find_signatures(
        page: $page
        page_size: $page_size
        sort_by: [{field: "last_update_date", order: DESC}]
        filter: {and: [{manifest_digest: {eq: $manifest_digest}}, {repository: {eq: $repository}}]}
    ) {
        error {
            detail
            status
        }
        data {
            _id
            reference
        }
    }
}
    """
    has_more = True
    page = 0
    while has_more:
        LOGGER.debug(f"current page: {page}")
        variables = {
            "repository": repository,
            "manifest_digest": manifest_digest,
            "page": page,
            "page_size": page_size
        }
        body = {"query": query, "variables": variables}

        data = pyxis.graphql_query(graphql_api, body)
        signatures = data["find_signatures"]["data"]
        LOGGER.debug(f"Found {len(signatures)} signatures.")
        references.update([signature["reference"] for signature in signatures])

        has_more = len(signatures) == page_size
        page += 1
    LOGGER.info(f"Found {len(references)} references.")
    return references
```

This works for me. Tried it like this:
```
./pyxis/find_signatures.py --pyxis-graphql-api "https://graphql.pyxis.engineering.redhat.com/graphql/" --repository rhel9/bootc-image-builder --manifest_digest "sha256:9f0056436a269360a95f69ab0070cde0672fff397bc782862e04c7732463b165" --output_file ~/tmp/pyxis/signature_references.txt --verbose
```
If I changed the page_size to 2, it ran the request multiple times and worked as well.

pyxis/find_signatures.py
mmalina: I posted the modified function in another comment.

scoheb: adjusted the code. thanks!

pyxis/find_signatures.py
scoheb: adjusted the code. thanks!

mmalina (approved):

PR #269, opened 2024-10-02
refactor(RELEASE-1009): remove upload_sbom.py pyxis script

johnbieren:
This commit removes pyxis/upload_sbom.py and its test file as it is no longer used in our pipelines. Some of its functions were used in upload_rpm_data.py, so those and their tests were migrated to the upload_rpm_data.py file.


pyxis/upload_rpm_data.py
```diff
@@ -93,6 +94,34 @@ def upload_container_rpm_data(graphql_api: str, image_id: str, sbom_path: str):
         update_container_content_sets(graphql_api, image_id, content_sets)
 
 
+def parse_arguments() -> argparse.Namespace:  # pragma: no cover
+    """Parse CLI arguments
+
+    :return: Dictionary of parsed arguments
+    """
+
+    parser = argparse.ArgumentParser(description="Upload sbom metadata to Pyxis via graphql")
```
mmalina, line 103: Should be updated: "Upload RPM data to Pyxis via graphql" maybe?

johnbieren: agreed, nice catch. fixed

mmalina (approved):

PR #270, opened 2024-10-03
feat: add rpms-signature-scan task to .tekton pipelines

johnbieren:


happybhati (approved):

mmalina:
Here's a PR to copy cosign from the RHTAS cosign image instead of installing an unsigned rpm from the upstream cosign project on github: https://github.com/konflux-ci/release-service-utils/pull/271

When merged, that should get rid of the EC violation.

davidmogar (approved):

PR #271, opened 2024-10-04
fix: use cosign from RHTAS image

mmalina:
Konflux build now contains a new rpms-signature-scan task. We have one unsigned rpm in our image - cosign.
As suggested by Lui, let's copy cosign from the RHTAS Cosign image instead of installing the insigned rpm from the cosign project on Github.

Related discussion: https://redhat-internal.slack.com/archives/C07MRFX3ULS/p1727982405750819


lcarva (approved):

davidmogar (approved):

johnbieren:
@mmalina builds are now failing because we aren't able to pull the cosign image without authentication. Any ideas?

mmalina:
> @mmalina builds are now failing because we aren't able to pull the cosign image without authentication. Any ideas?

Hmm, I'm not sure why it would stop working now... But we should be able to fix it by setting up a pull secret?

PR #280, opened 2024-10-16
fix: gitlint installation

mmalina:
Latest ubuntu does not allow system wide pip install. It can be overriden with `--break-system-packages`, but it's better to create a virtual env.

This first came up here: https://github.com/hacbs-release/app-interface-deployments/pull/198

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -17,16 +17,19 @@ jobs:
         uses: frenck/action-yamllint@v1
   gitlint:
     name: Run gitlint checks
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Check out code
         uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


mmalina:
> This makes me wonder why we don't use a reusable workflow for this

Is there such a thing? If so, we should explore it. If you mean actions, that's an option too. But it's always a bit iffy to use a random user's action (we already do it though). Today I checked and saw two - not sure which of those two I would pick. Maybe the more popular.

johnbieren:
> > This makes me wonder why we don't use a reusable workflow for this
> 
> Is there such a thing? If so, we should explore it. If you mean actions, that's an option too. But it's always a bit iffy to use a random user's action (we already do it though). Today I checked and saw two - not sure which of those two I would pick. Maybe the more popular.

Yeah you just add (I think it is called) workflow_dispatch as a trigger to the workflow then any other repo can use it

PR #283, opened 2024-10-22
Tweak product SBOM generator

mprpic:
These guidelines are defined at:

https://redhatproductsecurity.github.io/security-data-guidelines/sbom/

Specific changes:

- Datetimes should always use UTC

- `documentNamespace` doesn't need to contain any product-specific information and serve only as an identifier (this information can be set in the `name` field")

- Data license should be `CC-BY-4.0`

- The `creators` field should note the Tool that is used to generate the SBOM. Also the SPDX 2.3 JSON schema requires the field to be called `creators` instead of `creator`: https://github.com/spdx/spdx-spec/blob/v2.3/schemas/spdx-schema.json

### DIFF
--- a/sbom/create_product_sbom.py
+++ b/sbom/create_product_sbom.py
@@ -5,7 +5,7 @@ data.json file.
 """
 import json
 import uuid
-from datetime import datetime
+from datetime import datetime, timezone
 import argparse
 from collections import defaultdict
 from typing import DefaultDict, Dict, List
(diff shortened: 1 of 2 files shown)


mmalina (commented):
I ran the tests and you'll need to apply Black formatting and also fix the commit message - we use conventional commits, see here: https://github.com/konflux-ci/release-service-catalog/blob/development/CONTRIBUTING.md#commit-message-formatting-and-standards

Otherwise it looks good.

johnbieren:
/ok-to-test
It looks like only one of the commits fits conventional commits. Can you squash it down to one commit?

johnbieren:
Ralph added the gitignore change in another PR so you should be able to remove that commit entirely here

johnbieren (approved):

mprpic:
> There is a merge commit here, which fails the gitlint check. Can you make the PR just one commit (the fix: one)?
> Once it is fixed and passes CI, do you want it merged or wait on it

Ok, rebased from main, hopefully should be good to merge now :crossed_fingers:

johnbieren (approved):

johnbieren:
```
1: T1 Title exceeds max length (77>72): "fix: update product SBOM generator to be in line with ProdSec SBOM guidelines"
12: B1 Line exceeds max length ([7](https://github.com/konflux-ci/release-service-utils/actions/runs/11503534729/job/32023055984?pr=283#step:4:8)4>72): "  information and serve only as an identifier (this information can be set"
```

PR #284, opened 2024-10-22
feat(KONFLUX-4892): supply layer size information to pyxis

ralphbean:
Notably, this needs to be populated by `oras` in the task that invokes this script, but the test suite here verifies that the script will not break if this is rolled out before that task is modified.

https://issues.redhat.com/browse/KONFLUX-4892

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -179,7 +179,28 @@ def prepare_parsed_data(args) -> Dict[str, Any]:
         "digest": args.architecture_digest,
         "architecture": args.architecture,
         "layers": [layer["digest"] for layer in oras_manifest_fetch.get("layers", [])],
+        "uncompressed_layer_sizes": [
+            {"layer_id": layer["digest"], "size_bytes": layer["size"]}
+            for layer in oras_manifest_fetch.get("uncompressed_layers", [])
+        ],
+        "uncompressed_size_bytes": sum(
+            [
+                layer.get("size", 0)
+                for layer in oras_manifest_fetch.get("uncompressed_layers", [])
+            ]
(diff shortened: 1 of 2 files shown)


pyxis/create_container_image.py
```diff
@@ -178,8 +178,28 @@ def prepare_parsed_data(args) -> Dict[str, Any]:
...
+        "uncompressed_layer_sizes": [
+            {"layer_id": layer["digest"], "size_bytes": layer["size"]}
+            for layer in reversed(oras_manifest_fetch.get("uncompressed_layers", []))
+        ],
+        "uncompressed_size_bytes": sum(
+            [layer["size"] for layer in oras_manifest_fetch.get("uncompressed_layers", [])]
+        ),
+        "sum_layer_size_bytes": sum(
+            [layer.get("size", 0) for layer in oras_manifest_fetch.get("layers", [])]
+        ),
```
mmalina, line 193: Any reason to use a different approach for these two? The first will fail if there is no `size` for a `layer`, the other will return `0` in that case. So probably makes sense to make the first one safer as well?

mmalina (approved):
Just a small nit, but otherwise it looks good.

pyxis/create_container_image.py
ralphbean: No, I think that's a thinko or an accident of multiple revisions while working on this locally. I'll convert both to use the safer `.get("size", 0)`.

ralphbean: Should be resolved now.

mmalina: ok, thanks

mmalina (approved):

PR #254, opened 2024-09-17
feat(ISV-5129): add script to create product-level SBOMs

jedinym:
Add a python script to create product-level SBOMs in the `create-product-sbom` tekton task.

PR #290, opened 2024-10-31
feat(ISV-5128): add script to update component sboms

wcheang:


sbom/update_component_sbom
mmalina, line 1: Correct me if I'm wrong, but it looks like you created this as an actual file and not a symlink? At least that's what it looks like if I compare this to this PR: https://github.com/konflux-ci/release-service-utils/pull/254/files#diff-26bc81b0774ec9f78e7aa4ae5d90c3b8a4f13cac3a2c2a46b10d2217698830d1 - you can see a different icon for the file on the left.

sbom/update_component_sbom.py
```diff
@@ -0,0 +1,125 @@
...
+    Args:
+        data_path: path to data.json file containing image metadata.
+        input_path: path to directory holding SBOMs files to be updated.
+        output_path: path to directory to save updated SBOMs.
+    """
+    with open(data_path, "r") as data_file:
+        data = json.load(data_file)
+        release_notes = data["releaseNotes"]
+
+    component_to_purls_map = get_component_to_purls_map(release_notes["images"])
```
mmalina, line 79: Technically, only the `json.load` file needs the open file, right? And also, the rest of `release_notes` is never used, only images. So maybe you can simplify it like this?
```suggestion
    with open(data_path, "r") as data_file:
        data = json.load(data_file)

    component_to_purls_map = get_component_to_purls_map(data["release_notes"]["images"])
```

mmalina (commented):
Only two small comments, but otherwise it looks good.

sbom/update_component_sbom
wcheang: You are correct. Somehow my brain saw what's done for other scripts and didn't read them as symlinks, and thought it was some repo-specific mechanism :sweat_smile: Fixed and added the proper symlink.

mmalina: ok, thanks, no worries :)

mmalina (approved):
You will need to rebase and then we can merge.

wcheang:
Rebased and ran black on the test file.

mmalina (approved):

PR #292, opened 2024-11-01
fix(TEAMNADO-7734): don't mangle long strings when generating advisories

ralphbean:
This fixes a particular case with long urls which got split when over the 76 character limit. Those then got a space inserted in the middle of them because of the way markdown works when you split long strings.

The solution here is just to not wordwrap the advisory yaml output.

I figure, it is meant for machines anyways, not humans, so it doesn't matter if it has long lines.


ralphbean:
I confirmed that the test failed _first_ before I fixed the template (which then passed the test).

mmalina (approved):

PR #303, opened 2024-11-08
fix(KFLUXBUGS-1848): associate existing images with a new repository

ralphbean:
This adds a new behavior.

If asked to create an image associated with a repository, and the image does not exist at all - then create it, as before.

If the image does exist, but is not associated with the desired repository, then submit a PATCH request to update it, adding the desired repository.

This is related to https://github.com/konflux-ci/release-service-utils/pull/300 and KFLUXBUGS-1848.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -132,29 +132,28 @@ def setup_argparser() -> Any:  # pragma: no cover
     return parser
 
 
-def image_already_exists(args, digest: str, repository: str) -> bool:
+def proxymap(repository: str) -> str:
+    """Map a backend repo name to its proxy equivalent.
+
+    i.e., map quay.io/redhat-pending/foo----bar to foo/bar
+    """
+    return repository.split("/")[-1].replace("----", "/")
+
+
+def image_already_exists(args, digest: str, repository: str) -> Any:
     """Function to check if a containerImage with the given digest and repository
     already exists in the pyxis instance
 
-    :return: True if one exists, else false
+    If `repository` is None, then the return True if the image exists at all.
+
+    :return: the image id, if one exists, else None if not found
     """
 
-    # we need the repository name without the registry and organization
-    # given:
-    #   quay.io/redhat-pending/ubi9----buildah
-    # we need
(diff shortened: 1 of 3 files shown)


pyxis/create_container_image.py
```diff
@@ -358,6 +322,43 @@ def add_container_image_repository(args, parsed_data: Dict[str, Any], image: Dic
         raise Exception("Image metadata was not successfully added to Pyxis.")
 
 
+def construct_repositories(args, parsed_data):
```
ralphbean, line 325: A docstring for this would be nice.

Is it constructing the *expected* or *desired* repositories that we want to see in pyxis? Or is it constructing a list of repositories that are actually in pyxis? (I feel sure its the former after thinking about it for a sec, but that's the first question that popped into my mind when I read the function name).

pyxis/create_container_image.py
```diff
@@ -329,7 +368,25 @@ def main():  # pragma: no cover
 
     parsed_data = prepare_parsed_data(args)
 
-    if not image_already_exists(args, args.architecture_digest, args.name):
+    # First check if it exists at all
+    image = image_already_exists(args, args.architecture_digest, repository=None)
+    identifier = image["_id"]
```
mmalina, line 373: This should be moved below into the `if` - this will fail if image is None.

ralphbean: Fixed in [b2483ab](https://github.com/konflux-ci/release-service-utils/pull/303/commits/b2483abb00d15c98f39326def64209bf90f1edef).

pyxis/create_container_image.py
ralphbean: I think we decided that @mmalina is going to take up this refactor in a separate PR.

mmalina (approved):

pyxis/create_container_image.py
mmalina: Correct. Draft here: https://github.com/konflux-ci/release-service-utils/pull/305

johnbieren:
https://issues.redhat.com/browse/RELEASE-793 added for unit tests

PR #305, opened 2024-11-08
refactor: patching ContainerImage to add repositories

mmalina:
This is a followup of what was done in #303

Most notably:

- rename the image_already_exists to find_image to reflect what it does - it used to return bool, now it returns the image if found.
- only get the image from Pyxis once - no need to get it a second time with the repository in the filter, we can just search through the repositories in the image that we got from the first request
- only include the repositories field in the patch payload
- for consistency, create the same two repo entries when adding a new repository like we do when creating a brand new image

It seems this mostly covers https://issues.redhat.com/browse/RELEASE-1285 as well - I just need to add tests for pyxis.patch.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -149,22 +149,16 @@ def proxymap(repository: str) -> str:
     return repository.split("/")[-1].replace("----", "/")
 
 
-def image_already_exists(args, digest: str, repository: str) -> Any:
-    """Function to check if a containerImage with the given digest and repository
-    already exists in the pyxis instance
+def find_image(pyxis_url, digest: str) -> Any:
+    """Function to find a containerImage with the given digest.
 
-    If `repository` is None, then the return True if the image exists at all.
-
-    :return: the image id, if one exists, else None if not found
+    :return: the image, if one exists, else None if none found
     """
 
-    # quote is needed to urlparse the quotation marks
     raw_filter = f'repositories.manifest_schema2_digest=="{digest}";not(deleted==true)'
-    if repository:
-        raw_filter += f';repositories.repository=="{proxymap(repository)}"'
+    # quote is needed to urlparse the quotation marks
     filter_str = quote(raw_filter)
-
-    check_url = urljoin(args.pyxis_url, f"v1/images?page_size=1&filter={filter_str}")
+    check_url = urljoin(pyxis_url, f"v1/images?page_size=1&filter={filter_str}")
 
     # Get the list of the ContainerImages with given parameters
     rsp = pyxis.get(check_url)
@@ -178,9 +172,21 @@ def image_already_exists(args, digest: str, repository: str) -> Any:
     if "_id" in query_results[0]:
         emit_id(query_results[0]["_id"])
     else:
-        raise Exception("Image metadata was found in Pyxis, but the id key was missing.")
+        raise RuntimeError("Image metadata was found in Pyxis, but the id key was missing.")
 
     return query_results[0]
+
+
+def repo_in_image(repository_str: str, image: Dict[str, Any]) -> bool:
+    """Check if a repository already exists in the ContainerImage
+
+    :return: True if repository_str string is found in the ContainerImage repositories,
+        False otherwise
+    """
+    for repository in image["repositories"]:
+        if repository["repository"] == repository_str:
+            return True
+    return False
 
 
 def prepare_parsed_data(args) -> Dict[str, Any]:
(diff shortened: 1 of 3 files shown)


pyxis/create_container_image.py
```diff
@@ -378,12 +370,11 @@ def main():  # pragma: no cover
 
     # First check if it exists at all
     LOGGER.info(f"Checking to see if digest {args.architecture_digest} exists in pyxis")
-    image = image_already_exists(args, args.architecture_digest, repository=None)
-    if image:
-        # Then, check if it exists in association with the given repository
+    image = find_image(args.pyxis_url, args.architecture_digest)
+    if image is not None:
         identifier = image["_id"]
-        LOGGER.info(f"It does! Checking to see if it's associated with {args.name}")
```
johnbieren, line 385: Didn't Ralph just add this? Shouldn't we keep it?

pyxis/test_create_container_image.py
```diff
@@ -4,87 +4,87 @@
...
-def test_image_already_exists__image_does_exist(mock_get):
+def test_find_image__image_does_exist(mock_get):
     # Arrange
     mock_rsp = MagicMock()
     mock_get.return_value = mock_rsp
-    args = MagicMock()
-    args.pyxis_url = mock_pyxis_url
-    args.architecture_digest = "some_digest"
-    args.name = "server/org/some_name"
+    mock_pyxis_url
```
johnbieren, line 35: what is this? no equal sign or anything

pyxis/test_create_container_image.py
```diff
@@ -4,87 +4,87 @@
...
         + "v1/images?page_size=1&filter="
         + "repositories.manifest_schema2_digest%3D%3D%22some_digest%22"
         + "%3Bnot%28deleted%3D%3Dtrue%29"
-        + "%3Brepositories.repository%3D%3D%22some_name%22"
     )
 
 
 @patch("create_container_image.pyxis.get")
-def test_image_already_exists__image_does_not_exist(mock_get):
+def test_find_image__image_does_not_exist(mock_get):
```
johnbieren, line 56: A test for data existing but no `_id` in it would be nice

pyxis/test_create_container_image.py
```diff
@@ -467,3 +470,110 @@ def test_prepare_parsed_data__with_layer_sizes(mock_open):
...
+    result = repository_digest_values(args)
+
+    assert result == {
+        "manifest_schema2_digest": "mydigest",
+        "manifest_list_digest": "mytopdigest",
+    }
+
+
+@patch("create_container_image.datetime")
+def test_construct_repositories(mock_datetime):
```
johnbieren, line 529: one with rh-push false perhaps?

johnbieren (commented):
Couple small comments but otherwise lgtm

pyxis/create_container_image.py
mmalina: This is now covered by the messages below - the nested `if - else`. We will either log that it exists, including the repo, or we log that it exists, but the repo needs to be added.

pyxis/test_create_container_image.py
mmalina: Not sure, good catch. Removed. Also, I renamed `mock_pyxis_url` in the whole file to `PYXIS_URL` to make it consistent with how we would usually set up a "constant" at the top of the file.

pyxis/test_create_container_image.py
mmalina: TBH, I don't think it's realistic. Since this is a REST call, if the call is successful (code 200), then all the objects returned should be legitimate instances of the ContainerImage object. I understand that we have the check in the code for this. I guess I could remove it, but on the other hand it doesn't hurt to have it. So I added a test. Also, I changed the exception to `RuntimeError` as that's what we would typically use - generic `Exception` shouldn't be used if possible.

pyxis/test_create_container_image.py
mmalina: Added.

pyxis/test_pyxis.py
mmalina, line 1: @johnbieren FYI, also added tests for `pyxis.patch`.

johnbieren (approved):

mmalina:
I did some testing - I created an image. Then I ran the script again with the same arguments - it skipped the creation. Then I ran it again, but with a different repository - it added the repository. Then I ran it again with repo1 and repo2 and it skipped everything both times. The image I created is here: https://pyxis.stage.engineering.redhat.com/v1/images/id/6746e22dee93ab8a1d8d7d16

PR #307, opened 2024-11-11
feat(RELEASE-1214): add support for spdx sbom format

mmalina:
Build team is moving from cyclonedx to spdx sbom format. To support both formats until the transition is over:
* The original upload_rpm_data was copied as upload_rpm_data_cyclonedx to preserve the original behavior for now
* upload_rpm_data was modified to work with spdx format


mmalina:
> So for the review, upload_rpm_data_cyclonedx.py and test_upload_rpm_data_cyclonedx.py are just copies and do not need review? 

Correct.

> So then as part of this ticket, the tekton task will be switched to use `upload_rpm_data_cyclonedx` and once the transition is over the task will be modified to use `upload_rpm_data` and then the *_cyclonedx` files can be removed here? Do I have that right?

Yeah, for now we will run one of the scripts depending on the sbom type: https://github.com/konflux-ci/release-service-catalog/pull/685/files

And eventually we will remove the upload_rpm_data_cyclonedx.py script and the logic in the task and support just spdx.

pyxis/test_upload_rpm_data.py
```diff
@@ -18,38 +17,78 @@
...
-    {  # no purl
-        "bom_ref": "ref",
+        "externalRefs": [
+            {
+                "referenceType": "purl",
+                "referenceLocator": "pkg:golang/./staging/src@(devel)#k8s.io/api",
+            }
+        ]
     },
+    {},  # no purl
```
johnbieren, line 84: Can we also get one that has an `externalRefs` entry but the `referenceType` is not `purl`?

johnbieren (commented):
lgtm otherwise

pyxis/test_upload_rpm_data.py
mmalina: In [the examples](https://github.com/containerbuildsystem/cachi2/pull/608/files#diff-fd85b5dd3c4c6a63070c68cb8e96cb01a194799dace9e06867858787cc0649c0) I was given I actually don't see anything that is not a purl. But I see pypi purls and they should be ignored, so I added one.

johnbieren: The new test case is good, but whether it will happen or not, I think it would be nice to have
```
            if externalRef.get("referenceType") != "purl":
                continue
```
from the code, tested

johnbieren (approved):

johnbieren (approved):

PR #310, opened 2024-11-11
fix: supply manifest list digest to pyxis

ralphbean:
We're populating containerImage entities incorrectly.

I think this has no real effect, but if you look at one that we generate, the manifest_list_digest and manifest_schema2_digest are always the same, but they're not supposed to be. We currently incorrectly always set them to the value of the arch-specific image manifest, when the point is to provide both values, the multi-arch image index manifest as the manifest_list_digest and the arch-specific image manifest as the manifest_schema2_digest.

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -246,11 +246,11 @@ def pyxis_tags(args, date_now):
     ]
 
 
-def repository_digest_values(args, docker_image_digest):
+def repository_digest_values(args):
     """Return digest values for the repository entry in the image entity"""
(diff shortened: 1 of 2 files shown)


mmalina (approved):

mmalina:
This actually brings one potential issue - the manifest_list_digest will no longer be unique - meaning that if we were to search for existing images using `repositories.manifest_list_digest`, we would find all the arch images. But the search for existing images was already broken for multiarch images anyway, AFAICS, fixing here: https://github.com/konflux-ci/release-service-utils/pull/319

PR #311, opened 2024-11-11
fix: handle none-type values more carefully

ralphbean:
In the event that an image does not have any gzipped layers, then no uncompressed layer information will be handed to this tool and it will pass on the None value to pyxis. That fails pyxis' input validation, blocking containerImage creation.

In this change, simply drop these values if they happen to be none.


ralphbean:
Apologies for the heavy hand, but I'm going to merge this one without review. Working on a late-breaking blocker.

johnbieren:
Ticket for tests https://issues.redhat.com/browse/RELEASE-1289

PR #312, opened 2024-11-12
fix: ignore gpg-pubkey package

ralphbean:
This rpm is never signed. It delivers the public key used to validate all other packages.

It is present in the sbom, but the container catalog gets confused when it is present. As an unsigned rpm, it displays it as "excluded" from security scans.

Rather than claim that it is signed, let's exclude it from the manifest uploaded to pyxis.


ralphbean:
You can see that the rpm-verifier task doesn't report the gpg-pubkey, here: https://github.com/redhat-appstudio/tools/blob/main/verify_rpms/rpm_verifier.py#L87-L99

ralphbean:
OSBS ignores them here: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/plugins/rpmqa.py#L76-L80

ralphbean:
.. and also here: https://github.com/containerbuildsystem/atomic-reactor/blob/master/atomic_reactor/utils/rpm.py#L121-L122

mmalina (approved):
Thanks for linking other places where this is done. I think it makes sense to do the same here. Keep in mind that I am in the middle of implementing https://issues.redhat.com/browse/RELEASE-1214 - the utils image is merged, but I haven't updated catalog yet. I can wait for this update and use this image in the catalog update. 

But you're only changing it in the spdx script - for now build still produces cyclonedx and for that upload_rpm_data_cyclonedx will be run. So if you'd like to get this working even before the switch to spdx, you should add it to the other script as well. See this PR for the change where we added the second script: https://github.com/konflux-ci/release-service-utils/pull/307

ralphbean:
Ack - this is not urgent. Let's let your change proceed and then refactor mine on top of it if necessary. Mine is ad hoc and non-urgent. Yours is planned.

mmalina (approved):

mmalina:
PR to bump the image in catalog (Draft - waiting for the new image): https://github.com/konflux-ci/release-service-catalog/pull/693

PR #315, opened 2024-11-13
fix(ISV-5128): also update sbom metadata component purl

wcheang:
Previously the update-component-sbom script is only updating the component purl in the list of components. But in CycloneDX, there is also a component purl in the metadata.

### DIFF
--- a/sbom/update_component_sbom.py
+++ b/sbom/update_component_sbom.py
@@ -42,6 +42,12 @@ def update_cyclonedx_sbom(sbom: Dict, component_to_purls_map: Dict[str, List[str
         component_to_purls_map: dictionary mapping of component names to list of purls.
     """
     LOG.info("Updating CycloneDX sbom")
+
+    component_name = sbom["metadata"]["component"]["name"]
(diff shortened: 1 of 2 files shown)


sbom/update_component_sbom.py
```diff
@@ -42,6 +42,12 @@ def update_cyclonedx_sbom(sbom: Dict, component_to_purls_map: Dict[str, List[str
         component_to_purls_map: dictionary mapping of component names to list of purls.
     """
     LOG.info("Updating CycloneDX sbom")
+
+    componenet_name = sbom["metadata"]["component"]["name"]
+    if componenet_name in component_to_purls_map:
+        # only one purl is supported for CycloneDX
+        sbom["metadata"]["component"]["purl"] = component_to_purls_map[componenet_name][0]
```
masikor, line 49: Typo in `componenet_name` should be `component_name`.

johnbieren (commented):
lgtm once `s/componenet/component/g' change is made

mmalina (approved):

PR #319, opened 2024-11-18
fix(RELEASE-999): check for existing Pyxis image

mmalina:
The function that checks if the image already exists in Pyxis would always use `repositories.manifest_schema2_digest` in the query. But we only set that field for non-multiarch images. For multiarch, we set `repositories.manifest_list_digest`.

So that means that for multiarch images, I don't see how this could ever work.

The obvious fix would be to search for the right field for the particular image. The issue with that is that for multiarch images, we would have the same manifest list entry for each of the arch images, so we would only create the first one and all the rest would be skipped. So instead, let's use the image_id field which is set to the arch specific digest (for single arch, it's the main digest).

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -149,7 +149,7 @@ def proxymap(repository: str) -> str:
     return repository.split("/")[-1].replace("----", "/")
 
 
-def image_already_exists(args, digest: str, repository: str) -> Any:
+def image_already_exists(args, repository: str) -> Any:
     """Function to check if a containerImage with the given digest and repository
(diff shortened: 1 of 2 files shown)


mmalina:
Actually, it turns out I wasn't reading the code correctly. My PR is not needed. The check works fine, because manifest_schema2_digest is set even for multiarch images. What we do is:

* Always set repositories.manifest_schema2_digest
* For multiarch, also set repositories.manifest_list_digest

Given that repositories.manifest_schema2_digest is always the unique image digest (not the manifest list one), using it in the existing image search is fine.

PR #328, opened 2024-11-27
fix(KFLUXBUGS-1846): fail get-image-architectures if skopeo failed

mmalina:
If skopeo inspect fails, the variable would be empty and then the jq query would return an empty string instead of null, so we would assume it's an oci artifact and run oras resolve.

If we were unable to fetch the image details with skopeo inspect we should just fail the script.


johnbieren (approved):
nice

PR #329, opened 2024-11-29
feat(ISV-5321): update SPDX document name value

Allda:
This commit sets a name field of the component image SBOM to image pullspec made of repository and image digest. The pullspec is parsed from image purl.

The name is originally set in the build phase but it refers to an internal Quay repository. In the release time we need to set it to a public repository name that is available to customer.

JIRA: ISV-5321

### DIFF
--- a/sbom/update_component_sbom.py
+++ b/sbom/update_component_sbom.py
@@ -5,10 +5,12 @@ This script updates the purls in component-level SBOMs with release time info.
 import argparse
 import glob
 import json
+import logging
 import os
 from collections import defaultdict
 from typing import DefaultDict, Dict, List
-import logging
+
+from packageurl import PackageURL
 
 LOG = logging.getLogger("update_component_sbom")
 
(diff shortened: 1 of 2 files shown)


sbom/update_component_sbom.py
```diff
@@ -54,26 +56,48 @@ def update_cyclonedx_sbom(sbom: Dict, component_to_purls_map: Dict[str, List[str
...
+    the version in the purl.
+
+    Args:
+        purl (str): A package URL for an image.
+
+    Returns:
+        str: Image pullspec with repository URL and digest.
+    """
+    parsed_purl = PackageURL.from_string(purl)
+    repository = parsed_purl.qualifiers.get("repository_url")
```
mmalina, line 73: I wonder how sure we are this will be there. I have a sample spdx sbom locally and it didn't have `repository_url`. Do you think this will be included once they switch to spdx?

Also, using `get()` instead of `[]` suggests that we want to be safe and want it to work even if it's not available. But here the resulting string would be "None@version". If we do require it, wouldn't it be better not to use `get()` and just fail there?

And what about `version`? Are we sure that one will be set to the sha? And in the first format (including `sha:` prefix)?

Allda: The purl is generated in the previous step and we know it has a `repository_url`.
https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/populate-release-notes-images/populate-release-notes-images.yaml#L80

If you want I can switch it from `.get()` to using `[]`. I just wanted to play safe here.

The version is represented by the digest in the `oci` purls (https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#oci). From the previous link you can see it include the `sha` prefix as well.

mmalina (approved):

sbom/update_component_sbom.py
Allda: I changed it to `[]`. Can you merge it now? Thanks

mmalina (approved):

PR #333, opened 2024-12-04
feat(ISV-5447): add multi-arch support to component sbom update

wcheang:


utils/get-image-architectures
```diff
@@ -37,7 +37,7 @@ if [ "$ARTIFACT_TYPE" != "null" ] ; then
     # Just report that the image is for linux/amd64, which is not exactly true - but,
     # downstream release-service-catalog tasks expect to find something. Use this as a default.
     jq -cr -n --arg digest "$digest" \
-        '{"platform": {"architecture": "amd64", "os": "linux"}, "digest": $ARGS.named["digest"]}'
+        '{"platform": {"architecture": "amd64", "os": "linux"}, "digest": $ARGS.named["digest"], "multiarch": false}'
```
mmalina, line 40: I was worried if this breaks something - the script is used in many places. I checked all the usage of this in the catalog repo and it seems it should be ok. I think ideally we should refactor this and return one json that would say this at the top level. But I guess this is good enough for now.

sbom/update_component_sbom.py
```diff
@@ -30,7 +37,25 @@ def get_component_to_purls_map(images: List[Dict]) -> Dict[str, List[str]]:
     for image in images:
         component = image["component"]
         purl = image["purl"]
-        component_purls[component].append(purl)
+        arch = image.get("arch")
+        multiarch = image.get("multiarch", False)
+
+        if multiarch and arch:
+            # replace sha for index purl
+            index_sha = image.get("imageSha")
```
mmalina, line 45: Where is this coming from? Based on the above, it looks like this is the output from get-image-architectures, right? But that one won't have `imageSha`. For multiarch, it will be result of `skopeo inspect`. Here's an example:
```
skopeo inspect --raw docker://registry.redhat.io/ansible-automation-platform-23/aap-must-gather-rhel8@sha256:133fa0b20ad42831b64653af3119e1b6ec9daf17d70f623a49febfd08f42e3e1 | jq '.manifests[]'
{
  "digest": "sha256:bef9b68a7abb08e717f35bd2416aca0a157f1a233a2b53c3dd36de6928bd0c45",
  "mediaType": "application/vnd.docker.distribution.manifest.v2+json",
  "platform": {
    "architecture": "amd64",
    "os": "linux"
  },
  "size": 595
}```

mmalina: Ah, I see it's added in the other PR: https://github.com/konflux-ci/release-service-catalog/pull/724/

So these two PRs both rely on each other? The catalog PR needs the get-image-architecture update and the utils PR needs the populate-release-notes-images

wcheang:
My PTO is starting tomorrow, so @jedinym from my team will be taking over getting this PR merged.

mmalina (approved):

PR #343, opened 2024-12-13
fix(RELEASE-1345): avoid failing on invalid purl string

mmalina:
The sboms can contain invalid non-rpm purl strings and we would fail on parsing them using the packageurl module.

Avoid these failures by parsing the purl type in our own function and only pass it to packageurl if it's an rpm purl. Otherwise skip it.


pyxis/upload_rpm_data.py
```diff
@@ -271,6 +272,34 @@ def construct_rpm_items_and_content_sets(
...
+    Copied and adapted from packageurl package. The reason we need this function
+    and cannot simply use the type component of
+    packageurl.PackageURL.from_string(purl) is that there can be invalid non-rpm
+    purls generated by syft. By getting just the type first and skipping those
+    purls, we avoid failing on those invalid purls.
+
+    Raise ValueError on errors.
+    """
+    scheme, sep, remainder = purl.partition(":")
+    if not sep or scheme != "pkg":
```
johnbieren, line 288: Now we are maintaining a list of types that a function we do not maintain supports. Wouldn't it be easier to just
```
try:
    purl_dict = PackageURL.from_string(externalRef["referenceLocator"]).to_dict()
catch:
    printf("%s is not a valid purl so is not being added to the rpm_data", purl_dict)
    continue
```
?

mmalina: Not sure if you thought it was, but `pkg` is not a type. `pkg` is called a scheme here. Just saying.

I did consider what you suggest. The problem is that this would also skip invalid rpm purls if we ever encounter one. And if that happens, we should fail. At least I think we should. It might be ok to just skip those as well, but I don't know about that. So my solution seems safer (but yeah, at the expense of more complexity).

johnbieren: Okay I think I misunderstood. So, currently we fail on invalid non-rpm purls and invalid rpm purls. This PR skips invalid non-rpm purls but we still fail on invalid rpm urls?

mmalina: Correct. But so far all the user reports of this issue I've seen were caused non-rpm purls, so this will help in 100 % of the cases we've seen so far. But of course you never know if we encounter an invalid rpm purl in the future (and we will still fail on that and we probably should).

pyxis/upload_rpm_data_cyclonedx.py
```diff
@@ -292,6 +293,34 @@ def construct_rpm_items_and_content_sets(
     return rpms_items, sorted(content_sets)
 
 
+def get_purl_type(purl: str):
```
johnbieren, line 296: Quick eye test says this is the same function. Can't we just import it in one instead of defining it in both python files?

pyxis/test_upload_rpm_data_cyclonedx.py
johnbieren, line 1: Can you add a test where get_purl_type throws an error?

mmalina: ok, added. but it's a completely made up value - I've never seen one without type.

pyxis/upload_rpm_data_cyclonedx.py
johnbieren: That's fair. I'd say that is a reason to import it in cyclonedx and just define it in spdx, but up to you

mmalina: Here's the Jira for removing support for cyclonedx: https://issues.redhat.com/browse/RELEASE-1292

johnbieren (approved):

pyxis/upload_rpm_data_cyclonedx.py
mmalina: I just got some updates on the spdx work in build team. Turns out they will continue to support both. I commented on the Jira. We should probably refactor this then. But that will happen later.

For now I'd prefer to keep everything duplicated.

PR #351, opened 2025-01-08
fix(RELEASE-1354): path for internal pipelines

scoheb:
- should now be /internal/


mmalina (approved):

johnbieren:
Once the tasks using this script are updated to the image with this change, we can remove the sym link https://github.com/konflux-ci/release-service-catalog/blob/development/pipelines/simple-signing-pipeline/simple-signing-pipeline.yaml

PR #354, opened 2025-01-17
feat(RELEASE-1327): update internal-request script to support resolvers

johnbieren:
This commit updates the utils/internal-request script to support calling pipelines via resolvers. It keeps functionality for using request, although the goal is to only keep it temporarily. I did want to keep request support though in case a bug is found in the utils image before the catalog task hitting the bug has a chance to be migrated to using pipeline instead of request.


utils/internal-request
```diff
@@ -55,24 +61,30 @@ PARAMS=""
 PIPELINE_TIMEOUT=1h0m0s
 TASK_TIMEOUT=0h55m0s
 FINALLY_TIMEOUT=0h5m0s
+RESOLVER=git
 
 function usage {
-    echo "Usage: $0 -r request [-p parameters] [-l labels] [-s sync] [-t timeout] [--service-account name] [--pipeline-timeout 1h0m0s] [--task-timeout 0h55m0s] [--finally-timeout 0h5m0s]"
+    echo "Usage: $0 [-r request] [--pipeline check-embargoed-cves] [-p parameters] [-l labels] [-s sync] [-t timeout] [--service-account name] [--pipeline-timeout 1h0m0s] [--task-timeout 0h55m0s] [--finally-timeout 0h5m0s] [--resolver git]"
```
mmalina, line 67: Same as above:

Why `check-embargoed-cves`? Should probably say `pipeline` or `pipeline-name`.

utils/internal-request
```diff
@@ -138,11 +166,24 @@ while true; do
...
 
+# Make sure resolver is a supported value
+# The -n PIPELINE can be dropped once PIPELINE is mandatory
+if [[ -n "$PIPELINE" ]]; then
+    if [[ "$RESOLVER" != "git" ]] && [[ "$RESOLVER" != "cluster" ]] ; then
+        echo "Only git and cluster resolvers are supported at this time. Please use one of those."
+        exit 1
+    fi
+    if [[ "$RESOLVER" == "git" ]] && ([ -z "$TASKGITURL" ] || [ -z "$TASKGITREVISION" ]) ; then
+        echo "You must pass --task-git-url and --task-git-revision when using a pipeline resolver."
```
mmalina, line 182: Should be "when using git resolver."?

utils/internal-request
```diff
@@ -199,22 +240,70 @@ LABEL_JSON=$(echo "${LABEL_JSON_ARRAY[@]}" | jq -s 'add')
...
       "spec": {
-        "request": $request,
         "params": $parameters,
         "timeouts": $timeouts
       }
     }'
 )
+# We will inject the pipeline stuff as we did request (without the if)
+# when we drop support for request
+if [[ -n "${REQUEST}" ]]; then
```
mmalina, line 256: Here's an idea: How about we treat `$REQUEST` as a shorthand for `RESOLVER=cluster` and `PIPELINE=$request`? (This could be done at the option level directly, so we wouldn't even have a `$REQUEST` variable.) Then we don't need separate handling for it.

In the description you say you want to avoid breakage of tasks using `request` in case there is a bug in this script. But if such a task uses this updated script (while still using request) there will always be a risk because this commit changes a lot of things. So I don't think my proposal would make it more risky.

utils/internal-request
```diff
@@ -109,6 +125,14 @@ while true; do
             shift
             TIMEOUT=$1
             ;;
+        --task-git-url)
```
mmalina, line 128: I would suggest to call these `git-url` and `git-revision` or `git-resolver-url` and `git-resolver-revision` - they are not used for tasks, so I don't think it makes sense to have `task` in the name. I understand they are copied from the task git parameters in the task that calls this, but there is no reason that this script should we concerned with that.

As I say elsewhere, ideally the internal-services operator would be able to also reuse these for the PLR git parameters in the future (once we only support git resolvers).

utils/internal-request
johnbieren: Because then we still need to keep the big and repetitive if statements like 
`if [[ -n "${PIPELINE}" ]]; then`. It is very ugly. We are dropping `request` from the CRD and only us (release team) uses this script. I don't see a good reason to keep it

utils/internal-request
johnbieren: I disagree, I copied them from `internal-pipelinerun` for consistency

utils/internal-request
johnbieren: nice catch, updated

utils/internal-request
johnbieren: updated

utils/internal-request
```diff
@@ -15,27 +15,33 @@
 # for status updates.
 #
 # Usage:
-#   ./internal-request.sh -r request [-p <key=value> ...] [-s sync] [-t timeout] [--service-account name]
-#   [--pipeline-timeout 1h0m0s] [--task-timeout 0h55m0s] [--finally-timeout 0h5m0s]
+#   ./internal-request.sh [-r request] [--pipeline check-embargoed-cves] [-p <key=value> ...] [-s sync]
```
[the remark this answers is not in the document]
johnbieren: updated

utils/internal-request
johnbieren: And they are reused, via this script though not the operator. Same as `internal-pipelinerun`
```
        --arg url "$TASKGITURL" \
        --arg revision "$TASKGITREVISION" \
        --arg pipeline "pipelines/internal/${PIPELINE}/${PIPELINE}.yaml" \
        '.spec.pipeline.pipelineRef.params = [
          {
            "name": "url",
            "value": $url
          },
          {
            "name": "revision",
            "value": $revision
```

utils/internal-request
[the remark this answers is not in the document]
mmalina: OK, now I see that the timeouts are real examples. Other things aren't. So I guess fine to keep, but I would still prefer to change it (and you already did - thanks).

utils/internal-request
mmalina: OK. What about the other question - should we even support cluster resolvers at all going forward?

johnbieren: I don't really see any good reason to support them, but it is only a few extra lines of code to do so. So, I am basically open to either

utils/internal-request
mmalina: > for consistency

OK, that's a good reason.

> they are reused

What you pasted is the main usage of them. What I meant by "reuse" is to also add the extra pipeline params (not pipelineRef) based on them. At least I think that's what's happening.

Look at this: https://github.com/konflux-ci/release-service-catalog/blob/development/pipelines/managed/rh-advisories/rh-advisories.yaml#L59 - that's the `taskGitRevision` pipeline parameter. And it's appended to the pipeline params in the operator here, based on the git revision of the pipelineRef:

https://github.com/konflux-ci/release-service/blob/main/tekton/utils/pipeline_run_builder.go#L200

Another option would be to do this in this script directly - that way we could still avoid setting duplicate params when calling this script.

Currently when using internal-pipelinerun the params are added twice (and it will be the case here as well):

https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/rh-sign-image/rh-sign-image.yaml

```
        if [ "${REQUESTTYPE}" == "internal-pipelinerun" ] ; then
          requestType=internal-pipelinerun
          EXTRA_ARGS=(
          --task-git-url "$(params.taskGitUrl)"
          --task-git-revision "$(params.taskGitRevision)"
          --service-account "${service_account_name}"
          )
```
```
-p taskGitUrl="$(params.taskGitUrl)" \
-p taskGitRevision="$(params.taskGitRevision)" \
```

So yeah, maybe better to stay consistent and for both scripts do this in the script itself? That is: Find the git parameters in the `-p` parameters and take them from there?

It can be a subsequent Jira. WDYT?

utils/internal-request
johnbieren: Removed `resolver` as an argument and hardcoded it to git

utils/internal-request
johnbieren: Gotcha. I for whatever reason thought we already did this in the script, but you are right, we don't. I think a subsequent jira is appropriate, but yes we should certainly do it. Easiest may be to just keep the `--task-git-url` and `--task-git-revision` options passed to this script and then inject them in the params?

mmalina: Created https://issues.redhat.com/browse/RELEASE-1380

mmalina (approved):

PR #357, opened 2025-01-21
feat(RELEASE-1380): determine internal-* pipelineRef from params

johnbieren:
This commit updates the internal-request and internal-pipelinerun scripts to determine the git url and git revision to use for the pipelineRef in the git resolver based on the taskGitUrl and taskGitRevision parameters.


utils/internal-pipelinerun
```diff
@@ -195,8 +181,19 @@ do
...
     VALUE=$(echo "$param" | cut -d'=' -f2-)
     PARAM_JSON_ARRAY+=("$(jq -n --arg key "$KEY" --arg value "$VALUE" '.name=$key | .value=$value'),")
+    if [[ "$KEY" = "taskGitUrl" ]]; then
+        PIPELINEGITURL="$VALUE"
+    elif [[ "$KEY" = "taskGitRevision" ]]; then
+        PIPELINEGITREVISION="$VALUE"
+    fi
 done
 
+if [ -z "$PIPELINEGITURL" ] || [ -z "$PIPELINEGITREVISION" ] ; then
```
mmalina, line 191: I'd suggest to make sure these are defined. This script doesn't have anything like `set -eux`, would be nice to have it, but of course at this point it might be risky to add it without thorough testing. But even without it, it's nice to define them before accessing. So I would add this somewhere before the `for` loop:
```sh
PIPELINEGITURL=
PIPELINEGITREVISION=
```

johnbieren: added

mmalina: For completeness, I just realized there is also a test for existence of a variable:
```
[[ -v foo ]]
```
But setting it to empty definitely works.

mmalina (approved):

PR #361, opened 2025-02-03
feat(RELEASE-1246): implement script to push files to CGW

seanconroy2021:
This script automates pushing files to the CGW:

- Reads metadata and processes files from the content directory.
- Checks if a file already exists in the product version based on the label, short URL, and download URL, skipping duplicates to avoid failure.
- Creates new files and returns their IDs upon success.
- Generates a JSON report summarising the number of files created, skipped, and the metadata used.

RELATES TO: https://github.com/konflux-ci/release-service-catalog/pull/781

### DIFF
--- a/publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
+++ b/publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
@@ -1,0 +1,346 @@
+#!/usr/bin/env python3
+"""
+This script interacts with the Content Gateway (CGW) API to create and manage content files.
+It ensures each file is checked before creation and skips files that already exist.
+The script is idempotent,it can be executed multiple times as long as the label,
+short URL, and download URL remain unchanged.
+
+### **Functionality:**
+1. Reads a JSON metadata file and a directory containing content files.
+2. Retrieves the product ID using the provided product name and product code.
+3. Retrieves the version ID using the product version name.
+4. Generates metadata for each file in the content directory.
+5. Checks for existing files and skips them if they match the label, short URL, and download
+URL.
+6. Creates new files using the metadata.
+7. Rolls back created files if an error occurs during execution.
+8. Writes the final result, including processed, created, and skipped files, to a JSON file.
+9. Outputs the path of the generated result.json file to an output file.
+"""
+
+import os
+import argparse
+import json
+import hashlib
+import logging
+import requests
+from requests.auth import HTTPBasicAuth
+
+# Default values for each component,
+# values from data_file takes precedence over these
+default_values_per_component = {
+    "type": "FILE",
+    "hidden": False,
+    "invisible": False,
+}
+
+logging.basicConfig(level=logging.INFO, format="%(asctime)s - %(levelname)s - %(message)s")
+
+
+def parse_args():
+    """Parse command line arguments."""
+    parser = argparse.ArgumentParser(
+        prog="publish_to_cgw_wrapper", description="Publish content to the Content Gateway"
+    )
+    parser.add_argument(
+        "--cgw_host",
+        required=True,
+        help="The hostname of the content-gateway to publish the metadata to",
+    )
+    parser.add_argument(
+        "--data_file",
+        required=True,
+        help="Path to the JSON file containing merged data",
+    )
+    parser.add_argument(
+        "--content_dir",
+        required=True,
+        help="Path to the directory containing content to push",
+    )
+    parser.add_argument(
+        "--output_file",
+        required=True,
+        help="Path to the file which write the result.json file path",
+    )
+
+    return parser.parse_args()
+
+
+def call_cgw_api(*, host, method, endpoint, session, data=None):
+    """Make an API call to the Content Gateway service."""
+    try:
+        response = session.request(
+            method=method.upper(),
+            url=f"{host}{endpoint}",
+            json=data,
+        )
+
+        if not response.ok:
+            error_message = (
+                response.text.strip() or f"HTTP {response.status_code}:{response.reason}"
+            )
+            raise RuntimeError(f"API call failed: {error_message}")
+
+        return response
+    except requests.RequestException as e:
+        raise RuntimeError(f"API call failed: {e}")
+
+
+def get_product_id(*, host, session, product_name, product_code):
+    """Retrieve the product ID by name and product code."""
+    products = call_cgw_api(host=host, method="GET", endpoint="/products", session=session)
+    products = products.json()
+    for product in products:
+        if product.get("name") == product_name and product.get("productCode") == product_code:
+            logging.info(f"Found product: {product_name} with ID {product.get('id')}")
+            return product.get("id")
+    raise ValueError(f"Product {product_name} not found with product code {product_code}")
+
+
+def get_version_id(*, host, session, product_id, version_name):
+    """Retrieve the version ID for a specific product."""
+    versions = call_cgw_api(
+        host=host, method="GET", endpoint=f"/products/{product_id}/versions", session=session
+    )
+    versions = versions.json()
+    for version in versions:
+        if version.get("versionName") == version_name:
+            logging.info(f"Found version: {version_name} with ID {version.get('id')}")
+            return version.get("id")
+    raise ValueError(f"Version not found: {version_name}")
+
+
+def generate_download_url(content_dir, file_name):
+    """
+    Generate a download URL in this format:
+    /content/origin/files/sha256/{checksum[:2]}{checksum}/{file_name}
+    """
+    prefix = "/content/origin/files/sha256"
+    sha256_hash = hashlib.sha256()
+    with open(content_dir + "/" + file_name, "rb") as f:
+        for byte_block in iter(lambda: f.read(4096), b""):
+            sha256_hash.update(byte_block)
+    checksum = sha256_hash.hexdigest()
+    return f"{prefix}/{checksum[:2]}/{checksum}/{file_name}"
+
+
+def generate_metadata(
+    *, content_dir, components, product_Code, version_id, version_name, mirror_openshift_Push
+):
+    """
+    Generate metadata for each file in
+    content_list that starts with the component name
+    """
+    shortURL_base = "/pub/"
+    if mirror_openshift_Push:
+        shortURL_base = "/pub/cgw"
+    metadata = []
+    shasum_files_processed = []
+    logging.info(f"Generating metadata for files in {content_dir}")
+    for file in os.listdir(content_dir):
+        matching_component = None
+        for component in components:
+            if file.startswith(component["name"]):
+                matching_component = component.copy()
+                break
+
+        if matching_component:
+            logging.info(f"Processing file: {file}")
+            matching_component.update(
+                {
+                    "productVersionId": version_id,
+                    "downloadURL": generate_download_url(content_dir, file),
+                    "shortURL": f"{shortURL_base}/{product_Code}/{version_name}/{file}",
+                    "label": file,
+                }
+            )
+            del matching_component["name"]
+            metadata.append(
+                {"type": "file", **default_values_per_component, **matching_component}
+            )
+        else:
+            if file.startswith("sha256") and file not in shasum_files_processed:
+                shasum_files_processed.append(file)
+                logging.info(f"Processing file: {file}")
+                if file.endswith(".gpg"):
+                    label = "Checksum - GPG"
+                elif file.endswith(".sig"):
+                    label = "Checksum - Signature"
+                elif file.endswith(".txt"):
+                    label = "Checksum"
+
+                metadata.append(
+                    {
+                        "productVersionId": version_id,
+                        "downloadURL": generate_download_url(content_dir, file),
+                        "shortURL": f"{shortURL_base}/{product_Code}/{version_name}/{file}",
+                        "label": label,
+                        **default_values_per_component,
+                    }
+                )
+            else:
+                # Skip files that do not start with any component name or
+                # sha256
+                logging.info(
+                    f"Skipping file: {file} as it does not start with any component name"
+                )
+                continue
+
+    return metadata
+
+
+def file_already_exists(existing_files, new_file):
+    """Check if a file already exists"""
+    for file in existing_files:
+        if all(
+            file.get(key) == new_file.get(key) for key in ["label", "downloadURL", "shortURL"]
+        ):
+            return file
+    return None
+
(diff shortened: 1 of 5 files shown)


publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
mmalina, line 1: Is this basically a copy of the python script that was originally in the task? Or is it that plus changes for the idempotency? Or anything else?

seanconroy2021: The only thing that is the same is `generate_download_url `and `generate_metadata` with the new script calling the endpoint itself instead of using pub tools. The way it works now is it gets the product ID and product version, creates the metadata, and then checks each item in metadata against the existing files in the product version based on the label, short URL and download URL if they match, it will skip the creation of the item.

mmalina:
@swickersh @pkhander can you review this please?

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
mmalina: ok, thanks. it would be good if Scott or Parthey could review. I pinged them here in the PR.

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -0,0 +1,317 @@
...
+It ensures each file is checked before creation and skips files that already exist.
+The script is idempotent,it can be executed multiple times as long as the label,
+short URL, and download URL remain unchanged.
+
+### **Functionality:**
+1. Reads a JSON metadata file and a directory containing content files.
+2. Retrieves the product ID using the provided product name and product code.
+3. Retrieves the version ID using the product version name.
+4. Generates metadata for each file in the content directory.
+5. Checks for existing files and skips them if they match the label, short URL, and download
```
swickersh, line 13: I think this is more complex than what we discussed.
Instead of checking for existing files and skipping, it's more efficient to wrap the original call in a try/except block.
If (and only if) the error returned from Content Gateway is something along the lines of `Short URL is already present in the system!` or `File is already present in the system!` then we would retry the API call as an `UPDATE`.

Any other failures should not be ignored or trigger a rollback as anything else would be unexpected and we would want to fail and investigate.

pkhander: +1 to use a try.except block and `UPDATE` on those two cases. That will save extra API calls to get all the info from CGW every time. 

I like the rollback option thou! This ensures that we do not release "half" content when a release fails.

swickersh: I guess the roll back functionality could be kept. So just to clarify the workflow would be something like this:

1. try to create files
3. if we get `already present` error,  attempt the api call again, but this time as `UPDATE`
4. if we get any other error, attempt to rollback (delete) whatever files were just successfully pushed during this pipeline and then exit/fail the pipelinerun.

seanconroy2021: So the idea was that if the file is already present due to a short URL or download URL, there is no need to update it; instead, just skip ..... as the content should not to change upon re-run.

swickersh: Parthey and I synced up real quick.  Since the main point of this effort is to be idempotent, then we can skip the small edge case of updating. (That is just a nice-to-have if something were tweaked)
But I still just wanted to be clear, you plan to remove the stuff where you're fetching if the files exist first or not, correct? Because you can simply determine they exist when you attempt to create them and are met with a 'already present' error from the API.

Also, when you are iterating through the files, if one exists already and warrants skipping, can you print a line to the logs stating it was skipped for that reason.

seanconroy2021: Yeah, I will remove the fetching of the files and switch to try/except to see if the file can be created, then output the success or skip.

It would be nice to have the update, but without getting each file ID,  I can't call the update endpoint.

johnbieren: To clarify, a try/except of the script will unfortunately *not* work for making this idempotent for us. Here's why - each task that pushes content must be made idempotent, yes, and try/except should work for that. However, we also have tasks to make each pipeline idempotent. This means, when the pipeline begins, we run the final task's idempotent check and remove any already pushed components from the snapshot. The key here is that if the component was already pushed, it will now no longer be part of the EC validation. This is very important. Of course, we can't start a pipeline, try/except push all content to cgw, and then proceed like normal. That "start of pipeline" idempotent check must be a simple "is this stuff already done? yes/no" type of check

seanconroy2021: @johnbieren So what would work here would be to pull all the files from the product version and check them against the metadata. If all files are present, then we exit gracefully with a message else, we continue and create the new files.

johnbieren: Sure - you know the specifics way better than me. All I know is that "try to push and see what happens" can't be our idempotent check for any of our tasks

swickersh: @johnbieren 
Interesting.  After reading KONFLUX-4137 I see the intent here. But I'm still confused on the implementation suggested.

>when the pipeline begins, we run the final task's idempotent check and remove any already pushed components from the snapshot.
So currently the pipeline `push-binaries-to-dev-portal` is one that uses this task.  Are you saying when that pipeline begins, it will run _only_ this new idempotent section of the content-gateway task and remove already published components if necessary? Or will this be a new idempotent-check task altogether?

johnbieren: I don't know offhand that the pipelines that use this are, but the general idea is this.
For each pipeline, what is the last "push" type activity in it? Take that, and run the idempotent check for it first. Because, if the image is already done for that activity, it must be done for *all* activities in the pipeline, so we can completely disregard it and remove it from the snapshot. However, if one single task still needs to be done for it, we must include it in the pipeline flow and let the individual tasks' idempotent behaviors handle it.

One of the biggest pushes for making the pipelines idempotent is "I have this image in my snapshot that I already released 4 months ago and haven't changed, but now I am failing my release pipeline because EC fails on the old image that doesn't even need to be processed again"

Each pipeline will have its own specific "start of pipeline idempotent check" task based on its last "push activity"

swickersh: Thanks for clarifying.

 In order to check this at the beginning, you'd need to know _what_ to check. The extract-binaries task that runs _before_ the CGW task [places files into a content directory](https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/publish-to-cgw/publish-to-cgw.yaml#L61)  and then the CGW task works with those. 

If I'm thinking about this correctly, you couldn't check if the CGW publish exists without knowing the filename. And the filename is derived from the binary extracted. The binaries extracted depend on collect-data to know which components in the snapshot are intended to be released to CGW. So running this check at the beginning might have some complexity to consider.  @pkhander can you please verify this. I don't want to over-complicate anything for Sean.

But even if there's some overhead to check idempotent, it's still worthwhile because this would save the signing tasks and cdn tasks from running unnecessarily.

johnbieren: This is going to be a common pattern (and a pain for us to implement). The fbc pushes depend on the ocp version we fetch in another task. The advisory stuff depends on tags, which we don't evaluate until apply-mapping. So, it won't be as clean as "just do this task first" but we will try to get as close to that as possible.

That work is an entirely separate ticket though. My commenting here was just to say that we can't try/except as our idempotent check, we need to do some pull and see if the content is already there

seanconroy2021: Your grand, don't worry about it. Thanks for all the help 😊
Just to close this out and make sure we are all on the same page, the current script in this pr will check each file before creating. If it's present, it will skip.

Example below:

`[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,468 - INFO - Skipping creation: File release-darwin-arm64.gz already exists with ShortURL /pub/cgw/product_code_1/1.2/release-darwin-arm64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,468 - INFO - Creating file: cosign with ShortURL /pub/cgw/product_code_1/1.2/cosign
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,469 - INFO - Succesfully created file with ID: 105
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,469 - INFO - Creating file: cosign-darwin-amd64.gz with ShortURL /pub/cgw/product_code_1/1.2/cosign-darwin-amd64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,471 - INFO - Succesfully created file with ID: 106
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,471 - INFO - Creating file: cosign-linux-arm64.gz with ShortURL /pub/cgw/product_code_1/1.2/cosign-linux-arm64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,472 - INFO - Succesfully created file with ID: 107
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,472 - INFO - Skipping creation: File release-linux-amd64.gz already exists with ShortURL /pub/cgw/product_code_1/1.2/release-linux-amd64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,472 - INFO - Skipping creation: File release-darwin-amd64.gz already exists with ShortURL /pub/cgw/product_code_1/1.2/release-darwin-amd64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,472 - INFO - Creating file: cosign-darwin-arm64.gz with ShortURL /pub/cgw/product_code_1/1.2/cosign-darwin-arm64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,474 - INFO - Succesfully created file with ID: 108
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,474 - INFO - Creating file: cosign-linux-amd64.gz with ShortURL /pub/cgw/product_code_1/1.2/cosign-linux-amd64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,475 - INFO - Succesfully created file with ID: 109
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,475 - INFO - Skipping creation: File release-linux-arm64.gz already exists with ShortURL /pub/cgw/product_code_1/1.2/release-linux-arm64.gz
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,475 - INFO - Skipping creation: File release already exists with ShortURL /pub/cgw/product_code_1/1.2/release
[run-task : run-push-cgw-metadata] 2025-02-04 09:44:39,475 - INFO - Created 5 files and skipped 5 files`

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -0,0 +1,317 @@
...
+URL.
+6. Creates new files using the metadata.
+7. Rolls back created files if an error occurs during execution.
+8. Writes the final result, including processed, created, and skipped files, to a JSON file.
+9. Outputs the path of the generated result.json file to the an output file.
```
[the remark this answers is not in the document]
seanconroy2021: fixed

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -0,0 +1,317 @@
...
+import os
+import argparse
+import json
+import hashlib
+import logging
+import requests
+from requests.auth import HTTPBasicAuth
+
+# Default values for each component,
+# values from data_file takes presedence over these
```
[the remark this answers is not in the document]
seanconroy2021: fixed

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -0,0 +1,317 @@
...
+
+        USERNAME = os.getenv("CGW_USERNAME")
+        PASSWORD = os.getenv("CGW_PASSWORD")
+
+        if not USERNAME or not PASSWORD:
+            raise ValueError(
+                "CGW_USERNAME and CGW_PASSWORD environment variables are required"
+            )
+
+        auth = HTTPBasicAuth(USERNAME, PASSWORD)
```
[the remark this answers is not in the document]
seanconroy2021: Fixed

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -0,0 +1,317 @@
...
+    sha256_hash = hashlib.sha256()
+    with open(content_dir + "/" + file_name, "rb") as f:
+        for byte_block in iter(lambda: f.read(4096), b""):
+            sha256_hash.update(byte_block)
+    checksum = sha256_hash.hexdigest()
+    return f"{prefix}/{checksum[:2]}/{checksum}/{file_name}"
+
+
+def generate_metadata(
+    content_dir, components, product_Code, version_id, version_name, mirror_openshift_Push
```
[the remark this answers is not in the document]
seanconroy2021: Fixed

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
seanconroy2021: @johnbieren

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
johnbieren, line 1: This file LGTM. It isn't in scope of your *task* ticket, but it would help the *pipeline* idempotent ticket if you added like a `--dry-run` arg and if dry_run is true, create_files only does the check and doesn't do the create. Not something you need to add in this PR, but something to keep in mind for later

johnbieren (approved):

seanconroy2021:
hey, @swickersh whenever you get a chance could you review this again :)

swickersh:
> hey, @swickersh whenever you get a chance could you review this again :)

LGTM 
I'm not on approve list for this repo FYI.

swickersh (approved):
Oh, maybe I can approve in this project.

PR #365, opened 2025-02-06
feat(CLOUDDST-25757): Added pubtools-sign and pubtools-pyxis deps

midnightercz:
Added pubtools-sign and pubtools-pyxis to Dockerfile as dependency to enable signing with pubtools-sign


mmalina:
Commit message needs lower case after the `feat:`. Also, ideally it should be imperative form, so I'd recommend `feat(CLOUDDST-25757): add pubtools-sign and pubtools-pyxis deps` (that saves a few characters as well)

johnbieren (approved):

PR #384, opened 2025-02-26
fix(RELEASE-1455): ignore purls without qualifiers in spdx sboms

mmalina:
The script was failing for a user when it tried to access qualifiers for a purl dict and there were none. I discussed this with the build team and they agreed that we should ignore such entries. In most cases this will be a duplicate package entry anyway. More details in Jira.

### DIFF
--- a/pyxis/upload_rpm_data.py
+++ b/pyxis/upload_rpm_data.py
@@ -245,6 +245,8 @@ def construct_rpm_items_and_content_sets(
             purl_dict = PackageURL.from_string(externalRef["referenceLocator"]).to_dict()
             if purl_dict["name"] in IGNORED_PACKAGES:
                 continue
+            if "qualifiers" not in purl_dict or purl_dict["qualifiers"] is None:
+                continue
             rpm_item = {
                 "name": purl_dict["name"],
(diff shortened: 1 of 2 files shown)


pyxis/upload_rpm_data.py
```diff
@@ -245,6 +245,8 @@ def construct_rpm_items_and_content_sets(
             purl_dict = PackageURL.from_string(externalRef["referenceLocator"]).to_dict()
             if purl_dict["name"] in IGNORED_PACKAGES:
                 continue
+            if "qualifiers" not in purl_dict:
```
johnbieren, line 248: Can you add to PACKAGES in the test file for this?

johnbieren (approved):

pyxis/upload_rpm_data.py
mmalina: Ah, I forgot we had all the use cases there - I only saw one test so assumed we didn't have tests for all of the paths. I was wrong :) 

And actually a good call, because my condition didn't work! It turns out that `purl_dict["qualifiers"]` would be there, but equal to `None`. Now I am checking for both for good measure.

PR #385, opened 2025-02-27
fix(ISV-5685): fix component-product relationship

jedinym:
The relationship in the product-level SBOM was the wrong way around. I will update the image tags in release-service-catalog when this is merged.

### DIFF
--- a/sbom/create_product_sbom.py
+++ b/sbom/create_product_sbom.py
@@ -85,9 +85,9 @@ def get_component_relationships(packages: List[Dict]):
     """Get SPDX relationship for each SPDX component package."""
     return [
         {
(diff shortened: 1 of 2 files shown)


johnbieren (approved):
/ok-to-test
(I doubt that is going to work. @mmalina can you just merge this when you get a chance to review it if it still says pending?)

mmalina:
Hmm, I thought this worked in the past, not sure why it's not working now. Trying again.

mmalina (approved):

mmalina:
Hmm, apparently it won't work. Merging.

mmalina:
@jedinym as a reminder, you'll also need an additional PR for the catalog to update the image that is used in the task.

PR #390, opened 2025-03-06
refactor(RELEASE-1340): rename request to pipeline in internal-pipeli…

johnbieren:
…nerun script

This is for parity with the internal-request script, which was already updated.

### DIFF
--- a/utils/internal-pipelinerun
+++ b/utils/internal-pipelinerun
@@ -15,11 +15,11 @@
 # Timeout: error code 124
 #
 # Usage:
-#   ./internal-pipelinerun -r request [-p <key=value> ...] [-s sync] [-t timeout] [--service-account name]
+#   ./internal-pipelinerun --pipeline pipeline [-p <key=value> ...] [-s sync] [-t timeout] [--service-account name]
 #   [--pipeline-timeout 1h0m0s] [--task-timeout 0h55m0s] [--finally-timeout 0h5m0s]
 #
 # Parameters:
-#   -r                  Request: the name of the request.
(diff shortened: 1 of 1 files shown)


utils/internal-pipelinerun
```diff
@@ -55,10 +55,10 @@ TASK_TIMEOUT=0h55m0s
 FINALLY_TIMEOUT=0h5m0s
 
 function usage {
-    echo "Usage: $0 -r request [-p parameters] [-l labels] [-s sync] [-t timeout] [--service-account name] \
+    echo "Usage: $0 --pipeline pipeline [-p parameters] [-l labels] [-s sync] [-t timeout] [--service-account name] \
```
mmalina, line 58: Just one question: I was gonna say it would be safer to support both. And I actually found one other place where it's used besides the signing thing: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/push-artifacts-to-cdn/push-artifacts-to-cdn.yaml#L72

So either support both, or make sure it's also updated in that task, so that it's not forgotten and then we'd risk a breakage in the future when bumping the utils image there?

johnbieren: I don't want to support both. I plan to remove support for `-r request` in `internal-request` with this ticket https://issues.redhat.com/browse/RELEASE-1378 
I will update a separate PR to update push-artifacts-to-cdn once this merges

mmalina: ok, sounds good, thanks.

mmalina (approved):

PR #397, opened 2025-03-20
chore(RELEASE-1221): bump oras image to get select-oci-auth update

mmalina:
The updated version supports docker.io.


johnbieren (approved):

mmalina:
> I wonder if we should have any select-oci-auth tests... I would say the argument for no is "the script isn't maintained here, why would we have tests for it" and the argument for yes is "if a new image update would break the script, we'd catch it before merge" (which has happened before)

I added tests for the select-oci-auth script in the oras repo. But yeah, that wouldn't catch the previous breakage. I think it would help to have some sanity test for all the tools we install, yeah. Simple stuff like `pub_tools --help` or whatever we have.

PR #399, opened 2025-03-24
feat(ISV-5760): rework SBOM generation

jedinym:
In this PR, the component-level SBOM update process for SPDX-2.X is rewritten. While before we relied on data from non-SBOM related tasks, now we work with inputs that are defined better.

The only input required to update component-level SBOMs is now the mapped snapshot spec (exported by the `apply-mapping` task).

The snapshot is then parsed into a Python object that contains references to all images (with arch-specific images in the case of multiarch).

This object is then used to asynchronously download the SBOMs for all images and update them based on the snapshot being released.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 pytest
+pytest-asyncio
 requests
 jinja2
 jinja2-ansible-filters
@@ -7,3 +8,6 @@ pubtools-pulp
 pubtools-exodus
 pubtools-content-gateway
 pubtools-marketplacesvm
+pydantic
+aiofiles
+types-aiofiles
--- a/sbom/README.md
+++ b/sbom/README.md
@@ -1,0 +1,30 @@
+# Konflux release-time SBOM generation
+This module contains scripts and libraries used to enrich SBOMs created at
+build-time (component-level) and create SBOMs (product-level) when a snapshot is
+being released.
+
+## Data sources
+### Snapshot spec
+The necessary data source for both component-level and product-level SBOM
+manipulation is the mapped snapshot spec in JSON format. This file is created by
+the `apply-mapping` Tekton Task, which exports a result containing the path to
+the file.
+
+The snapshot spec is parsed using the `sbom.sbomlib.make_snapshot` function into
+a Python representation of the snapshot. The `make_snapshot` function also
+handles multiarch images by fetching manifests of index images (using `oras`)
+and parsing them. The resulting object contains all component repositories and
+digests of images that are being released in the snapshot.
+     
+### Data file
+The product-level SBOM requires additional data to be constructed (such as the
+cpe id, product name and version). This data can be found in the data file,
+which is exported by the `collect-data` Tekton task as path to a file.
+
+## Component-level SBOM enrichment
+When component-level SBOMs are updated, the OCI PURLs generated during
+build-time are stripped and new PURLs are generated based on the parsed snapshot
+data. Non-OCI PURLs are preserved.
+
+## Product-level SBOM creation
+This should be filled once product-level SBOMs use the new inputs.
--- a/sbom/handlers/spdx2.py
+++ b/sbom/handlers/spdx2.py
@@ -1,0 +1,253 @@
+"""
+This module contains the SBOM handler for SPDX version 2 SBOMs.
+"""
+
+from typing import Optional, Union, Any
+
+from packageurl import PackageURL
+
+from sbom.logging import get_sbom_logger
+from sbom.sbomlib import (
+    Component,
+    Image,
+    IndexImage,
+    SBOMError,
+    construct_purl,
+    get_purl_arch,
+    get_purl_digest,
+    make_reference,
+    without_sha_header,
+)
+
+logger = get_sbom_logger()
+
+
+class SPDXPackage:
+    """
+    Wrapper class for easier SPDX package manipulation.
+    """
+
+    def __init__(self, package: Any) -> None:
+        self.package = package
+
+    @property
+    def external_refs(self) -> list[dict[str, Any]]:
+        """
+        Get the externalRefs field of the package.
+        """
+        return self.package.get("externalRefs", [])
+
+    @external_refs.setter
+    def external_refs(self, value: list[Any]) -> None:
+        self.package["externalRefs"] = value
+
+    @property
+    def spdxid(self) -> str:
+        return self.package.get("SPDXID", "UNKNOWN")
+
+    @property
+    def checksums(self) -> list[dict[str, Any]]:
+        """
+        Get the checksums field of the package.
+        """
+        return self.package.get("checksums", [])
+
+    @property
+    def sha256_checksum(self) -> Optional[str]:
+        """
+        Extracts a sha256 checksum from an SPDX package. Returns None if no such
+        checksum is found.
+        """
+        checksums = self.checksums
+        if checksums is None:
+            return None
+
+        for checksum in checksums:
+            if checksum.get("algorithm") == "SHA256":
+                return checksum.get("checksumValue")
+
+        return None
+
+    def update_external_refs(
+        self,
+        digest: str,
+        repository: str,
+        tags: list[str],
+        arch: Optional[str] = None,
+    ) -> None:
+        """
+        Update the external refs of an SPDX package by creating new OCI PURL
+        references and stripping all old OCI PURL references. Other types of
+        externalRefs are preserved.
+        """
+        new_oci_refs = SPDXPackage._get_updated_oci_purl_external_refs(
+            digest,
+            repository,
+            tags,
+            arch=arch,
+        )
+
+        self._strip_oci_purls_external_refs()
+        self.external_refs[:0] = new_oci_refs
+
+    def _strip_oci_purls_external_refs(self) -> None:
+        """
+        Remove all OCI purl externalRefs from a package.
+        """
+
+        def is_oci_purl_ref(ref: dict) -> bool:
+            ptype = ref.get("referenceType")
+            if ptype != "purl":
+                return False
+            purl_str = ref.get("referenceLocator")
+            if purl_str is None:
+                return False
+
+            purl = PackageURL.from_string(purl_str)
+            return purl.type == "oci"
+
+        new_external_refs = [ref for ref in self.external_refs if not is_oci_purl_ref(ref)]
+        self.external_refs = new_external_refs
+
+    @staticmethod
+    def _get_updated_oci_purl_external_refs(
+        digest: str, repository: str, tags: list[str], arch: Optional[str] = None
+    ) -> list[dict]:
+        """
+        Gets new oci purl externalRefs value based on input information.
+        """
+        purls = (construct_purl(repository, digest, tag=tag, arch=arch) for tag in tags)
+        return [SPDXPackage._make_purl_ref(purl) for purl in purls]
+
+    @staticmethod
+    def _make_purl_ref(purl: str) -> dict[str, str]:
+        """
+        Create an SPDX externalRefs field from a PackageURL.
+        """
+        return {
+            "referenceCategory": "PACKAGE-MANAGER",
+            "referenceType": "purl",
+            "referenceLocator": purl,
+        }
+
+
+class SPDXVersion2:  # pylint: disable=too-few-public-methods
+    """
+    Class containing methods for SPDX v2.x SBOM manipulation.
+    """
+
+    supported_versions = [
+        "SPDX-2.0",
+        "SPDX-2.1",
+        "SPDX-2.2",
+        "SPDX-2.2.1",
+        "SPDX-2.2.2",
+        "SPDX-2.3",
+    ]
+
+    @classmethod
+    def _find_purl_in_refs(cls, package: SPDXPackage, digest: str) -> Optional[str]:
+        """
+        Tries to find a purl in the externalRefs of a package the version of
+        which matches the passed digest.
+        """
+        for ref in filter(lambda rf: rf["referenceType"] == "purl", package.external_refs):
+            purl = ref["referenceLocator"]
+            if digest == get_purl_digest(purl):
+                return purl
+
+        return None
+
+    @classmethod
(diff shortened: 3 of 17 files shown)


sbom/handlers/spdx_2_3.py
```diff
@@ -0,0 +1,182 @@
...
+        for package in sbom.get("packages", []):
+            checksums = package.get("checksums")
+            if checksums is None:
+                continue
+
+            package_digest = None
+            for checksum in checksums:
+                if checksum.get("algorithm") == "SHA256":
+                    package_digest = checksum.get("checksumValue")
+                    break
```
wcheang, line 76: This looks pretty much a duplicate of  `_extract_sha256_checksum` below?

sbom/handlers/spdx_2_3.py
```diff
@@ -0,0 +1,182 @@
...
+        for package in sbom["packages"]:
+            if not cls._is_relevant(package, index):
+                continue
+
+            original_purl = cls._find_purl_in_refs(package, index.digest)
+            if original_purl is None:
+                continue
+
+            arch = get_purl_arch(original_purl)
+            package["externalRefs"] = cls._get_updated_external_refs(
```
wcheang, line 149: The reason that the original update script removed all the existing externalRefs with purls, then extended the list with the new externalRefs, instead of doing a straight replacement, was to be careful not to remove any other externalRefs that might contain other info. I think there could also be other externalRefs that are not purls. Please double check to make sure that there wouldn't be any non-purl externalRefs lost in this operation.

sbom/handlers/spdx_2_3.py
jedinym: Updated.

sbom/handlers/spdx_2_3.py
jedinym: Good call, I will preserve other refs that are not OCI purl refs.

jedinym:
@BorekZnovustvoritel please take a look.

sbom/sbomlib.py
```diff
@@ -0,0 +1,378 @@
...
+    registry = ref.split("/", 1)[0]
+
+    with open(auth, mode="r", encoding="utf-8") as f:
+        config = json.load(f)
+    auths = config.get("auths", {})
+
+    current_ref = ref
+
+    try:
+        tmpfile = tempfile.NamedTemporaryFile(mode="w", delete=False)
```
mmalina, line 335: This is basically a reimplementation of the select-oci-auth bash script, right? We added support for docker.io recently. I wonder if you need to worry about that at all. Probably not since this is only to be used for official Red Hat releases that should have both source and target repos in Quay.io?

requirements.txt
mmalina, line 1: I think you will also need to install some/all of these in the Dockerfile, no? Typically, where available it is preferred to install a rpm package with the given lib, the rest is installed via pip.

sbom/README.md
```diff
@@ -0,0 +1,36 @@
...
+Tekton task as path to a file.
+
+## Component-level SBOM enrichment
+When component-level SBOMs are updated, the OCI PURLs generated during
+build-time are stripped and new PURLs are generated based on the parsed snapshot
+data. Non-OCI PURLs are preserved.
+
+### Handling SBOM formats
+To update multiple formats of SBOMs (SPDX-2.X, SPDX-3.0, CycloneDX 1.X, etc), an
+SBOM handler must be implemented. An SBOM handler is a class which implements
```
mmalina, line 32: "must be implemeted"? Shouldn't this be "is implemeted"? Or is that not done yet and will come later? (I haven't checked all the code yet.)

mmalina: Oh, it seems only spdx support is implemented? Do we even need to support anything else? And if not, do we even need the logic to support more formats?

sbom/handlers/spdx2.py
```diff
@@ -0,0 +1,250 @@
...
+        "SPDX-2.2",
+        "SPDX-2.2.1",
+        "SPDX-2.2.2",
+        "SPDX-2.3",
+    ]
+
+    @classmethod
+    def _find_purl_in_refs(cls, package: SPDXPackage, digest: str) -> Optional[str]:
+        """
+        Tries and to find a purl in the externalRefs of a package the version of
```
mmalina, line 145: Typo:
```suggestion
        Tries to find a purl in the externalRefs of a package the version of
```

sbom/update_component_sbom.py
```diff
@@ -3,190 +3,163 @@
...
         required=True,
-        type=str,
-        help="Path to the directory holding the SBOM files to be updated.",
+        type=Path,
+        help="Path to the snapshot spec file in JSON format.",
     )
     parser.add_argument(
         "--output-path",
         required=True,
         type=str,
```
mmalina, line 155: Since you use `type=Path` for the input path, would it make sense here as well?

sbom/update_component_sbom.py
```diff
@@ -3,190 +3,163 @@
 This script updates the purls in component-level SBOMs with release time info.
 """
```
mmalina, line 4: Could we have some usage example(s) here?

sbom/README.md
jedinym: Right now only SPDX 2.X is implemented, we might have to support CycloneDX as well because that's what OCP art is generating at build-time and it is unclear how difficult it would it be for them to migrate to spdx. 

Here's their tracker for the migration: https://issues.redhat.com/browse/ART-12117
Our Jira for CycloneDX implementation: https://issues.redhat.com/browse/ISV-5785

sbom/update_component_sbom.py
jedinym: added

sbom/update_component_sbom.py
jedinym: Fixed.

sbom/sbomlib.py
jedinym: Yes, it is. For now it is used just for official Red Hat releases, but the SBOM functionality - updating the SBOMs and attesting them to the registries (not yet implemented) can be useful for other target repos. I'm not really sure how to handle it right now so we should probably worry about it when we come to it.

sbom/README.md
```diff
@@ -0,0 +1,36 @@
...
+the `SBOMHandler` interface for updating SBOMs.
+
+## Product-level SBOM creation
+TODO
```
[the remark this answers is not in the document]
jedinym: This section will be updated once product-level SBOM generation is reimplemented. I updated the README to make that clear.

sbom/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+The snapshot spec is parsed using the `sbom.sbomlib.make_snapshot` function into
+a Python representation of the snapshot. The `make_snapshot` function also
+handles multiarch images by fetching manifests of index images (using `oras`)
+and parsing them. The resulting object contains all component repositories and
+digests of images that are being released in the snapshot.
+     
+### ReleasePlanAdmission
+The product-level SBOM requires additional data to be constructed (such as the
+cpe id, product name and version). This data can be found in the
```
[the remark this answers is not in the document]
jedinym: Thanks for spotting that, I updated it.

requirements.txt
mmalina: > IMHO, this way of doing dependency management is a little fragile. The deps are not pinned so rebuilds of the image might quietly change the behaviour of the scripts.

Yeah, pinning to versions sounds like a good idea.

> Would you consider using something like [uv](https://docs.astral.sh/uv/) or [pdm](https://pdm-project.org/en/latest/) to manage Python deps in this project? Both of these provide a lockfile for deps and allow different deps for different directories. IMO it would be the more robust solution.

I'm definitely open to suggestions. I think nobody in our team is really up to date on the latest trends when it comes to Python, so if you think there is a better way to do it, I'm open to suggestions. Even better if you volunteer to contribute the changes :) But if not, could you create a RELEASE ticket for us with suggestions?

sbom/README.md
mmalina: OK, thanks. So we're not sure yet if the other implementation will be needed? Do you think it still makes sense to prepare for the possibility? I wonder if it would make sense to just have a direct implementation of spdx and if the need for cyclonedx arises, then reimplement it using the abstract class? My thinking is just that we might not need the extra layer of abstraction at all. If you think it's better to have the abstract class now, it's fine with me.

Either way, I think "will have to be" instead of "must be" would be clearer. Or even "would" since it's not clear yet if it will be needed?

jedinym: I decided to remove this section as it should be pretty obvious from the code itself.

The abstract class was there just to make sure that handlers have the right method signature for the `update_sbom` function. I replaced it with a Protocol to make it more obvious that no shared code is expected between SBOM handlers.

jedinym: I do expect we will have to implement SPDX3.0.

sbom/update_component_sbom.py
```diff
@@ -1,192 +1,175 @@
...
+    Determines format of the SBOM and calls the correct handler or throws
+    SBOMError if the format of the SBOM is unsupported.
+
     Args:
-        sbom: CycloneDX SBOM file to update.
-        component_to_purls_map: dictionary mapping of component names to list of purls.
+        repository (str): Repository of the image
+        image (IndexImage | Image): Object representing an image or an index
+                                    image being released.
+        destination (Path): Path to the directory to save the SBOMs to.
```
BorekZnovustvoritel, line 86: This docstring does not match the function signature.

sbom/README.md
mmalina: OK, sounds good, thanks.

BorekZnovustvoritel (commented):
I have written some rather minor notes, I hope I am not completely out of touch with them :sweat_smile: I really like the async rewrite and the code is quite clean, so other than that it looks good to me :+1:

sbom/sbomlib.py
```diff
@@ -0,0 +1,377 @@
...
+    return purl["qualifiers"].get("arch")
+
+
+def get_purl_digest(purl_str: str) -> str:
+    """
+    Get the image digest from a PackageURL.
+    """
+    purl = PackageURL.from_string(purl_str)
+    if purl.version is None:
+        raise ValueError()
```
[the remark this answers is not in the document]
jedinym: I don't believe in nitpicks! This is an oversight, I will fix it.

sbom/sbomlib.py
```diff
@@ -0,0 +1,377 @@
...
+    rh_registry_repo: str = pdc.Field(alias="rh-registry-repo")
+    tags: list[str]
+
+    @pdc.field_validator("image_digest", mode="after")
+    @classmethod
+    def is_valid_digest_reference(cls, value: str) -> str:
+        """
+        Validates that the digest reference is in the correct format.
+        """
+        if not re.match(r"^[^:]+@sha256:[0-9a-f]+$", value):
```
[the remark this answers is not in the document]
jedinym: Ports are not supported for now. The  `make_oci_auth_file` function would also fail with a port. For now we only support official RedHat releases, where we don't have to worry about it. I will make it more clear in the code.

sbom/logging.py
```diff
@@ -0,0 +1,24 @@
...
+    "disable_existing_loggers": False,
+    "formatters": {"simple": {"format": "%(asctime)s - %(levelname)s - %(message)s"}},
+    "handlers": {
+        "stdout": {
+            "class": "logging.StreamHandler",
+            "formatter": "simple",
+            "stream": "ext://sys.stdout",
+        }
+    },
+    "loggers": {"root": {"level": "DEBUG", "handlers": ["stdout"]}},
```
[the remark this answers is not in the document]
jedinym: I do want logs from other libraries but I decided to restricted to only the WARNING level.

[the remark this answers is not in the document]
BorekZnovustvoritel: Ok, makes sense :+1:

BorekZnovustvoritel (approved):
LGTM from my side. Thanks for addressing my comments so quickly.

sbom/sbomlib.py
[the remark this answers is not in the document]
johnbieren: Resolving since Marek approved

sbom/sbomlib.py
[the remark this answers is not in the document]
johnbieren: Resolving since Marek approved

sbom/update_component_sbom.py
johnbieren: Resolving since Marek approved

jedinym:
@wcheang could you please take a look?

wcheang (approved):

jedinym:
@johnbieren could we get this merged? thanks

johnbieren:
> @johnbieren could we get this merged? thanks

You can't merge a PR with open conversations in this repo, which this has. Also, please ping @konflux-ci/release-service-maintainers instead of me directly for something like asking to merge

requirements.txt
jedinym: I think it would be preferable if your team handled it, since you're the owners of this repo and we will be moving all the SBOM code from here somewhere else soon (https://issues.redhat.com/browse/CLOUDWF-10938) .

[RELEASE ticket](https://issues.redhat.com/browse/RELEASE-1539)

mmalina: Thanks for the ticket!

> we will be moving all the SBOM code from here somewhere else soon

I'm curious about this. Is there a ticket for it? None of the tickets in [CLOUDWF-10938](https://issues.redhat.com/browse/CLOUDWF-10938) seems to be about moving the code (at least looking at the titles).

mmalina (approved):

requirements.txt
jedinym: Sorry, the Epic is not filled with issues yet. You can check the [design doc](https://docs.google.com/document/d/1QMmOvS524ySMyXVXtZTKZh0gZVTvCn3_KHPEy0ZVL58). But the epic will have more stories soon.

mmalina: ok, thanks

mmalina:
@jedinym can you maybe squash the commits and add a few lines of a description in the commit message? I could just squash it here, but it's best if you do it yourself so that you can craft the commit message.

mmalina (approved):

PR #400, opened 2025-03-24
feat: support jinja expressions in data

scoheb:
- in some situations, users might want some control over how
  certain release notes fields are generated based on advisory
  values.
- an example would be a user placing this in their RP/RPA
   
``` 
spec:
   data:
     releaseNotes:
       synopsis: >-
      "{% if advisory.spec.type == 'RHEA' %}Enhancement
       {% endif %}"
```
    
- this change uses a 2 step rendering process. The 1st one renders
  the initial template. The 2nd rendering pass renders any template
  partials from the data model that may have been supplied by the
  RP/RPAs or via a Collector.
- this change also adds proper logging.

### DIFF
--- a/utils/apply_template.py
+++ b/utils/apply_template.py
@@ -1,8 +1,16 @@
 #!/usr/bin/env python3
-from jinja2 import Template, DebugUndefined
+import traceback
+
+import yaml
+from jinja2 import Template, DebugUndefined, exceptions
 from jinja2_ansible_filters import AnsibleCoreFiltersExtension
 import argparse
 import json
+import logging
+import sys
+from typing import Any
+
+LOGGER = logging.getLogger("apply_template")
 
 
 def setup_argparser() -> argparse.Namespace:  # pragma: no cover
@@ -29,6 +37,7 @@ def setup_argparser() -> argparse.Namespace:  # pragma: no cover
         help="The desired filename of the result.",
         required=True,
     )
+    parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
     return parser.parse_args()
 
 
@@ -36,17 +45,79 @@ def main():  # pragma: no cover
     """Main func"""
 
     args = setup_argparser()
+    log_level = logging.DEBUG if args.verbose else logging.INFO
+    setup_logger(level=log_level)
 
     with open(args.template) as t:
         template = Template(
             t.read(), extensions=[AnsibleCoreFiltersExtension], undefined=DebugUndefined
         )
-    content = template.render(json.loads(args.data))
+    LOGGER.info("Rendering 1st pass")
+    try:
+        content = template.render(json.loads(args.data))
+        LOGGER.debug(content)
+        first_pass = content
+    except exceptions.TemplateSyntaxError as jexc:
+        LOGGER.exception("Exception with Template Syntax:")
+        # we use this traceback to get the line number
+        LOGGER.error(traceback.format_exc())
+        raise jexc
+
+    # try 2nd pass
+    LOGGER.info("Rendering 2nd pass")
+    try:
+        content = Template(content).render(json.loads(args.data))
+        LOGGER.debug(content)
+    except exceptions.TemplateSyntaxError as jexc:
+        LOGGER.exception("Exception with Template Syntax:")
+        # we use this traceback to get the line number
+        LOGGER.error(traceback.format_exc())
+        raise jexc
+
+    try:
+        # load to check it is valid yaml
+        LOGGER.info("Load 2nd pass content")
+        yaml.safe_load(content)
+    except yaml.YAMLError:
+        LOGGER.exception("Invalid yaml...fall back to first pass rendered content")
+        # we use this traceback to get the line number
+        LOGGER.error(traceback.format_exc())
+        # fallback to valid first pass
+        content = first_pass
+        try:
+            # load to check it is valid yaml
+            LOGGER.info("Load 1st pass content")
+            yaml.safe_load(content)
+        except yaml.YAMLError as exc:
+            LOGGER.exception("Invalid yaml")
+            # we use this traceback to get the line number
+            LOGGER.error(traceback.format_exc())
+            raise exc
 
     filename = args.output
     with open(filename, mode="w", encoding="utf-8") as advisory:
         advisory.write(content)
-        print(f"Wrote {filename}")
+        LOGGER.info(f"Wrote {filename}")
+
+
+def setup_logger(level: int = logging.INFO, log_format: Any = None):
+    """Set up and configure logger.
+    Args:
+        level (str, optional): Logging level. Defaults to logging.INFO.
+        log_format (Any, optional): Logging message format. Defaults to None.
+    :return: Logger object
+    """
+    if log_format is None:
+        log_format = "%(asctime)s [%(name)s] %(levelname)s %(message)s"
+
+    stream_handler = logging.StreamHandler(sys.stdout)
+    stream_handler.setLevel(level)
+
+    logging.basicConfig(
+        level=level,
+        format=log_format,
+        handlers=[stream_handler],
+    )
 
 
 if __name__ == "__main__":  # pragma: no cover
--- a/utils/test_apply_template.py
+++ b/utils/test_apply_template.py
@@ -1,11 +1,14 @@
 import tempfile
 import json
 import os
+
+from jinja2 import TemplateSyntaxError
 import yaml
 
 import pytest
 
 from unittest.mock import patch, MagicMock
+
 from apply_template import setup_argparser, main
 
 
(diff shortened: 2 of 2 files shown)


utils/apply_template.py
mmalina, line 1: At first glance this seems risky. How will we ensure that after the first render it's a valid jinja template? And that after the second pass it's a valid yaml? I think it would be very easy to make a mistake in the RPA and break it. It might make more sense to implement something like this in the konflux-release-data repo, that way you can validate it in MRs.

Have you considered the risks? I wonder what others think as well.

johnbieren: Well, Scott's example wouldn't work if it was put in konflux-release-data, specifically
`"{%- if advisory.spec.type == 'RHEA' %} Enhancement{%- endif %} synopsis"` because we overwrite type for the user if they provide it incorrectly. As far as first render vs second, agreed that is a bit risky. What about a while loop? While unrendered variables exist, render

scoheb: my real motivation was having a **collector** supply `.spec.data.releaseNotes.synopsis` ... the repo where the collector pull these templates from (see https://issues.redhat.com/browse/CPAAS-5845) will have CI testing to hopefully catch invalid templates.

we could have a check to make sure that after the 2nd render, the resulting file is parseable...if not, do not include that 2nd render.

mmalina: @johnbieren it would work, just with the caveat that your stated type (in RPA) might be overwritten, no? 

> we could have a check to make sure that after the 2nd render, the resulting file is parseable...if not, do not include that 2nd render.

I like that. That would give us more confidence that it's ok.

scoheb: maybe my example is not correct and could be confusing due to what @johnbieren mentioned.

johnbieren: Depends what you mean by work. The task would complete, yeah. And you would have
```
type: RHSA
synopsis: This is my RHEA synopsis
```
I would call that not working

mmalina: > type: RHSA
> synopsis: This is my RHEA synopsis

Oh I see what you mean. But that won't happen. The template rendering is the last thing that happens. So when `{%- if advisory.spec.type == 'RHEA' %}` is rendered, the `advisory.spec.type` will definitely be RHSA - especially now that you moved the type to the populate task, but even before - it happened before the apply_template python script was called - that's the very last thing.

johnbieren: I was saying 

>  It might make more sense to implement something like this in the konflux-release-data repo

would not work for this reason

mmalina: But this brings up other potential issues - what if we rely on some normal string values in some of the fields and now it can contain a jinja template instead. Like, for example if for the type they defined `"{% if advisory.spec.type.startsWith(\"Security\") %} RHSA {% else %} RHEA {% endif %}"` or something like that. 

And then if we see CVEs we override that anyway. Which would be ok in this example I guess. But maybe there is some other use case where it would cause issues?

johnbieren: The only thing I was trying to get at is if you want to do things based on the value of other advisory fields, it has to happen in apply_template, not somewhere else, because the fields may be changed during the managed pipelineRun. I'm not really sure what you are saying with your example because type can't be anything but RHSA/RHEA/RHBA so I don't know how `spec.type.startsWith(\"Security\")` could ever be true

mmalina: @johnbieren sorry, should have read your answer more carefully.

> Or after the first render do the check as part of a loop? I don't see why we would start checking after render 2 instead of 1

Doing a loop sounds even worse to me. I think 2 renders should be enough. After the first render, it should be guaranteed by us that it's a valid yaml (right now we actually don't handle it properly, but we have a jira for that). But now I'm thinking if it's a real concern that after the second render it will be invalid yaml. I think it's more likely that it will be invalid template - if people use `"` in a template snippet, we will escape it so that it's valid yaml. But then that would break the second render.

mmalina: > so I don't know how spec.type.startsWith(\"Security\") could ever be true

I meant to use spec.synopsis.startsWith(\"Security\") as an example.

johnbieren: I don't understand how doing it 2 times is less risky than 3 times. If we do it 2x, we are saying, "okay, you can have your synopsis be a jinja2 template based on type" (for example). But we are only doing it twice. So, you can do that, but you can't then have description based on synopsis. That is very strange to me. If we are going to support recursive fields, support them no matter the recursion level, or don't at all IMO

mmalina: > If we are going to support recursive fields, support them no matter the recursion level, or don't at all IMO

I agree with this. My preference would be no recursion, i.e. two renders. Two renders means that you are allowed to use jinja template in a field and refer to another field, but you are not allowed to have that other field also use jinja template, because that would require recursion - we would only allow one level of references. That would be my preference.

But I'm fine with recursion as well.

johnbieren: Isn't two renders one level of recursion?

scoheb: @mmalina 

> so if we agree to go with this, maybe you could fix https://issues.redhat.com/browse/KONFLUX-7227 as a bonus? ;)

not sure I can do this in this PR

utils/apply_template.py
```diff
@@ -29,24 +37,87 @@ def setup_argparser() -> argparse.Namespace:  # pragma: no cover
...
         template = Template(
             t.read(), extensions=[AnsibleCoreFiltersExtension], undefined=DebugUndefined
         )
-    content = template.render(json.loads(args.data))
+    LOGGER.info("Rendering 1st pass")
+    try:
+        content = template.render(json.loads(args.data))
+        LOGGER.debug(content)
+        first_pass = content
+    except exceptions.TemplateSyntaxError as jexc:
```
mmalina, line 60: I'm curious. At this point it's simply the template we have in git, right? So you don't have to worry about syntax errors in the template, no? That would be a big problem.

utils/apply_template.py
```diff
@@ -29,24 +37,87 @@ def setup_argparser() -> argparse.Namespace:  # pragma: no cover
...
+        raise jexc
+
+    # try 2nd pass
+    LOGGER.info("Rendering 2nd pass")
+    try:
+        content = Template(content).render(json.loads(args.data))
+        LOGGER.debug(content)
+    except exceptions.TemplateSyntaxError as jexc:
+        LOGGER.exception("Exception with Template Syntax:")
+        # we use this traceback to get the line number
```
mmalina, line 73: But does that help you? You don't get to see the full content after the first round, right?

utils/apply_template.py
scoheb: You would hope not. I was trying to be consistent with error handling for rendering

mmalina: OK, fine with me.

utils/apply_template.py
mmalina: OK, if you think it's still useful, I'm fine with keeping it.

mmalina (dismissed):

utils/apply_template.py
```diff
@@ -29,24 +37,87 @@ def setup_argparser() -> argparse.Namespace:  # pragma: no cover
...
 
     filename = args.output
     with open(filename, mode="w", encoding="utf-8") as advisory:
         advisory.write(content)
-        print(f"Wrote {filename}")
+        LOGGER.info(f"Wrote {filename}")
+
+
+def setup_logger(level: int = logging.INFO, log_format: Any = None):
+    """Set up and configure 'pyxis' logger.
```
johnbieren, line 104: This isn't pyxis related

utils/test_apply_template.py
```diff
@@ -66,6 +70,8 @@ def test_apply_template_advisory_template_in_full(mock_argparser: MagicMock):
         "on-and-on-and-on-and-on-and-on,-it's-so-long-that-you-would-"
         "expect-that-something-is-going-to-linewrap-it-at-some-point.-so-long."
     )
+    # Confirm collector partials are rendered
```
johnbieren, line 73: This isn't collector related IMO

utils/test_apply_template.py
```diff
@@ -103,5 +109,50 @@ def test_apply_template_advisory_template_in_full(mock_argparser: MagicMock):
...
+        os.remove(filename)
+
+
+@patch("apply_template.setup_argparser")
+def test_apply_template_advisory_template_fail_syntax_error(mock_argparser: MagicMock):
+    _, filename = tempfile.mkstemp()
+
+    # error in this partial template
+    synopsis = "{% if advisory.spec.type == ' %}FAILURE{%- endif %}"
+    #                    error is here     ^^^
```
johnbieren, line 123: this is great

utils/test_apply_template.py
scoheb: corrected the text

mmalina (approved):

PR #404, opened 2025-03-25
fix(RELEASE-1491): cleanup_tags supports multiple repos

theflockers:
this PR fixes the case when the image has multiple repositories and ends up not cleaning tags of all
of them.

### DIFF
--- a/pyxis/cleanup_tags.py
+++ b/pyxis/cleanup_tags.py
@@ -40,6 +40,7 @@ def parse_arguments() -> argparse.Namespace:  # pragma: no cover
         ),
         help="Pyxis Graphql endpoint",
     )
+    parser.add_argument("--repository", required=True, help="Repository to cleanup tags from")
     parser.add_argument(
         "image_id",
         help="Pyxis Container Image ID",
@@ -58,6 +59,7 @@ def parse_arguments() -> argparse.Namespace:  # pragma: no cover
 def cleanup_tags_with_retry(
     graphql_api: str,
     image_id: str,
+    target_repository: str,
     retries: int = 3,
     backoff_factor: float = 5.0,
 ):
@@ -65,7 +67,7 @@ def cleanup_tags_with_retry(
     for attempt in range(retries):
         try:
             time.sleep(backoff_factor * attempt)
-            cleanup_tags(graphql_api, image_id)
+            cleanup_tags(graphql_api, image_id, target_repository)
             return
         except RuntimeError as e:
             LOGGER.warning(f"Attempt {attempt+1} failed.")
@@ -74,13 +76,14 @@ def cleanup_tags_with_retry(
     raise last_err
 
 
-def cleanup_tags(graphql_api, image_id: str):
+def cleanup_tags(graphql_api, image_id: str, target_repository: str):
     image = get_image(graphql_api, image_id)
 
-    registry, repository, tags = get_rh_registry_image_properties(image)
+    registry, repository, tags = get_rh_registry_image_properties(image, target_repository)
 
     LOGGER.info(f"Image id: {image['_id']}")
     LOGGER.info(f"Image architecture: {image['architecture']}")
+    LOGGER.info(f"Repository: {repository}")
     LOGGER.info(f"Image tags: {tags}")
 
     images_for_cleanup = {}
@@ -102,7 +105,7 @@ def cleanup_tags(graphql_api, image_id: str):
                 images_for_cleanup[id] = candidate
 
     LOGGER.info(f"Found {len(images_for_cleanup)} images for cleanup.")
-    update_images(graphql_api, tags, images_for_cleanup)
+    update_images(graphql_api, tags, images_for_cleanup, repository)
 
 
 def get_image(graphql_api: str, image_id: str) -> dict:
@@ -137,19 +140,23 @@ query ($id: ObjectIDFilterScalar!) {
     return image
 
 
-def get_rh_registry_image_properties(image: Dict):
+def get_rh_registry_image_properties(image: Dict, target_repository: str):
     """Get the registry.access.redhat.com repository properties of the image
     needed to search for related images.
 
-    Returns (registry, repository, tags)
+    :return: (registry, repository, tags)
     """
     for repo in image["repositories"]:
-        if repo["registry"] == "registry.access.redhat.com":
+        if (
+            repo["registry"] == "registry.access.redhat.com"
+            and repo["repository"] == target_repository
+        ):
             if repo["tags"] is None:
                 tags = []
             else:
                 tags = [tag["name"] for tag in repo["tags"]]
             return repo["registry"], repo["repository"], tags
+
     raise RuntimeError(
         "Cannot find the registry.access.redhat.com repository entry for the image"
     )
@@ -246,7 +253,7 @@ query (
     return images
 
 
-def update_images(graphql_api: str, tags: List[str], images: Dict):
+def update_images(graphql_api: str, tags: List[str], images: Dict, target_repository: str):
     """Update images to remove unwanted tags from them
 
     For each image in `images` it will remove all `tags`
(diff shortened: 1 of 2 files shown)


johnbieren (commented):
lgtm but I don't think we should merge until Martin gets a chance to review

pyxis/cleanup_tags.py
```diff
@@ -76,33 +76,33 @@ def cleanup_tags_with_retry(
 
 def cleanup_tags(graphql_api, image_id: str):
     image = get_image(graphql_api, image_id)
+    for properties in get_rh_registry_image_properties(image):
+        registry, repository, tags = properties
 
-    registry, repository, tags = get_rh_registry_image_properties(image)
+        LOGGER.info(f"Image id: {image['_id']}")
+        LOGGER.info(f"Image architecture: {image['architecture']}")
+        LOGGER.info(f"Image tags: {tags}")
```
mmalina, line 84: I think it would make sense to print image id and architecture just once (before the `for` loop starts) and then inside of it print repository and tags.

pyxis/cleanup_tags.py
```diff
@@ -76,33 +76,33 @@ def cleanup_tags_with_retry(
...
+                    id != image["_id"]
+                    and id not in images_for_cleanup
+                    and candidate["architecture"] == image["architecture"]
+                ):
+                    images_for_cleanup[id] = candidate
 
-    LOGGER.info(f"Found {len(images_for_cleanup)} images for cleanup.")
-    update_images(graphql_api, tags, images_for_cleanup)
+        LOGGER.info(f"Found {len(images_for_cleanup)} images for cleanup.")
+        update_images(graphql_api, tags, images_for_cleanup)
```
mmalina, line 105: We are doing this once for each repository now. Currently the `update_images` function will just remove the tags from all repositories of that image. It needs to be modified to only remove them from the given repository (so a new `repository` parameter will be needed).

pyxis/cleanup_tags.py
```diff
@@ -141,18 +141,24 @@ def get_rh_registry_image_properties(image: Dict):
...
             if repo["tags"] is None:
                 tags = []
             else:
                 tags = [tag["name"] for tag in repo["tags"]]
-            return repo["registry"], repo["repository"], tags
-    raise RuntimeError(
-        "Cannot find the registry.access.redhat.com repository entry for the image"
-    )
+            image_properties = (repo["registry"], repo["repository"], tags)
+            properties_list.append(image_properties)
```
mmalina, line 154: We should just update the repository that is being released. I think that would be more correct. Imagine the current scenario of rhel9 images - they have two release pipelines, releasing the same images. Once to ubi9/ and once to rhel9/. So for each release, we should only be cleaning up the tags in the given repository, not both of them. (The release pipeline that runs first will only have one and the second one will have both.)

If we wanted to do that, we would need to supply a new repository parameter when calling the script in the create-pyxis-image task.

Maybe in practice this is good enough, but it doesn't sound too complicated to add that - you can just use the `$REPOSITORY` variable that's already used for the `create_container_image` call.

theflockers: Indeed. So it would require adding the `--repository` to `cleanup_tags `  and also change the task to use the new parameter.

pyxis/cleanup_tags.py
mmalina: > the images_for_cleanup variable contains the information to remove the tags, including the repository.

It contains the full images, including the repositories with tags.

> What you mean here is that the update_images should also have the "repository" as a parameters, so when update_images is executed, it will only update the image (remove tags) of the one that matches this given repository, correct?

Yes, it should only update tags for the given repository of each image in `images_for_cleanup`.

theflockers: I did it a bit different. I changed yet in `get_rh_registry_image_properties` but now with the repository parameter, so it would only return the properties for the repository we want to delete. This way I would not need to add the parameter in all chain until the update_image. WDYT?

pyxis/cleanup_tags.py
theflockers: parameter added. I will create a PR to release-service-catalog to update the task.

pyxis/cleanup_tags.py
theflockers: log lines moved.

pyxis/cleanup_tags.py
mmalina: If I understand correctly what you mean - that the `images_for_cleanup` will only have the one repository inside - then unfortunately you cannot do that. Because the image in Pyxis is then patched with the image from `images_for_cleanup`. And if the image previously had 2 (or more) repositories, you would delete them and only end up with the one.

pyxis/cleanup_tags.py
```diff
@@ -74,35 +76,38 @@ def cleanup_tags_with_retry(
     raise last_err
 
 
-def cleanup_tags(graphql_api, image_id: str):
+def cleanup_tags(graphql_api, image_id: str, target_repository: str = None):
```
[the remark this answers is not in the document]
theflockers: Yeah. I was wondering if for any reason if this is imported by someone else this would avoid breaking. Not the case I guess, so I will  remove the default.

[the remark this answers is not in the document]
theflockers: default removed.

pyxis/cleanup_tags.py
```diff
@@ -40,6 +40,7 @@ def parse_arguments() -> argparse.Namespace:  # pragma: no cover
         ),
         help="Pyxis Graphql endpoint",
     )
+    parser.add_argument("--repository", help="Repository to cleanup tags from")
```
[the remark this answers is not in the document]
theflockers: added.

pyxis/cleanup_tags.py
theflockers: Logic moved to update_images.

pyxis/cleanup_tags.py
```diff
@@ -58,6 +59,7 @@ def parse_arguments() -> argparse.Namespace:  # pragma: no cover
 def cleanup_tags_with_retry(
     graphql_api: str,
     image_id: str,
+    target_repository: str = None,
```
mmalina, line 62: This parameter needs to be propagated to the `cleanup_tags` call in this function. The `cleanup_tags_with_retry` should also be modified to check this.

pyxis/cleanup_tags.py
```diff
@@ -141,18 +146,24 @@ def get_rh_registry_image_properties(image: Dict):
     """Get the registry.access.redhat.com repository properties of the image
     needed to search for related images.
 
-    Returns (registry, repository, tags)
+    :return: List ([registry, repository, tags])
     """
+    properties_list = []
```
mmalina, line 151: Based on our discussion, this should be changed back to just a single tuple, right? But we are only interested in the one entry - the repository we are looking for. So this function will need a new `repository` parameter and then instead of `if repo["registry"] == "registry.access.redhat.com":` you will have `if repo["registry"] == "registry.access.redhat.com" and repo["repository"] == repository:`. Otherwise the function can be exactly as it was originally - return early, or else raise the error.

pyxis/cleanup_tags.py
mmalina: I added a new comment. The logic in update_images looks good.

pyxis/cleanup_tags.py
theflockers: changed back to return a single item.

pyxis/cleanup_tags.py
theflockers: fixed.

pyxis/cleanup_tags.py
mmalina: Looks good.

mmalina (dismissed):
It looks good, but it would be a good idea to try running the script on something. I can do some testing, but maybe not today. Probably best to do it with stage pyxis - pick the latest image in some repo and run it on it - it should just say that there is nothing to cleanup. But then you could add one of the floating tags to an older image and repeated run should clean that.

pyxis/test_cleanup_tags.py
```diff
@@ -152,13 +154,25 @@ def test_get_rh_registry_image_properties__no_tags():
...
+    registry, repository, tags = get_rh_registry_image_properties(image, REPOSITORY)
 
     assert registry == REGISTRY
     assert repository == REPOSITORY
     assert tags == []
 
 
+def test_get_rh_registry_image_properties__multiple_images__repository_set__success():
+    """Basic scenario where the function parses the image and returns
+    multiple images
```
mmalina, line 166: It will not return multiple images, right?

theflockers: Yeah, not anymore. I will change it.

theflockers:
> It looks good, but it would be a good idea to try running the script on something. I can do some testing, but maybe not today. Probably best to do it with stage pyxis - pick the latest image in some repo and run it on it - it should just say that there is nothing to cleanup. But then you could add one of the floating tags to an older image and repeated run should clean that.

Sure. I will built it and try to run e2e with it and see how it goes.

theflockers:
@mmalina I got this NoneType error and add the same fix as you did. It seems from the error output that it found the tags, but there was some `None`  values in the Hash it found. Please take a look again just to make sure the fix is proper.

https://gist.github.com/theflockers/9570014e1fab391c3da7a39e6b5ed354

theflockers:
Cool @mmalina . I am waiting to have a successful run of e2e using an image with this change to move forward.

mmalina (approved):

theflockers:
got a good one: https://console.redhat.com/application-pipeline/workspaces/rhtap-release-2/applications/release-service-catalog/pipelineruns/konflux-e2e-tests-catalog-p6xcc merging.

PR #407, opened 2025-03-27
feat(RELEASE-1461): add curl-with-retry to handle Jira rate limiting

mmalina:
This new script will retry in case of code 429 - api rate limiting. It can be used as a drop in replacement for curl with a few caveats:

* -o/--output should not be specified - the response will always be printed to stdout
* --fail should not be specified. The script will always fail if curl ends up failing.

Is needed by https://github.com/konflux-ci/release-service-catalog/pull/866

### DIFF
--- a/utils/curl-with-retry
+++ b/utils/curl-with-retry
@@ -1,0 +1,50 @@
+#!/bin/bash
+#
+# script:      curl-with-retry
+#
+# description: This script handles Jira API rate limiting by retrying requests with exponential backoff.
+#              It will retry up to 5 times, and will wait for a random amount of time between retries.
+#              There are a few caveats:
+#              - The script will only retry if the response code is 429 (rate limited).
+#              - You can still use --retry as needed to retry 5xx errors.
+#              - -o/--output should not be specified - the response will always be printed to stdout.
+#              - --fail should not be specified. The script will always fail if the request fails.
+#
+# example command:
+#              curl-with-retry
+#
+
+set -eu
+
+#!/bin/bash
+
+MAX_RETRIES=5
+BASE_SLEEP_TIME=1  # Initial wait time in seconds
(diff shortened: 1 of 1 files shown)


utils/curl-with-retry
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
+#
+# script:      curl-with-retry
+#
+# description: This script handles Jira API rate limiting by retrying requests with exponential backoff.
+#              It will retry up to 5 times, and will wait for a random amount of time between retries.
+#              There are a few caveats:
+#              - The script will only retry if the response code is 429 (rate limited).
+#              - You can still use --retry as needed to retry 5xx errors.
+#              - -o/--output should not be specified - the response will always be printed to stdout.
```
johnbieren, line 10: Can this be an option? The embargo-check task has to curl jira and currently has `-o /dev/null` when it is checking the fields. We are talking about *a lot* of output, and we don't really need it which is why we currently redirect it to /dev/null. I see this task benefiting from this script (probably even more - it will need it) due to the jira api changes, but printing everything to stdout will make the taskRun log unusable

mmalina: Making it optional would make the script more complex - I could no longer just take all the arguments and use them for curl. It's doable, but I wonder if it's needed - if you check my linked draft PR for catalog, I just added `> /dev/null` for that call, so it shouldn't be an issue: https://github.com/konflux-ci/release-service-catalog/pull/866/files#diff-bd2ced5357d04ee9e6fd86e7396643d88ed01cd5cee8a5258b74922857169f2cR112

johnbieren: Okay, as long as we aren't forced to have it in the taskRun log, I am all good

seanconroy2021 (approved):

theflockers (approved):

PR #413, opened 2025-04-09
upload_rpm_data: propagate epoch and module information

hdonnay:
If the incoming PURLs have this information, make sure to forward it. As of this writing, the Pyxis `RpmsItems` schema doesn't contain these fields.

### DIFF
--- a/pyxis/upload_rpm_data.py
+++ b/pyxis/upload_rpm_data.py
@@ -254,6 +254,10 @@ def construct_rpm_items_and_content_sets(
                 "summary": purl_dict["name"],
                 "architecture": purl_dict["qualifiers"].get("arch", "noarch"),
             }
+            if purl_dict["epoch"] is not None:
+                rpm_item["epoch"] = purl_dict["epoch"]
+            if purl_dict["module"] is not None:
+                rpm_item["module"] = purl_dict["module"]
             if purl_dict["version"] is not None:
                 rpm_item["version"] = purl_dict["version"].split("-")[0]
                 rpm_item["release"] = purl_dict["version"].split("-")[1]


johnbieren (approved):

happybhati (approved):

lslebodn (commented):
Couple of nitpicks.

There is missing code coverage.

And more important pyxis does nt support such fields ATM

-1

mmalina:
> And more important pyxis does nt support such fields ATM

Yep, I agree this is definitely not ready to be merged - at least until Pyxis supports this. I think you created a Jira for Pyxis to add support for this? Can you link it?

> There is missing code coverage.

Yep, we have all kinds of purls in the test, so this should be added.

mmalina:
@johnbieren @happybhati I dismissed your approvals to make sure this isn't accidentally merged.

johnbieren:
> @johnbieren @happybhati I dismissed your approvals to make sure this isn't accidentally merged.

thank you! I didn't realize pyxis would care if we added extra fields

hdonnay:
The Pyxis change to support these fields is being merged currently.

mmalina:
> The Pyxis change to support these fields is being merged currently.

Where? Can you link the Jira or at least the PR? Jira would be better, because it will take some time before the change makes it to production.

johnbieren:
Ping - no activity on this PR in a month and a half

mmalina:
> > > The Pyxis change to support these fields is being merged currently.
> > 
> > 
> > Where? Can you link the Jira or at least the PR? Jira would be better, because it will take some time before the change makes it to production.
> 
> https://gitlab.cee.redhat.com/metadata/pyxis/pyxis/-/merge_requests/2074

OK, thanks.

@kosciCZ can you confirm that this should be safe for us to start pushing to Pyxis (both prod and stage)? I see the new fields here: https://pyxis.engineering.redhat.com/docs/objects/RpmsItems.html?tab=Fields But just to be on the safe side...

mmalina:
No idea why it still says `Queued — Pending approval, waiting for an /ok-to-test`

Trying again.

/ok-to-test

mmalina:
Note that we'll also need a PR in the release-service-catalog repo to use the new image in the task. I can create one later.

mmalina:
PR for catalog: https://github.com/konflux-ci/release-service-catalog/pull/1109

PR #415, opened 2025-04-17
[ISV-5784] Rewrite product-level SBOM generation

jedinym:
The product-level SBOM generation is refactored to use the mapped snapshot spec and the merged data file as a source of truth. This makes the implementation simpler to understand and maintain.

The spdx-tools library is used to generate the SBOM and validate it, to assure its quality.
Tests are written to check that the expected properties on the SBOM hold.

This PR depends on the changes in #399.


sbom/create_product_sbom.py
```diff
@@ -1,159 +1,215 @@
...
+            release_notes = ReleaseData.model_validate_json(raw_json).release_notes
+
+            sbom = create_sbom(release_notes, snapshot)
 
-    sbom = create_sbom(args.data_path)
-    with open(args.output_path, "w") as fp:
-        json.dump(sbom, fp)
+        write_file(document=sbom, file_name=str(args.output_path), validate=True)
+    except Exception:  # pylint: disable=broad-except
+        logger.exception("Creation of the product-level SBOM failed.")
```
mmalina, line 212: That seems very broad. Is that a good idea? If we want to allow the script to fail, wouldn't it be better to handle it in the bash script calling it? Or if we go this way, shouldn't we at least print the exception for debugging purposes? With the current code you won't even know what happened.

mmalina (commented):
I added a few comments, but I didn't go deep into the functionality. Hopefully someone else from your team can review this as well.

sbom/create_product_sbom.py
```diff
@@ -1,159 +1,215 @@
...
-def main():
+def main() -> None:
+    """
+    Script entrypoint.
+    """
     parser = argparse.ArgumentParser(
-        prog="create-product-sbom", description="Create product-level SBOM from releaseNotes."
+        prog="create-product-sbom",
+        description="Create product-level SBOM from merged data file"
+        "and mapped snapshot spec.",
```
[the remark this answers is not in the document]
jedinym: Good eye, I will fix this.

sbom/create_product_sbom.py
jedinym: The [exception()](https://docs.python.org/3/library/logging.html#logging.Logger.exception) method captures the raised exception message and logs it automatically, so we will always see the traceback.

jedinym: Although, there is a missing `raise` statement here :+1:

mmalina: Ah, ok. I assumed you intentionally wanted to swallow the exception.

> The [exception()](https://docs.python.org/3/library/logging.html#logging.Logger.exception) method captures the raised exception message and logs it automatically, so we will always see the traceback.

Good point.

jedinym:
@Allda please take look :squid:

sbom/create_product_sbom.py
```diff
@@ -1,159 +1,215 @@
...
 
+from sbom import sbomlib
+from sbom.logging import get_sbom_logger, setup_sbom_logger
+from sbom.sbomlib import Component, Snapshot, construct_purl
 
-def create_product_package(name: str, version: str, cpe: str) -> Dict:
+logger = get_sbom_logger()
+
+
+class ReleaseNotes(pdc.BaseModel):
```
johnbieren, line 38: It worries me recreating this in yet another spot. cpe is going to change https://issues.redhat.com/browse/KONFLUX-6854 which will break this, no?

jedinym: It will break but if it is just changed into a list, we can fix it easily. 

Will it be enough to adapt this code to support both use cases or do you have another solution in mind?

johnbieren: No, I don't have a better idea. I am just worried about this being forgotten when the change is made

sbom/create_product_sbom.py
jedinym: Fixed.

mmalina (approved):

PR #422, opened 2025-04-29
feat: determine name for product SBOM in-script

jedinym:
By naming the file according to its contents in this script, we can simplify the Tekton task that uses it.

With the move to Atlas V2, the name is no longer used as an ID, so we can use a more human-readable name.

### DIFF
--- a/sbom/create_product_sbom.py
+++ b/sbom/create_product_sbom.py
@@ -180,7 +180,8 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
 
 
 def get_filename(release_notes: ReleaseNotes) -> str:
-    return f"{release_notes.product_name} {release_notes.product_version}.json"
+    normalized_name = "-".join(release_notes.product_name.split())
+    return f"{normalized_name}-{release_notes.product_version}.json"
 
 
 def main() -> None:  # pragma: nocover
@@ -190,8 +191,8 @@ def main() -> None:  # pragma: nocover
     parser = argparse.ArgumentParser(
         prog="create-product-sbom",
         description="Create product-level SBOM from merged data file"
-        " and mapped snapshot spec. Names the SBOM file according to the SBOM"
-        " contents and prints the name to stdout.",
+        " and mapped snapshot spec. The SBOM file is named according to the SBOM"
+        " contents and is printed to stdout.",
     )
     parser.add_argument(
         "--data-path",
@@ -223,9 +224,9 @@ def main() -> None:  # pragma: nocover
             sbom = create_sbom(release_notes, snapshot)
 
         fname = get_filename(release_notes)
-        output = str(args.output_path.joinpath(fname))
-        write_file(document=sbom, file_name=output, validate=True)
-        print(output)
+        output_path = str(args.output_path.joinpath(fname))
+        write_file(document=sbom, file_name=output_path, validate=True)
+        print(f"SBOM file: {output_path}")
     except Exception:  # pylint: disable=broad-except
         logger.exception("Creation of the product-level SBOM failed.")
         raise
--- a/sbom/create_product_sbom.py
+++ b/sbom/create_product_sbom.py
@@ -179,6 +179,10 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
     return ReleaseData.model_validate_json(raw_json).release_notes
 
 
+def get_filename(release_notes: ReleaseNotes) -> str:
+    return f"{release_notes.product_name} {release_notes.product_version}.json"
+
+
 def main() -> None:  # pragma: nocover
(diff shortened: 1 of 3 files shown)


sbom/create_product_sbom.py
```diff
@@ -179,14 +179,19 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
     return ReleaseData.model_validate_json(raw_json).release_notes
 
 
+def get_filename(release_notes: ReleaseNotes) -> str:
+    return f"{release_notes.product_name} {release_notes.product_version}.json"
```
johnbieren, line 183: you want a space in the filename?

jedinym: On second thought, it's not worth the trouble :smile:

sbom/create_product_sbom.py
```diff
@@ -179,14 +179,20 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
     return ReleaseData.model_validate_json(raw_json).release_notes
 
 
+def get_filename(release_notes: ReleaseNotes) -> str:
+    normalized_name = str.join("-", release_notes.product_name.split())
```
mmalina, line 183: Nit: `normalized_name = "-".join(release_notes.product_name.split())` seems more idiomatic.

sbom/create_product_sbom.py
```diff
@@ -179,14 +179,20 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
...
+
 def main() -> None:  # pragma: nocover
     """
     Script entrypoint.
     """
     parser = argparse.ArgumentParser(
         prog="create-product-sbom",
         description="Create product-level SBOM from merged data file"
-        " and mapped snapshot spec.",
+        " and mapped snapshot spec. Names the SBOM file according to the SBOM"
```
mmalina, line 194: Nit: Use passive form: The SBOM file is named ... and the name is printed ...

sbom/create_product_sbom.py
```diff
@@ -179,14 +179,20 @@ def parse_release_notes(raw_json: str) -> ReleaseNotes:
     return ReleaseData.model_validate_json(raw_json).release_notes
 
 
+def get_filename(release_notes: ReleaseNotes) -> str:
```
mmalina, line 182: Can we have a unit test for this?

sbom/create_product_sbom.py
jedinym: I guess it is, but it makes me sad.

sbom/create_product_sbom.py
jedinym: Added.

sbom/logging.py
```diff
@@ -6,14 +6,14 @@
     "disable_existing_loggers": False,
     "formatters": {"simple": {"format": "%(asctime)s - %(levelname)s - %(message)s"}},
     "handlers": {
-        "stdout": {
+        "stderr": {
```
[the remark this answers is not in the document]
jedinym: I noticed it because of that, but logging to stderr is the standard which we should use.

The output is there just for sanity, we won't actually be using it, since upload-sbom-to-atlas uploads a directory.

sbom/create_product_sbom.py
```diff
@@ -215,10 +221,12 @@ def main() -> None:  # pragma: nocover
         with open(args.data_path, "r", encoding="utf-8") as fp:
             raw_json = fp.read()
             release_notes = parse_release_notes(raw_json)
-
             sbom = create_sbom(release_notes, snapshot)
 
-        write_file(document=sbom, file_name=str(args.output_path), validate=True)
+        fname = get_filename(release_notes)
+        output = str(args.output_path.joinpath(fname))
```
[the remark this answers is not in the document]
jedinym: Changed.

sbom/create_product_sbom.py
jedinym: Changed.

sbom/create_product_sbom.py
mmalina: :D Up to you.

sbom/logging.py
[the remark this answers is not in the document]
mmalina: Ah, so you won't be using it like `FILE=$(create_product_sbom ...)`? So if you're not worried about the exact format of the output, would it make sense to print a little more, e.g. "SBOM file: $file"? I agree about sending logging to stderr either way.

johnbieren (dismissed):

jedinym:
I'll be going on PTO until 9.5. We can merge then, it's blocking only me.

mmalina (approved):

jedinym:
Could we get this merged?

PR #444, opened 2025-06-03
fix(ISV-5990): fix SBOM update failing silently

jedinym:
The update_component_sboms script was silently failing on a cosign authorization issue. This PR fixes the cause of the silent failures and the auth issue itself.

Cosign was failing because the authentication file was not named "config.json" as expected.

### DIFF
--- a/sbom/sbomlib.py
+++ b/sbom/sbomlib.py
@@ -326,7 +326,7 @@ def make_reference(repository: str, image_digest: str) -> str:
 
 @contextmanager
 def make_oci_auth_file(
-    reference: str, auth: Optional[Path] = None
+    reference: str, authfile: Optional[Path] = None
 ) -> Generator[str, Any, None]:
     """
     Gets path to a temporary file containing the docker config JSON for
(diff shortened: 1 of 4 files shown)


sbom/sbomlib.py
```diff
@@ -355,29 +364,39 @@ def make_oci_auth_file(
     with open(auth, mode="r", encoding="utf-8") as f:
         config = json.load(f)
     auths = config.get("auths", {})
```
mmalina, line 366: I think it's a bit confusing that `auth` is the filepath and `auths` is the auths content inside that file. Maybe rename `auth` to `auth_file` or `auth_path`?

sbom/update_component_sbom.py
```diff
@@ -112,6 +113,7 @@ async def update_sbom(
         destination (Path): Path to the directory to save the SBOMs to.
     """
 
+    reference = None
     try:
         reference = f"{component.repository}@{image.digest}"
```
mmalina, line 118: Do we expect a ValueError here? If not, this can probably be moved before the `try`, no?

sbom/sbomlib.py
mmalina: Fixed.

sbom/update_component_sbom.py
mmalina: Fixed.

PR #454, opened 2025-06-16
fix(RELEASE-1693): add tags even if empty

mmalina:
We got a report from the Pyxis team that we should not remove the tags field completely when we want
to remove all tags. Instead, we should use an empty list.

In fact, they are planning to stop allowing this, so once fixed on their side, it would break us.


kosciCZ:
Approved from Pyxis POV

johnbieren (approved):

PR #459, opened 2025-06-20
feat(RELEASE-1664): use --retry-times for skopeo inspect

theflockers:
this PR adds the --skopeo-retries to the script
get-image-architectures to enable the skopeo
option --retry-times to use it.

### DIFF
--- a/utils/get-image-architectures
+++ b/utils/get-image-architectures
@@ -10,14 +10,37 @@
 #
 set -e
 
-IMAGE=$1
+show_help(){
+    echo "Usage: $1 [--skopeo-retries <#retries>] [-h|--help] <image>"
+    exit 1
+}
+
+if [ ${#@} -eq 0 ]; then
+    show_help
+fi
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        "--skopeo-retries")
+            SKOPEO_RETRIES=(--retry-times "$2")
+            shift 2
+            ;;
+        -h|--help)
+            show_help
+            ;;
+        *)
+            IMAGE=$1
+            shift
+            ;;
+    esac
+done
 
 if [ -z "${IMAGE}" ] || [[ "${IMAGE}" == docker://* ]] ; then
     echo "Error: Please pass an image (without the docker:// prefix) to find the architectures for" >&2
     exit 1
 fi
 
-RAW_OUTPUT=$(skopeo inspect --no-tags --raw docker://${IMAGE})
+RAW_OUTPUT=$(skopeo inspect "${SKOPEO_RETRIES[@]}" --no-tags --raw docker://${IMAGE})
 
 ARTIFACT_TYPE=$(jq -r '.artifactType' <<< $RAW_OUTPUT)
 if [ "$ARTIFACT_TYPE" != "null" ] ; then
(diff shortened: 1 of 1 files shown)


utils/get-image-architectures
```diff
@@ -10,14 +10,35 @@
...
+    esac
+done
 
 if [ -z "${IMAGE}" ] || [[ "${IMAGE}" == docker://* ]] ; then
     echo "Error: Please pass an image (without the docker:// prefix) to find the architectures for" >&2
     exit 1
 fi
 
-RAW_OUTPUT=$(skopeo inspect --no-tags --raw docker://${IMAGE})
+RAW_OUTPUT=$(skopeo ${SKOPEO_RETRIES} inspect --no-tags --raw docker://${IMAGE})
```
FilipNikolovski, line 41: I think --retry-times should be added after the inspect command, otherwise it fails.
```suggestion
RAW_OUTPUT=$(skopeo inspect ${SKOPEO_RETRIES} --no-tags --raw docker://${IMAGE})
```

theflockers: Ah ok. Now I got it. Just change the order.

FilipNikolovski (dismissed):
lgtm

utils/get-image-architectures
```diff
@@ -10,14 +10,35 @@
...
 
-IMAGE=$1
+show_help(){
+    echo "Usage: $1 [--skopeo-retries <#retries>] <image>"
+    exit 1
+}
+
+if [ ${#@} -eq 0 ]; then
+    show_help
+    exit
```
mmalina, line 20: The exit is dead code. show_help will exit already.

utils/get-image-architectures
FilipNikolovski: We need skopeo 1.19 which includes the fix I believe.

utils/get-image-architectures
```diff
@@ -10,14 +10,35 @@
...
+fi
+
+# get parameters
+for P in $@; do
```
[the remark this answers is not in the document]
theflockers: changed. A ticket was added to add shellcheck to the repository.

utils/get-image-architectures
theflockers: Hum.. this might be tricky. The version we using of skopeo is the binary from the rhel repo. Not sure we already have the binary for it already.

FilipNikolovski: It seems it hasn't landed yet, the version is 1.18.1.

mmalina: Hmm. And I checked ubi10 and that also has 1.18.1 - I was hoping it would have the newer. So we'll have to decide what to do here - maybe we'll have to put this on hold until we can get the newer version of skopeo.

mmalina: Yeah, ok.

johnbieren: I'm confused - I saw this and it made me think we are on hold until we update to a version of skopeo with this flag. However, the catalog PR to add the flag already merged. Shouldn't that one have been on hold too? Or is the thinking just that it doesn't hurt for now so might as well add it

FilipNikolovski: Afaik the [fix](https://github.com/containers/common/issues/2299) improves the retry logic to also retry on 5xx http errors, so it shouldn't hurt to add it now?

johnbieren:
> This is a strange gitlint error, has anyone come across this? [B4 rule](https://jorisroovers.com/gitlint/dev/rules/builtin_rules/#b4-body-first-line-empty) says that the 2nd line in the commit message should be empty, which it is?

I think it is complaining about the 2nd and 3rd commits, not the first one (not even sure why there are 3 commits instead of 1 in this pr)

utils/get-image-architectures
```diff
@@ -10,14 +10,37 @@
...
+if [ ${#@} -eq 0 ]; then
+    show_help
+fi
+
+while [[ $# -gt 0 ]]; do
+    case "$1" in
+        "--skopeo-retries")
+            shift
+            SKOPEO_RETRIES=(--retry-times "$1")
+            ;;
```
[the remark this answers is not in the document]
mmalina: Thanks. This works, but maybe a slightly nicer version is:
```
        "--skopeo-retries")
            SKOPEO_RETRIES=(--retry-times "$2")
            shift 2
            ;;
```

PR #478, opened 2025-07-17
Put file locking around writes to sbom files

ralphbean:
If we do this in parallel and if a pullspec is listed in a snapshot twice, then the two routines collide, one fetching and the other writing the augmented sbom - resulting in malformed json.


arewm (approved):
I have verified this change on a pipeline that was previously failing with duplicate component images.

(I am not configured as a CODEOWNER though)

johnbieren (approved):

jedinym:
> @jedinym can you make sure this is incorporated into mobster as well?

We have it in the current sprint:
https://issues.redhat.com/browse/ISV-6147

PR #501, opened 2025-08-12
feat(RELEASE-1240): add generic retry script

mmalina:
The main goal is to implement
retry for kinit calls in the release-service-catalog. But it's a generic function, so it can be used for other things in the future as well.


davidmogar:
> > Great script! Just one comment. I see that the time between attempts is linear. Should we implement a backoff strategy?
> 
> I would say this is good enough. But if you think it should be exponential, I can modify it.

I also think this is good enough. It was more of an open question. If we face any issue (service hammering) we can quickly adapt it.

PR #508, opened 2025-08-19
feat: output mediaType from get-image-archs

yftacherzog:
Include artifact's config.mediaType in the output of get-image-architectures so it can later be used by apply-mapping.

### DIFF
--- a/utils/get-image-architectures
+++ b/utils/get-image-architectures
@@ -57,8 +57,13 @@ if [ "$ARTIFACT_TYPE" != "null" ] || [ "$CONFIG_MEDIA_TYPE" == "application/vnd.
 
     # Just report that the image is for linux/amd64, which is not exactly true - but,
     # downstream release-service-catalog tasks expect to find something. Use this as a default.
-    jq -cr -n --arg digest "$digest" \
-        '{"platform": {"architecture": "amd64", "os": "linux"}, "digest": $ARGS.named["digest"], "multiarch": false}'
+    jq -cr -n --arg digest "$digest" --arg config_media_type "$CONFIG_MEDIA_TYPE" \
+        '{
+            "platform": {"architecture": "amd64", "os": "linux"},
+            "digest": $ARGS.named["digest"],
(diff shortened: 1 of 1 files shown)


utils/get-image-architectures
```diff
@@ -45,8 +45,12 @@ fi
...
 
 ARTIFACT_TYPE=$(jq -r '.artifactType' <<< $RAW_OUTPUT)
-CONFIG_MEDIA_TYPE=$(jq -r '.config.mediaType' <<< $RAW_OUTPUT)
-if [ "$ARTIFACT_TYPE" != "null" ] || [ "$CONFIG_MEDIA_TYPE" == "application/vnd.cncf.helm.config.v1+json" ] ; then
+
+# Extract and convert config.mediaType: raw string to proper JSON value for use with --argjson
+CONFIG_MEDIA_TYPE=$(jq -cn --arg config_media_type "$(jq -r '.config.mediaType' <<< $RAW_OUTPUT)" \
+    'if $ARGS.named["config_media_type"] == "null" then null else $ARGS.named["config_media_type"] end')
+if [ "$ARTIFACT_TYPE" != "null" ] || \
+   [ "$(jq -r '.' <<< "$CONFIG_MEDIA_TYPE")" == "application/vnd.cncf.helm.config.v1+json" ] ; then
```
mmalina, line 53: Why complicate this? Why not just inject the value just for this `if` branch? That way it won't even be set in any of the other branches and in apply-mapping, you can just check
```
config_media_type="$(jq -rs '.[0].configMediaType' <<< "$arch_json")"
if [ "$config_media_type" = "application/vnd.cncf.helm.config.v1+json" ]; then ...
```
and be done with it.

yftacherzog: no problem

mmalina (approved):

utils/get-image-architectures
mmalina: It looks good, thanks.

PR #528, opened 2025-09-17
feat(RELEASE-376): print pipelineRun in wait-for-internal-request

johnbieren:
This commit modifies the wait-for-internal-request script to print out the pipelineRun in its loop. It will only be printed out if the IR is running or done. It is added to the existing "running" and "succeeded" prints.


johnbieren:
Before:
```
Checking IR statuses...
Found 1 InternalRequests matching the name or label
Conditions:
  pipeline-7vzhk: succeeded
```
After:
```
Checking IR statuses...
Found 1 InternalRequests matching the name or label
Conditions:
  pipeline-6phpg: succeeded - pipelineRun: default/internalrequest-h7vn9
```
(same for running)

jinqi7 (approved):

PR #534, opened 2025-09-24
fix: preserve boolean false values in JSON merge

seanconroy2021:
The merge-json script was using the jq // operator which treats boolean false as falsy, causing it to fall back to the first JSON value instead of preserving the false value from the second JSON.

This caused singleComponentMode: false in RPA to be converted to null during merging leading to check-data-key failing.

Changed the merge logic to check for null values instead of using // allowing boolean false values to be preserved.


johnbieren (dismissed):

seanconroy2021:
Ok, so I've got it working by removing the has(.)

FilipNikolovski (approved):

johnbieren (approved):

PR #540, opened 2025-10-06
fix: handle gh rate limits and null responses in promote-overlay.sh

FilipNikolovski:
Fix the promote-overlay script to avoid GitHub rate limiting errors and handle null/empty API responses.


elenagerman:
It is not entirely clear what the acceptance criteria and final proposal for this fix are.
- your solution only checks for literal "null", not empty responses.
- you are not checking if this is an actual rate-limiting error or any other, I'm not sure how it is important.
- in such cases, it would probably be good/worth to implement a retry.
- I'm not a big fan of using static sleeps, but in any case, if there's no other option, there are still a few questions: 
          - why exactly 2 seconds? 
          - we have this sleep even on the first iteration, where it is unnecessary.

elenagerman:
> > It is not entirely clear what the acceptance criteria and final proposal for this fix are.
> > 
> > * your solution only checks for literal "null", not empty responses.
> > * you are not checking if this is an actual rate-limiting error or any other, I'm not sure how it is important.
> > * in such cases, it would probably be good/worth to implement a retry.
> > * I'm not a big fan of using static sleeps, but in any case, if there's no other option, there are still a few questions:
> >   
> >   * why exactly 2 seconds?
> >   * we have this sleep even on the first iteration, where it is unnecessary.
> 
> There's no jira issue for this, it's just an issue i've encountered when trying to promote the [release-service](https://issues.redhat.com/browse/RELEASE-service) to staging - it had a lot of PRs from development which we haven't promoted to staging (over 30), so the script failed due to rate limiting issues (authenticated requests have a 30 req/min which we hit). The script creates a PR and populates the description with the commits like so: [redhat-appstudio/infra-deployments#8521](https://github.com/redhat-appstudio/infra-deployments/pull/8521) (this is where we call gh api to fetch the PRs).
> 
> This is just a workaround for this problem, initially I tried to update the script to check for rate-limiting errors but that seemed to complicated since we have to parse the response headers, and just adding a "sleep 2" seems to do the job.

I see
If I remember correctly, we had a very similar problem for the release-service-catalog. 
And we integrated the GraphQL API solution instead of the standard one to solve it: https://github.com/konflux-ci/release-service-catalog/pull/1404.
Perhaps we should do the same here?
We can accept this MR as a workaround solution and open a new Jira to replace the logic with GraphQL. WDYT?

elenagerman:
> > > It is not entirely clear what the acceptance criteria and final proposal for this fix are.
> > > 
> > > * your solution only checks for literal "null", not empty responses.
> > > * you are not checking if this is an actual rate-limiting error or any other, I'm not sure how it is important.
> > > * in such cases, it would probably be good/worth to implement a retry.
> > > * I'm not a big fan of using static sleeps, but in any case, if there's no other option, there are still a few questions:
> > >   
> > >   * why exactly 2 seconds?
> > >   * we have this sleep even on the first iteration, where it is unnecessary.
> > 
> > 
> > There's no jira issue for this, it's just an issue i've encountered when trying to promote the [release-service](https://issues.redhat.com/browse/RELEASE-service) to staging - it had a lot of PRs from development which we haven't promoted to staging (over 30), so the script failed due to rate limiting issues (authenticated requests have a 30 req/min which we hit). The script creates a PR and populates the description with the commits like so: [redhat-appstudio/infra-deployments#8521](https://github.com/redhat-appstudio/infra-deployments/pull/8521) (this is where we call gh api to fetch the PRs).
> > This is just a workaround for this problem, initially I tried to update the script to check for rate-limiting errors but that seemed to complicated since we have to parse the response headers, and just adding a "sleep 2" seems to do the job.
> 
> I see If I remember correctly, we had a very similar problem for the [release-service-catalog](https://issues.redhat.com/browse/RELEASE-service-catalog). And we integrated the GraphQL API solution instead of the standard one to solve it: [konflux-ci/release-service-catalog#1404](https://github.com/konflux-ci/release-service-catalog/pull/1404). Perhaps we should do the same here? We can accept this MR as a workaround solution and open a new Jira to replace the logic with GraphQL. WDYT?

Please let me know if you'd like me to open a new Jira for this

PR #670, opened 2026-03-02
feat(PVSEC-12961): add rh-model-signing to image

jrepucci-rh:
I am currently creating a Konflux task for AI model signing in the [release-service-catalog](https://github.com/konflux-ci/release-service-catalog/tree/development) repo. The preferred way to do this is with the new [Red Hat model-signing tool](https://pypi.org/project/rh-model-signing/0.1.0/#description), which I propose we add to this image.

JIRA link: [PVSEC-12961](https://issues.redhat.com/browse/PVSEC-12961)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -68,7 +68,7 @@ RUN dnf -y --setopt=tsflags=nodocs install \
     git \
     git-lfs \
     jq \
-    python3-devel \
+    python3.10-devel \
     diffutils \
     python3-pip \
     python3-requests \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "release-service-utils"
 version = "0.1.0"
 description = "Konflux Release Service Utils"
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
     "jinja2",
     "check-jsonschema",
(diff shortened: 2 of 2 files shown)


querti:
Could you please change the first commit line to follow conventional commits specification? Similar to: https://github.com/konflux-ci/release-service-utils/pull/654

swickersh (commented):
I see the model-signing package is in the approved feature refinement doc. So this LGTM other than resolving the failing check.

We currently require several approvals up the chain before PRs are merged into main. So I'd expect at least a few days before this will get merged.

jrepucci-rh:
According to the failed test, we needed the container to run on Python 3.10 rather than 3.9. However, I realize that may be a bigger change than just downloading a single package

jrepucci-rh:
Thank you for the heads up! I'll keep an eye on that ticket

midnightercz:
Btw, just question. How much is rh-model-signing just wrapper for cosign cli? Because from the documentation it looks like it's exactly just it. I guess I understand there are probably some complex use cases which you can support by running single python command and bash equivalent of those would be complicated if you want to run them in tekton tasks directly. But new dependency add more complexity to dependency management and eventually makes it dependency hell. So I'm wondering if it would be better to have these use cases covered by small script directly hosted here. Or do you use the tool also elsewhere?

PR #699, opened 2026-03-31
feat(RELEASE-2209): add helper scripts to create catalog fork

johnbieren:
This commit is part of a multi commit effort to enable running the catalog pipeline e2e test suites for PRs to this repo. This commit adds a bash script that will push a fork of the catalog repo with the utils imageRefs updated to one built from the PR. It also includes a python file (with unit tests) to cleanup the created fork once testing is complete.

We can then use this catalog fork via git resolver in the e2e test suites to test with our new image.

Assisted-By: Cursor

### DIFF
--- a/integration-tests/lib/catalog_clone_patch_push.sh
+++ b/integration-tests/lib/catalog_clone_patch_push.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# Clone release-service-catalog, replace konflux-ci release-service-utils image refs with UTILS_IMAGE,
+# push to a new GitHub repo.
+# Outputs results for Tekton (stdout markers + optional result files).
+#
+# Required env: GITHUB_TOKEN, UTILS_IMAGE
+# Optional: CATALOG_REPO (default konflux-ci/release-service-catalog), CATALOG_REF (development),
+#           DEST_REPO_PREFIX (default hacbs-release-tests/catalog-e2e), PIPELINE_UID
+set -euo pipefail
+
+: "${GITHUB_TOKEN:?GITHUB_TOKEN is required}"
+: "${UTILS_IMAGE:?UTILS_IMAGE is required}"
+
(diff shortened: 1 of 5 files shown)


swickersh (commented):
I use cursor to help review PRs locally. I'm starting to think it sees "Assisted-by Cursor" and leaves glowing reviews 😄 

`This is solid work — clean code, good tests, sensible architecture for a real-world problem (testing image changes against downstream pipelines). The security concern about tokens in URLs is the most notable item, though it matches existing patterns in Tekton-based repos. The missing timeout on delete-repository.sh and the unhandled TimeoutExpired in _ls_remote_head are worth fixing before merge. Everything else is minor.`

johnbieren:
qodo gave one comment instead of opening threads (which is kinda nice) so I don't know how to respond to it. But, I implemented the first 2 changes (anonymous clone of the catalog repo + timeout in the cleanup script call), but not the third

integration-tests/lib/catalog_clone_patch_push.sh
```diff
@@ -0,0 +1,122 @@
...
+  echo "Removing .github/workflows before image patch (required for PAT push of development ref)."
+  rm -rf .github/workflows
+  git config user.email "catalog-e2e@konflux-ci"
+  git config user.name "konflux-release-team"
+  git add -A
+  git commit -m "chore(e2e): drop workflows for GitHub PAT push"
+fi
+
+export UTILS_IMAGE
+python3 << 'PATCHPY'
```
swickersh: What's the reasoning for this nested python section?
Why not call a separate python script from this shell script?

Or better yet, why do catalog_clone_patch_push in a bash script at all and not just make the whole thing python?
I'll still give LGTM either way, but this pattern feels off to me.

swickersh (dismissed):

swickersh (approved):

PR #704, opened 2026-04-06
fix(KONFLUX-12712): drop cgw prefix from shortURL

swickersh:
When mirrorOpenshiftPush is not set or false, the shortURL prefix was being set to /cgw, producing unnecessary folder nesting in CGW URLs. Now the prefix is empty for non-MOC releases, resulting in shortURLs of the form /{productCode}/{productVersion}/{filename}. The /pub/cgw prefix is still used when mirrorOpenshiftPush is true.

Assisted-by: Cursor AI


elenagerman (approved):

seanconroy2021 (approved):

johnbieren (approved):

davidmogar (approved):

ach912 (approved):

johnbieren:
@swickersh idk how it passes on the PR check and not the merge queue check, but this is failing gitlint for the title being 74 characters (72 max). Can you shorten it so we can merge?

swickersh:
> @swickersh idk how it passes on the PR check and not the merge queue check, but this is failing gitlint for the title being 74 characters (72 max). Can you shorten it so we can merge?

done

seanconroy2021 (approved):

PR #711, opened 2026-04-08
feat(RELEASE-2209): add script to execute catalog e2e plr

johnbieren:
This commit is part of a multi commit effort to enable running the catalog pipeline e2e test suites for PRs to this repo. This commit adds a python module that will create a pipelineRun for catalog integration tests. After this is merged, the final piece will be to add the pipeline yaml file that orchestrates the entire workflow (with documentation).

Assisted-By: Cursor

### DIFF
--- a/integration-tests/lib/run_single_catalog_e2e_suite.py
+++ b/integration-tests/lib/run_single_catalog_e2e_suite.py
@@ -1,0 +1,339 @@
+#!/usr/bin/env python3
+"""Create one PipelineRun for catalog integration tests and wait for completion.
+
+Usage: python3 run_single_catalog_e2e_suite.py
+
+Invokes the pipeline defined in catalog at
+``integration-tests/pipelines/e2e-tests-staging-pipeline.yaml``.
+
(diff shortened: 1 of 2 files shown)


johnbieren:
Depends on https://github.com/konflux-ci/release-service-catalog/pull/2107

integration-tests/lib/run_single_catalog_e2e_suite.py
```diff
@@ -0,0 +1,337 @@
+#!/usr/bin/env python3
+"""Create one PipelineRun for catalog integration tests and wait for completion.
```
mmalina, line 2: I was gonna ask for usage examples, but it seems it relies solely on env vars, so there is not much to it. But saying `Usage: python3 run_single_catalog_e2e_suite.py` or similar might help to make it obvious?

johnbieren: Added. Fwiw, this script wouldn't ever be called standalone. It is just called as part of the larger workflow (either ITS or run-test.sh is the entrypoint)

mmalina (approved):

seanconroy2021 (approved):
LGTM

PR #720, opened 2026-04-14
feat(RELEASE-2386): add optional 404 retry to curl-with-retry

mmalina:
Jira issue GET can intermittently return 404 on Atlassian Cloud even when the issue exists and the token is valid. Add optional handling via CURL_WITH_RETRY_RETRY_404: up to three attempts (two retries) with exponential backoff. Keep 429 handling at five attempts. Refactor to a while loop with separate counters.

### DIFF
--- a/utils/curl-with-retry
+++ b/utils/curl-with-retry
@@ -3,12 +3,12 @@
 # script:      curl-with-retry
 #
 # description: This script handles Jira API rate limiting by retrying requests with exponential backoff.
-#              It will retry up to 5 times, and will wait for a random amount of time between retries.
-#              There are a few caveats:
-#              - The script will only retry if the response code is 429 (rate limited).
-#              - You can still use --retry as needed to retry 5xx errors.
-#              - -o/--output should not be specified - the response will always be printed to stdout.
-#              - --fail should not be specified. The script will always fail if the request fails.
+#              429 responses are retried up to MAX_429_ATTEMPTS times (sleeps only before another attempt).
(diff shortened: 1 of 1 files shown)


utils/curl-with-retry
```diff
@@ -3,40 +3,66 @@
...
 #              curl-with-retry
 #
 
 set -eu
 
-#!/bin/bash
-
-MAX_RETRIES=5
+MAX_429_ATTEMPTS=5
+MAX_404_ATTEMPTS=3
```
FilipNikolovski, line 20: Why should we retry on 404, shouldn't we retry only on transient errors?

mmalina: Did you read the description above? :) Also, the jira has even more context. I saw this in e2e where the embargo check would fail on this sometimes - the jira api call would sometimes return 404 for the testing jira that we use. Normally 404 means a jira either does not exist, or you don't have permissions to see it.

> shouldn't we retry only on transient errors?

Normal transient errors are covered by `curl --retry 3` which we have everywhere. This script is a wrapper that adds more on top of that.

swickersh (approved):
this lgtm

utils/curl-with-retry
FilipNikolovski: Apologies, went straight to the code :sweat_smile: thanks for the context.

FilipNikolovski (approved):
lgtm

johnbieren (approved):

ach912 (approved):

davidmogar (approved):

PR #722, opened 2026-04-15
feat(RELEASE-1193): added subst_template script

midnightercz:
New script replaces '{{var}}' style templates with variable provided in data file

### DIFF
--- a/utils/subst_template.py
+++ b/utils/subst_template.py
@@ -1,0 +1,331 @@
+#!/usr/bin/env python3
+"""Jinja2 template substitution with custom label extension support."""
+
+import argparse
+import json
+import logging
+import sys
+
+from typing import Any, Dict, List, Optional
+
+from jinja2.sandbox import SandboxedEnvironment, Environment
+from jinja2 import DebugUndefined, StrictUndefined
+from jinja2.lexer import Lexer, Token, TokenStream
+
+LOGGER = logging.getLogger("subst_template")
+
+
+class CustomLexer(Lexer):
+    """Custom Lexer that modifies tokenization to not treat '-' as a mathematical operator.
+
+    This allows strings like 'foo-bar' in templates to be treated as identifiers
+    rather than being parsed as mathematical expressions (foo minus bar).
+    """
+
+    def tokenize(self, source, name=None, filename=None, state=None):
+        """Merge hyphenated identifiers into single tokens."""
+        # Get token stream from original tokenizer
+        token_stream = super().tokenize(source, name, filename, state)
+
+        # Convert to list to process
+        tokens = list(token_stream)
+
+        # Process tokens to merge hyphenated names
+        result = []
+        i = 0
+        while i < len(tokens):
+            if (
+                i + 2 < len(tokens)
+                and tokens[i].test("name")
+                and tokens[i + 1].test("sub")
+                and tokens[i + 2].test("name")
+            ):
+                # Merge NAME - NAME into a single NAME token
+                merged_value = f"{tokens[i].value}-{tokens[i + 2].value}"
+                merged_token = Token(tokens[i].lineno, "name", merged_value)
+                result.append(merged_token)
+                i += 3  # Skip the three tokens we just merged
+            else:
+                result.append(tokens[i])
+                i += 1
+
+        # Return TokenStream instead of plain iterator
+        return TokenStream(result, name, filename)
+
+
+class CustomEnvironment(SandboxedEnvironment):
+    """Custom Jinja2 Environment that uses CustomLexer for tokenization."""
+
+    def _tokenize(self, source, name, filename=None, state=None):
+        """Override _tokenize to use our custom lexer."""
+        # Create our custom lexer if not already created
+        if not hasattr(self, "_custom_lexer"):
+            self._custom_lexer = CustomLexer(self)
+        return self._custom_lexer.tokenize(source, name, filename, state)
+
+
+class LabelsProvider:
+    """A custom class to provide label access in Jinja2 templates.
+
+    Supports both simple and nested label access:
+    - {{labels.mylabel}} - accesses label "mylabel"
+    - {{labels.mylabel.with-dash}} - accesses label "mylabel.with-dash"
+
+    The class builds up the label path through attribute access and resolves
+    it when the value is needed.
+    """
+
+    def __init__(self, labels: Dict[str, Any], path: str = "", strict: bool = False):
+        """Initialize the LabelsProvider.
+
+        Args:
+            labels: Dictionary containing all available labels
+            path: Current path being built (used internally for nested access)
+            strict: Whether to raise errors on missing labels
+
+        """
+        self._labels = labels
+        self._path = path
+        self._strict = strict
+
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "release-service-utils"
 version = "0.1.0"
 description = "Konflux Release Service Utils"
 readme = "README.md"
-requires-python = ">=3.10"
+requires-python = ">=3.9"
```
FilipNikolovski: Why lower the version number?

davidmogar:
This PR has no internal reviews. Marking it as not ready. Please update the sheet once it receives an approval.

utils/subst_template.py
```diff
@@ -0,0 +1,284 @@
...
+                i += 3  # Skip the three tokens we just merged
+            else:
+                result.append(tokens[i])
+                i += 1
+
+        # Return TokenStream instead of plain iterator
+        return TokenStream(result, name, filename)
+
+
+class CustomEnvironment(Environment):
```
FilipNikolovski: Should we perhaps use a SandboxedEnvironment instead of Environment here? Don't know if templates come from trusted sources or not, if it comes from a user provided template, they could execute arbitrary python code via certain expressions in the template. SandboxedEnvironment prevents this: https://jinja.palletsprojects.com/en/stable/sandbox/

utils/subst_template.py
```diff
@@ -0,0 +1,284 @@
...
+        tokens = list(token_stream)
+
+        # Process tokens to merge hyphenated names
+        result = []
+        i = 0
+        while i < len(tokens):
+            if (
+                i + 2 < len(tokens)
+                and tokens[i].test("name")
+                and tokens[i + 1].test("sub")
```
FilipNikolovski, line 40: This would break subtraction for the whole template, are we okay with this?

theflockers:
LGTM once Fitz's comments are clear.

utils/subst_template.py
midnightercz: That's why it's enabled only when `--labels-ext` is set. Of course it's not possible to use substractions in templates when `--labels-ext` is set, but 1. we don't use it anywhere yet, 2. You can run script twice if you need to substitute labels and substractions in same template

utils/subst_template.py
midnightercz: I wasn't aware of that. Great tip!

pyproject.toml
midnightercz: I don't remember, I guess maybe uv was locked to 3.9 so I changed to be in sync

utils/subst_template.py
FilipNikolovski: Ok, is it possible to narrow the vars scope perhaps, are the vars strictly {{ foo-bar }} or they come from some object like {{labels.foo-bar}}?

midnightercz: I don't think it's possible. Lexer is applied for whole string. Labels provider is running after everything is "lexed" already

FilipNikolovski: Ok as long as we're aware of this I'm cool :+1:

FilipNikolovski (dismissed):
lgtm

utils/rsutils/bootstrap.py
johnbieren, line 1: What is rsutils? Release service utils instead the utils dir of the release-service-utils image? That doesn't sound good to me. Why can't we just have it in the normal utils dir?

johnbieren, line 1: Why is it called bootstrap if all it does is set up a logger?

pyproject.toml
johnbieren: Why are you making this change? I'd think something else called for python 3.10

Dockerfile
```diff
@@ -89,9 +89,13 @@ RUN dnf -y --setopt=tsflags=nodocs install \
 RUN curl -LO https://github.com/release-engineering/exodus-rsync/releases/latest/download/exodus-rsync && \
     chmod +x exodus-rsync && mv exodus-rsync /usr/local/bin/rsync
 
+# Copy utils before installation
+COPY utils /home/utils
+
 # Install Python dependencies using uv
-COPY pyproject.toml uv.lock ./
-RUN uv pip install -r pyproject.toml --system && \
+COPY README.md pyproject.toml uv.lock /home/
```
johnbieren: Why do we need the README inside /home?

utils/rsutils/bootstrap.py
midnightercz: utils is filled with combinations of bash and python executables. rsutils is meant to be python module which you can import. Also this was done before you started moving things to scripts/python/helpers/ - which is also not distributed as python module btw. I'll move it there

pyproject.toml
midnightercz: Reason for this was simple, in the docker file there's still registry.access.redhat.com/ubi9/ubi:9.7-1777858048. So pyproject requiring > 3.10 simply doesn't work. It was ok in the past as pyproject was used only as install -r pyproject.toml which does install only requirements from it but doesn't really respect the required python version. Now when I wanted to package everything into release_utils python module, I had to fix it

Dockerfile
midnightercz: because README is needed by pyproject (because there's just readme = "README.md" so it has to be in the same dir) if you use it to install the project (not just deps)

utils/rsutils/bootstrap.py
midnightercz: well, idea was to add more in the future

johnbieren: I still think "rsutils" in a release-service-utils repo is a pretty bad name. Might as well call it `foo`

johnbieren: Can't we combine with https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/logger.py? The bash to python conversion is a large effort that I asked many times for feedback on. If it was missed, that's fine, but I don't think we should have two separate python ways to set up a logger in the same repo. Please take a look at what now exists and see what can be reused. Anything can be moved or extended

pyproject.toml
johnbieren: okay, thanks for the explanation. It is going to conflict with https://github.com/konflux-ci/release-service-utils/pull/770 though

utils/rsutils/bootstrap.py
midnightercz: Yeah, I'll merge it to the helpers.

FilipNikolovski: +1 on having the project as a python module, don't have any suggestion for the naming unfortunately.

midnightercz: fixed

midnightercz: removed

johnbieren (dismissed):

ach912 (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-02T07:32:10.233Z |

Approved

theflockers (approved):

PR #734, opened 2026-04-22
fix(KONFLUX-13012): clean up orphaned IRs before creating new ones

swickersh:
When the managed task is retried (e.g. due to timeout), the previous InternalRequest and its PipelineRun on the internal cluster continue running. The retry creates a new InternalRequest, leading to concurrent operations that cause duplicate Pulp pushes and CGW "Record already present" errors.

Before creating a new InternalRequest, the script now deletes any existing InternalRequests with the same PipelineRun UID label. Combined with the finalizer change in[ internal-services(PR 709)](https://github.com/konflux-ci/internal-services/pull/709), this ensures the associated PipelineRun is cancelled before the new one starts.

Assisted-by: Cursor AI

### DIFF
--- a/utils/internal-request
+++ b/utils/internal-request
@@ -48,7 +48,7 @@
 # Note: This script is intended to be used with a specific Kubernetes API
 # that includes the 'InternalRequest' resource type.
 
-set -e
+set -eo pipefail
 
 # Set defaults
 SYNC=true
@@ -258,10 +258,62 @@ if [[ -n ${LABELS[@]} ]]; then
         --argjson labels "$LABEL_JSON" \
         '.metadata.labels += $labels' <<< $PAYLOAD)
 fi
+# Always stamp the pipeline name onto the IR so the cleanup selector can scope
+# deletions to this specific pipeline, preventing accidental removal of IRs
+# created by other tasks running in parallel within the same PipelineRun.
+PIPELINE_NAME_LABEL="internal-services.appstudio.openshift.io/pipeline-name"
+PAYLOAD=$(jq \
+    --arg key "$PIPELINE_NAME_LABEL" \
+    --arg value "$PIPELINE" \
+    '.metadata.labels += {($key): $value}' <<< $PAYLOAD)
 if [[ -n "${SERVICEACCOUNT}" ]]; then
     PAYLOAD=$(jq \
         --arg serviceAccount "$SERVICEACCOUNT" \
         '.spec.serviceAccount = $serviceAccount' <<< $PAYLOAD)
+fi
+
+# Clean up any existing InternalRequests from prior attempts within the same
+# PipelineRun. This handles cases where a managed task retry or timeout left
+# an orphaned InternalRequest whose PipelineRun may still be running on the
+# internal cluster. The internal-services controller will cancel the associated
+# PipelineRun when it detects the InternalRequest deletion.
+#
+# The selector uses both pipelinerun-uid (supplied by the caller) and the
+# pipeline-name that was auto-stamped on the IR above. Combining them scopes
+# the cleanup to only IRs from this specific pipeline invocation, so parallel
+# tasks in the same PipelineRun that call different pipelines are never
+# accidentally deleted.
+PIPELINERUN_UID_LABEL="internal-services.appstudio.openshift.io/pipelinerun-uid"
+PIPELINERUN_UID_VALUE=""
+for label in "${LABELS[@]}"; do
+    KEY=$(echo "$label" | cut -d'=' -f1)
+    VALUE=$(echo "$label" | cut -d'=' -f2-)
+    if [ "$KEY" = "$PIPELINERUN_UID_LABEL" ]; then
+        PIPELINERUN_UID_VALUE="$VALUE"
+        break
+    fi
+done
+
+if [ -n "$PIPELINERUN_UID_VALUE" ]; then
+    LABEL_SELECTOR="${PIPELINERUN_UID_LABEL}=${PIPELINERUN_UID_VALUE},${PIPELINE_NAME_LABEL}=${PIPELINE}"
+    EXISTING_IRS=$(kubectl get internalrequest \
+        -l "$LABEL_SELECTOR" \
+        -o json | jq -r '.items[].metadata.name')
+
+    if [ -n "$EXISTING_IRS" ]; then
+        echo "Found existing InternalRequests from prior attempts. Cleaning up..."
+        while IFS= read -r IR_NAME; do
+            echo "Deleting InternalRequest ${IR_NAME}..."
+            kubectl delete internalrequest "$IR_NAME" --wait=true --timeout=60s
+        done <<< "$EXISTING_IRS"
+        # Sleep to give the internal-services controller time to react to the IR
+        # deletion and cancel the associated PipelineRun before we create a new one.
+        # This is a best-effort heuristic; --wait=true above ensures the IR is gone
+        # from the API server, but cancellation propagation to the internal cluster
+        # is asynchronous and its duration is not observable from here.
+        echo "Cleanup complete. Waiting for PipelineRun cancellation to propagate..."
+        sleep 5
+    fi
 fi
 
 # Create InternalRequest using kubectl
--- a/utils/tests/test_internal_request.py
+++ b/utils/tests/test_internal_request.py
@@ -1,0 +1,184 @@
+"""Tests for the internal-request utility script."""
+
+from __future__ import annotations
+
+import os
+import stat
+import subprocess
+from pathlib import Path
+
+SCRIPT_PATH = Path(__file__).resolve().parents[1] / "internal-request"
+
+
+def _write_executable(path: Path, content: str) -> None:
+    path.write_text(content, encoding="utf-8")
+    path.chmod(path.stat().st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
+
+
+def _run_internal_request(
+    tmp_path: Path,
+    labels: list[str] | None = None,
+    existing_irs: list[str] | None = None,
+    delete_fails: bool = False,
+):
+    bin_dir = tmp_path / "bin"
+    bin_dir.mkdir()
+    kubectl_log = tmp_path / "kubectl.log"
+    sleep_log = tmp_path / "sleep.log"
+
+    # Build a JSON array of IR objects for the mock get response.
+    ir_names = existing_irs or []
+    items_json = ", ".join(f'{{"metadata": {{"name": "{name}"}}}}' for name in ir_names)
+    mock_get_json = f'{{"items": [{items_json}]}}'
+
+    delete_exit = "1" if delete_fails else "0"
+
+    _write_executable(
+        bin_dir / "kubectl",
+        f"""#!/usr/bin/env bash
+set -euo pipefail
+echo "$*" >> "$KUBECTL_LOG"
+if [[ "${{1:-}}" == "get" && "${{2:-}}" == "internalrequest" ]]; then
+    echo '{mock_get_json}'
+    exit 0
+fi
+if [[ "${{1:-}}" == "delete" && "${{2:-}}" == "internalrequest" ]]; then
+    exit {delete_exit}
+fi
+if [[ "${{1:-}}" == "create" ]]; then
+    cat >/dev/null
+    echo '{{"metadata":{{"name":"new-ir"}}}}'
+    exit 0
+fi
+exit 1
+""",
+    )
+
+    _write_executable(
+        bin_dir / "sleep",
+        """#!/usr/bin/env bash
+set -euo pipefail
+echo "$*" >> "$SLEEP_LOG"
+""",
+    )
+
+    cmd = [
+        "bash",
+        str(SCRIPT_PATH),
+        "--pipeline",
+        "test-pipeline",
+        "-p",
+        "taskGitUrl=https://github.com/konflux-ci/release-service-catalog",
+        "-p",
+        "taskGitRevision=main",
+        "-s",
+        "false",
+    ]
+    for label in labels or []:
+        cmd.extend(["-l", label])
+
+    env = os.environ.copy()
+    env["PATH"] = f"{bin_dir}:{env['PATH']}"
+    env["KUBECTL_LOG"] = str(kubectl_log)
+    env["SLEEP_LOG"] = str(sleep_log)
+
+    result = subprocess.run(cmd, capture_output=True, text=True, check=False, env=env)
+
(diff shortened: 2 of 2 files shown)


utils/internal-request
```diff
@@ -264,6 +264,37 @@ if [[ -n "${SERVICEACCOUNT}" ]]; then
...
+
+if [ -n "$PIPELINERUN_UID_VALUE" ]; then
+    EXISTING_IRS=$(kubectl get internalrequest -l "${PIPELINERUN_UID_LABEL}=${PIPELINERUN_UID_VALUE}" \
+        --no-headers -o custom-columns=":metadata.name" 2>/dev/null || true)
+
+    if [ -n "$EXISTING_IRS" ]; then
+        echo "Found existing InternalRequests from prior attempts. Cleaning up..."
+        for IR_NAME in $EXISTING_IRS; do
+            echo "Deleting InternalRequest ${IR_NAME}..."
+            kubectl delete internalrequest "$IR_NAME" 2>/dev/null || true
```
theflockers: The deletion would happen right away? Should it check afterwards if the IR was really deleted?
Wondering cases where for any reason the controller might be eventually down and unable to delete the PR and the IR. - sorry, I know it is not ready for review, but just notice the PR.

swickersh: Good point. Added --wait=true --timeout=60s to the delete call

utils/internal-request
```diff
@@ -264,6 +264,37 @@ if [[ -n "${SERVICEACCOUNT}" ]]; then
...
+
+if [ -n "$PIPELINERUN_UID_VALUE" ]; then
+    EXISTING_IRS=$(kubectl get internalrequest -l "${PIPELINERUN_UID_LABEL}=${PIPELINERUN_UID_VALUE}" \
+        --no-headers -o custom-columns=":metadata.name" 2>/dev/null || true)
+
+    if [ -n "$EXISTING_IRS" ]; then
+        echo "Found existing InternalRequests from prior attempts. Cleaning up..."
+        for IR_NAME in $EXISTING_IRS; do
+            echo "Deleting InternalRequest ${IR_NAME}..."
+            kubectl delete internalrequest "$IR_NAME" --wait=true --timeout=60s 2>/dev/null || true
```
seanconroy2021: Should we remove 2>/dev/null here? If the delete fails we lose all  output for debugging.

swickersh: Good catch, and agreed. In the updated version I've removed both 2>/dev/null and || true from the delete. Since the script runs under set -e, a failed delete will now surface the error output and exit non-zero, which prevents a new IR being created while an old one might still be running. I also switched the kubectl get to use -o json | jq instead of custom-columns to be consistent with how wait-for-internal-request does it.

utils/internal-request
```diff
@@ -264,6 +264,43 @@ if [[ -n "${SERVICEACCOUNT}" ]]; then
...
+    if [ "$KEY" = "$PIPELINERUN_UID_LABEL" ]; then
+        PIPELINERUN_UID_VALUE="$VALUE"
+        break
+    fi
+done
+
+if [ -n "$PIPELINERUN_UID_VALUE" ]; then
+    EXISTING_IRS=$(kubectl get internalrequest \
+        -l "${PIPELINERUN_UID_LABEL}=${PIPELINERUN_UID_VALUE}" \
+        -o json | jq -r '.items[].metadata.name')
```
seanconroy2021, line 301: The script runs with set -e but not set -o pipefail.  So if kubectl get fails in this pipe only jqs exit code is checked. jq would likely also fail on bad input so it should work but it might better to do set -o pipefail. Maybe I am overthinking that :)

swickersh: Couldn't hurt. Added, thanks!

seanconroy2021 (dismissed):

johnbieren (commented):
What about two managed tasks that both create internalRequests running in parallel for the same plr? Shouldn't we also use the taskRunID label?

swickersh:
> What about two managed tasks that both create internalRequests running in parallel for the same plr? Shouldn't we also use the taskRunID label?

@johnbieren really good catch! 
After your comment, I started to use taskrun-uid as the additional discriminator, but the problem is that every task calling internal-request would need to be updated to pass internal-services.appstudio.openshift.io/taskrun-uid=$(context.taskRun.uid), which is a larger refactor touching all the managed tasks in release-service-catalog.

Instead, I'm trying to have the internal-request script automatically stamp a internal-services.appstudio.openshift.io/pipeline-name label (from the --pipeline argument it already receives) onto every IR it creates. The cleanup selector then becomes pipelinerun-uid=X,pipeline-name=push-disk-images. Since each parallel task in a release pipeline calls a different internal pipeline, this scopes the cleanup to only IRs from this specific task and no changes needed in the calling tasks.

I'll run some tests to make sure this works.

johnbieren:
> > What about two managed tasks that both create internalRequests running in parallel for the same plr? Shouldn't we also use the taskRunID label?
> 
> @johnbieren really good catch! After your comment, I started to use taskrun-uid as the additional discriminator, but the problem is that every task calling internal-request would need to be updated to pass internal-services.appstudio.openshift.io/taskrun-uid=$(context.taskRun.uid), which is a larger refactor touching all the managed tasks in [release-service-catalog](https://issues.redhat.com/browse/RELEASE-service-catalog).
> 
> Instead, I'm trying to have the internal-request script automatically stamp a internal-services.appstudio.openshift.io/pipeline-name label (from the --pipeline argument it already receives) onto every IR it creates. The cleanup selector then becomes pipelinerun-uid=X,pipeline-name=push-disk-images. Since each parallel task in a release pipeline calls a different internal pipeline, this scopes the cleanup to only IRs from this specific task and no changes needed in the calling tasks.
> 
> I'll run some tests to make sure this works.

love it, good idea

swickersh:
> > > What about two managed tasks that both create internalRequests running in parallel for the same plr? Shouldn't we also use the taskRunID label?
> > 
> > 
> > @johnbieren really good catch! After your comment, I started to use taskrun-uid as the additional discriminator, but the problem is that every task calling internal-request would need to be updated to pass internal-services.appstudio.openshift.io/taskrun-uid=$(context.taskRun.uid), which is a larger refactor touching all the managed tasks in [release-service-catalog](https://issues.redhat.com/browse/RELEASE-service-catalog).
> > Instead, I'm trying to have the internal-request script automatically stamp a internal-services.appstudio.openshift.io/pipeline-name label (from the --pipeline argument it already receives) onto every IR it creates. The cleanup selector then becomes pipelinerun-uid=X,pipeline-name=push-disk-images. Since each parallel task in a release pipeline calls a different internal pipeline, this scopes the cleanup to only IRs from this specific task and no changes needed in the calling tasks.
> > I'll run some tests to make sure this works.
> 
> love it, good idea

@johnbieren 
Manually tested the cleanup logic against prd-rh02. Created a fake orphaned IR with the matching labels, then ran the script to simulate a retry:
```
# Create orphaned IR with pipelinerun-uid and pipeline-name labels
$ oc create -f - <<EOF
apiVersion: appstudio.redhat.com/v1alpha1
kind: InternalRequest
metadata:
  generateName: push-disk-images-orphan-
  namespace: rhtap-releng-tenant
  labels:
    internal-services.appstudio.openshift.io/pipelinerun-uid: "test-plr-uid-12345"
    internal-services.appstudio.openshift.io/pipeline-name: "push-disk-images"
spec:
  pipeline:
    pipelineRef:
      resolver: git
      params:
        - name: url
          value: https://github.com/konflux-ci/release-service-catalog
        - name: revision
          value: development
        - name: pathInRepo
          value: pipelines/internal/push-disk-images/push-disk-images.yaml
  params: {}
EOF
internalrequest.appstudio.redhat.com/push-disk-images-orphan-dplwv created
# Simulate retry - script detects and cleans up the orphan before creating a new IR
$ internal-request --pipeline "push-disk-images" \
  -p taskGitUrl=https://github.com/konflux-ci/release-service-catalog \
  -p taskGitRevision=development \
  -l internal-services.appstudio.openshift.io/pipelinerun-uid=test-plr-uid-12345 \
  -s false
Found existing InternalRequests from prior attempts. Cleaning up...
Deleting InternalRequest push-disk-images-orphan-dplwv...
internalrequest.appstudio.redhat.com "push-disk-images-orphan-dplwv" deleted from rhtap-releng-tenant namespace
Cleanup complete. Waiting for PipelineRun cancellation to propagate...
InternalRequest 'push-disk-images-6ld6b' created.
```
The orphan was deleted and a fresh IR created as expected.

johnbieren (dismissed):

ach912 (dismissed):

davidmogar (dismissed):

twaugh (approved):

ach912:
@swickersh the Qodo AI review flagged potential bugs or security concerns in this PR.
Please acknowledge each finding by adding a comment with the following format:

```markdown
## Qodo Response

**<Finding title>** — <Your explanation (false positive / fixed in commit / accepted risk)>
**<Another finding>** — <Your response>
```

> The approval dashboard will pick up this section and show it to the approver alongside the review.
> Replace the `<…>` placeholders with the actual finding titles from the Qodo review and your responses.

ach912:
@swickersh the Qodo AI review flagged potential bugs or security concerns in this PR.
Please acknowledge each finding by adding a comment with the following format:

```markdown
## Qodo Response

**<Finding title>** — <Your explanation (false positive / fixed in commit / accepted risk)>
**<Another finding>** — <Your response>
```

> The approval dashboard will pick up this section and show it to the approver alongside the review.
> Replace the `<…>` placeholders with the actual finding titles from the Qodo review and your responses.

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-05-21T12:29:10.426Z |

swickersh:
## Qodo Response

**Cleanup ignores delete failures** — Fixed in current code. Removed `2>/dev/null || true` from both the `kubectl get` and `kubectl delete` calls, and changed `set -e` to `set -eo pipefail` at the top of the script. If `kubectl delete ... --wait=true --timeout=60s` fails or times out, the script now exits non-zero, preventing a new IR from being created while an old one may still be running.

**`sleep 5` without PipelineRun check** — Accepted limitation with rationale. The associated PipelineRun runs on a separate internal cluster and is cancelled by the internal-services controller when it observes the IR deletion. That propagation is cross-cluster and asynchronous — there is no way to poll for it from within this script without privileged cross-cluster access that the managed task context does not have. The `sleep 5` is a best-effort grace period to let propagation start; this is explicitly called out in a code comment. Fully reliable synchronous cancellation would require infrastructure outside the scope of this script. The companion finalizer change in internal-services#709 ensures the IR is removed from the API server before the new one is created.

**Fragile IR list parsing** — Fixed in current code. Replaced `kubectl get ... -o custom-columns` with `-o json | jq -r '.items[].metadata.name'` for structured machine-readable output, and replaced the word-splitting `for IR_NAME in $EXISTING_IRS` loop with `while IFS= read -r IR_NAME; do ...; done`, consistent with the pattern already used in `utils/wait-for-internal-request`.

cc @ach912

swickersh:
@johnbieren since we reverted https://github.com/konflux-ci/internal-services/pull/709 should we wait to merge this?
afaict this wouldn't harm anything.  But I don't want to break prod..

johnbieren:
@swickersh gitlint is saying you are at 73 chars instead of 72 (idk why it passes on the PR but not in merge queue)

johnbieren:
> @johnbieren since we reverted [konflux-ci/internal-services#709](https://github.com/konflux-ci/internal-services/pull/709) should we wait to merge this? afaict this wouldn't harm anything. But I don't want to break prod..

:man_shrugging: you tell me. If you think it will break things, can you move it to draft? If you think it is a good change, we can merge it. It will need to be followed by a catalog update which then has to go through the promotion process

swickersh:
@johnbieren 

> @swickersh gitlint is saying you are at 73 chars instead of 72 (idk why it passes on the PR but not in merge queue)

This is because the merge queue commit appends 7 chars (a space plus (# 734)) to the original commit.

We'd need logic in the gitlint command to limit 65 chars in the original commit title.
I'll edit the commit title for now


>🤷‍♂️ you tell me. If you think it will break things, can you move it to draft? If you think it is a good change, we can merge it. It will need to be followed by a catalog update which then has to go through the promotion process

yea, sorry. That's on me to check closer. I checked the logic and walked through the scenario with Cursor but it steered me wrong on the 709 change that broke stage so I'm overly cautious now.

This is what cursor is telling me which seems to check out. It seems fine to merge prior to the finalizer change landing in internal-services
```
PR 734 deletes the orphaned IR via kubectl delete --wait=true, but without the finalizer from #709, the internal-services controller has nothing to hook into — the IR is just gone, and the associated internal PipelineRun continues running to natural completion. So duplicate concurrent runs can still occur until #709 lands.

This is not a regression: without 734, the orphaned IR and its PipelineRun would also keep running — there was never any cancellation mechanism in the current codebase. The kubectl delete here is intentionally setting up the hook point: once #709 is re-implemented, the controller will intercept that deletion via the finalizer and cancel the internal PipelineRun before the new IR is created.
```

johnbieren:
> @johnbieren
> 
> > @swickersh gitlint is saying you are at 73 chars instead of 72 (idk why it passes on the PR but not in merge queue)
> 
> This is because the merge queue commit appends 7 chars (a space plus (# 734)) to the original commit.
> 
> We'd need logic in the gitlint command to limit 65 chars in the original commit title. I'll edit the commit title for now
> 
> > 🤷‍♂️ you tell me. If you think it will break things, can you move it to draft? If you think it is a good change, we can merge it. It will need to be followed by a catalog update which then has to go through the promotion process
> 
> yea, sorry. That's on me to check closer. I checked the logic and walked through the scenario with Cursor but it steered me wrong on the 709 change that broke stage so I'm overly cautious now.
> 
> This is what cursor is telling me which seems to check out. It seems fine to merge prior to the finalizer change landing in internal-services
> 
> ```
> PR 734 deletes the orphaned IR via kubectl delete --wait=true, but without the finalizer from #709, the internal-services controller has nothing to hook into — the IR is just gone, and the associated internal PipelineRun continues running to natural completion. So duplicate concurrent runs can still occur until #709 lands.
> 
> This is not a regression: without 734, the orphaned IR and its PipelineRun would also keep running — there was never any cancellation mechanism in the current codebase. The kubectl delete here is intentionally setting up the hook point: once #709 is re-implemented, the controller will intercept that deletion via the finalizer and cancel the internal PipelineRun before the new IR is created.
> ```

good catch about the gitlint thing. how annoying...

swickersh:
Merge ci failed due to python docstring requirements that were added.

johnbieren (approved):

PR #738, opened 2026-04-24
refactor: convert check-embargoed-cves to python

johnbieren:
This commit refactors the check-embargoed-cves bash script into python. It also adds some helper modules that other tasks will use as they are converted.

Assisted-By: Cursor

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,8 @@ description = "Konflux Release Service Utils"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
+    "requests",
+    "requests-kerberos",
     "jinja2",
     "check-jsonschema",
     "jinja2-ansible-filters",
@@ -25,7 +27,11 @@ dependencies = [
 line-length = 95
 
 [tool.pytest.ini_options]
-pythonpath = ["integration-tests/lib"]
+pythonpath = [
+  "integration-tests/lib",
+  "scripts/python/helpers",
+  "scripts/python/tasks/internal",
+]
 
 [dependency-groups]
 dev = [
@@ -40,7 +46,7 @@ omit = [
     "**/tests/*",
     ".github/*",
     "ci/*",
-    "scripts/*",
+    "scripts/bash/*",
     "templates/*",
     "data/*",
 ]
--- a/scripts/python/helpers/file.py
+++ b/scripts/python/helpers/file.py
@@ -1,0 +1,56 @@
+"""File, path, and temporary-file helpers for task scripts."""
+
+from __future__ import annotations
+
+import os
+import tempfile
+from pathlib import Path
+
+
+def path_from_env_variable(
+    name: str,
+    default: str | Path,
+) -> Path:
+    """
+    Return a filesystem path from the string value of an environment variable, or
+    a default.
+
+    The value of name in ``os.environ`` (if set and not blank after
+    ``str.strip``) is interpreted as a path; it is not a path to a file whose
+    contents you read, and this function does not open or stat paths.
+
+    If the variable is missing or only whitespace, default is returned (a str
+    or an existing ``Path``).
+
+    Typical use: a Tekton or pod env var that holds a mount directory path, with
+    tests setting the same variable to a temp directory. Existence of the path
+    is not checked.
+    """
+    raw = os.environ.get(name)
+    if raw is not None and str(raw).strip() != "":
+        return Path(str(raw).strip())
+    return default if isinstance(default, Path) else Path(default)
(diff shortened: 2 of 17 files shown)


johnbieren:
I tested this by building an image in my own namespace and using a draft catalog PR https://github.com/konflux-ci/release-service-catalog/pull/2179

scripts/python/helpers/file.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    raw = os.environ.get(name)
+    if raw is not None and str(raw).strip() != "":
+        return Path(str(raw).strip())
+    return default if isinstance(default, Path) else Path(default)
+
+
+def make_tempfile_path(
+    prefix: str,
+    data: Optional[bytes] = None,
```
mmalina: Maybe use `data: bytes | None = None` to be consistent with other places?

scripts/python/tasks/internal/check_embargoed_cves.py
```diff
@@ -0,0 +1,303 @@
...
+            if is_embargoed_flaw_response(data):
+                print(f"CVE {cve} is embargoed", flush=True)
+                found.append(cve)
+
+        # Return code 0/1 is logical outcome only; main() maps it to the Tekton result files.
+        return (found, 0 if not found else 1)
+    finally:
+        # kpath, ccache, kcfg: remove all temp files even when kinit or a later step raises.
+        for p in (kpath, ccache_path, kcfg):
+            p.unlink(missing_ok=True)  # type: ignore[call-arg]
```
mmalina: Claude says the comment is unnecessary (and you don't have it elsewhere anyway).

> The missing_ok parameter has been available since Python 3.8, and your minimum is
   3.10.

scripts/python/helpers/http_client.py
```diff
@@ -0,0 +1,68 @@
...
+def _retries() -> Retry:
+    """Transients similar to common ``curl --retry 3`` (connection + some HTTP)."""
+    return Retry(
+        total=3,
+        connect=3,
+        read=3,
+        status=2,
+        backoff_factor=0.4,
+        status_forcelist=(500, 502, 503, 504),
+        allowed_methods=frozenset({"GET", "POST"}),
```
mmalina: Currently there is only a function for performing GET. Why the POST here? Maybe a function for post is coming? In general it's risky to retry a POST request, so it depends on the case. But since there is no POST here yet, maybe remove it?

scripts/python/helpers/http_client.py
```diff
@@ -0,0 +1,68 @@
+"""HTTP helpers for task code using the third-party ``requests`` library.
+
+The catalog often shows shell with ``curl --retry 3`` and Kerberos or bearer
```
mmalina, line 3: What about covering rate limiting and the optional 404 retry that is included in the curl-with-retry bash script? Or will that be added later when tasks using it are migrated?

scripts/python/helpers/file.py
johnbieren: updated

scripts/python/tasks/internal/check_embargoed_cves.py
johnbieren: removed

scripts/python/helpers/http_client.py
johnbieren: removed

scripts/python/helpers/http_client.py
johnbieren: I added the functionality in your open PR for the 404 retry stuff. I also added retries to the kinit to cover what Lubo did in this commit https://github.com/konflux-ci/release-service-utils/commit/4d1661211b01402ccb9655a7440f0229a41ade2c

johnbieren:
Addressed comments and added coverage for the lines codecov complained about

scripts/python/helpers/http_client.py
mmalina: cool, thanks

mmalina (approved):

ach912 (approved):

davidmogar (approved):

pyproject.toml
```diff
@@ -5,6 +5,8 @@ description = "Konflux Release Service Utils"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
+    "requests",
+    "requests-kerberos",
```
ktdreyer, line 9: Check out https://pypi.org/project/requests-gssapi/ and https://fedoraproject.org/wiki/Changes/kerberos-in-python-modernization!

PR #739, opened 2026-04-24
fix(RELEASE-2367): make push-disk-image pipeline idempotent

swickersh:
This PR delivers two related idempotency changes for the push-disk-image pipeline.

Pulp idempotency prune (RELEASE-2367)
pulp_push_wrapper now queries the target Pulp repo and unassociates older timestamped variants of matching staged files before running pubtools-pulp-push. This prevents duplicate timestamped content from accumulating and keeps only the latest relevant artifacts associated. Unit tests were added for timestamp matching and prune selection behavior.

Shared CGW idempotency logic
Common Content Gateway create/update/skip/rollback behavior was extracted into utils/cgw_idempotency.py and reused by both developer_portal_wrapper and publish_to_cgw_wrapper. This removes duplicated logic and keeps both publish paths behaviorally aligned. Regression tests cover both flows:

disk-image style (filePrefix-based)
push-artifacts style (files[]-based)

### DIFF
--- a/developer-portal-wrapper/developer_portal_wrapper.py
+++ b/developer-portal-wrapper/developer_portal_wrapper.py
@@ -1,20 +1,16 @@
 #!/usr/bin/env python3
-"""
-Run push-cgw-metadata command to push metadata to CGW:
-1. Extract all components under contentGateway key from dataPath
-2. Find all the files in contentDir that starts with the component name
-4. Generate necessary metadata for each file
-5. Dump the metadata to a YAML file
-6. Run push-cgw-metadata to push the metadata
-"""
+"""Publish content to CGW using idempotent create/update behavior."""
 
 import os
 import sys
 import yaml
 import hashlib
-import subprocess
 import argparse
 import logging
+
+import requests
+from requests.auth import HTTPBasicAuth
+from utils import cgw_idempotency
 
 LOG = logging.getLogger("developer-portal-wrapper")
 DEFAULT_LOG_FMT = "%(asctime)s [%(levelname)-8s] %(message)s"
@@ -26,7 +22,6 @@ CGW_ENV_VARS_STRICT = (
 
 WORKSPACE_DIR = "/tmp"
 METADATA_FILE_PATH = f"{WORKSPACE_DIR}/cgw_metadata.yaml"
-RESULT_FILE_JSON_PATH = f"{WORKSPACE_DIR}/results.json"
 
 os.makedirs(WORKSPACE_DIR, exist_ok=True)
 
@@ -53,44 +48,40 @@ def generate_metadata(
     content_list that starts with the component name
     """
     metadata = []
-
     short_url_prefix = f"/cgw/{product_code}/{product_version_name}"
-    for file in content_list:
-        matching_component = file.startswith(file_prefix)
 
-        if matching_component:
-            LOG.info("processing file: %s", file)
-            component_item = {
-                "productName": product_name,
-                "productCode": product_code,
-                "productVersionName": product_version_name,
-                "downloadURL": generate_download_url(content_dir, file),
+    for order, file in enumerate(content_list, start=1):
+        if not file.startswith(file_prefix):
+            LOG.warning(
+                "Skipping file: %s as it does not start with the expected prefix", file
+            )
+            continue
+
+        LOG.info("Processing file: %s", file)
+        metadata.append(
+            {
                 "shortURL": f"{short_url_prefix}/{file}",
+                "downloadURL": generate_download_url(content_dir, file),
                 "label": file,
                 "type": "FILE",
                 "hidden": False,
                 "invisible": False,
+                "order": order,
             }
-            metadata.append(
-                {
-                    "type": "file",
-                    "action": "create",
-                    "metadata": component_item,
-                }
-            )
-        else:
-            LOG.warning(f"Skipping file: {file} as it does not start with any component name")
+        )
 
     return metadata
 
 
 def validate_env_vars():
+    """Assert that all required CGW environment variables are set."""
     assert all(
         [os.getenv(item) for item in CGW_ENV_VARS_STRICT]
     ), f"Provide all required CGW environment variables: {', '.join(CGW_ENV_VARS_STRICT)}"
 
 
 def parse_args():
+    """Parse and return command line arguments."""
     parser = argparse.ArgumentParser(
         prog="developer_portal_wrapper",
         description="Add binaries to Developer Portal.",
@@ -122,7 +113,10 @@ def parse_args():
     )
     parser.add_argument(
         "--cgw-hostname",
-        help="Content Gateway Hostname in Developer Portal",
+        help=(
+            "Content Gateway base URL in Developer Portal, "
+            "for example: https://developers.qa.redhat.com/content-gateway/rest/admin"
+        ),
         required=True,
     )
     parser.add_argument(
@@ -139,7 +133,73 @@ def parse_args():
     return parser.parse_args()
 
 
+def publish_metadata(
+    *,
+    cgw_hostname,
+    product_name,
+    product_code,
+    product_version_name,
+    metadata,
+    dry_run=False,
+):
+    """Idempotently publish file metadata to the Content Gateway.
+
+    Looks up the product and version by name, then creates, updates, or skips
+    each file entry depending on whether it already exists with matching content.
+    In dry-run mode no API calls are made; placeholder IDs are used instead.
+    """
+    username = os.getenv("CGW_USERNAME")
+    password = os.getenv("CGW_PASSWORD")
+
+    session = requests.Session()
+    session.auth = HTTPBasicAuth(username, password)
+    session.headers.update(
+        {
+            "Accept": "application/json",
+            "Content-Type": "application/json",
+        }
+    )
+
+    if dry_run:
+        product_id = 999999
+        product_version_id = 999999
+        created_ids = [999999 for _ in metadata]
+        updated_ids = []
+        skipped_ids = []
+    else:
+        product_id = cgw_idempotency.get_product_id(
+            host=cgw_hostname,
+            session=session,
+            product_name=product_name,
+            product_code=product_code,
+        )
+        product_version_id = cgw_idempotency.get_version_id(
+            host=cgw_hostname,
+            session=session,
+            product_id=product_id,
+            version_name=product_version_name,
+        )
+        for item in metadata:
+            item["productVersionId"] = product_version_id
+
+        created_ids, updated_ids, skipped_ids = cgw_idempotency.create_files(
+            host=cgw_hostname,
+            session=session,
+            product_id=product_id,
+            version_id=product_version_id,
+            metadata=metadata,
+        )
+
+    LOG.info(
+        "CGW publish summary: created=%s updated=%s skipped=%s",
+        len(created_ids),
+        len(updated_ids),
+        len(skipped_ids),
+    )
+
+
 def main():
+    """Entry point: parse arguments, generate metadata, and publish to CGW."""
     args = parse_args()
 
     loglevel = logging.DEBUG if args.debug else logging.INFO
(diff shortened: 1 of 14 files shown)


pyproject.toml
```diff
@@ -13,7 +13,8 @@ dependencies = [
     "pubtools-content-gateway==0.5.4",
     "pubtools-pulp==1.33.2",
     "pubtools-exodus==1.5.2",
-    "pubtools-marketplacesvm==1.9.0",
```
swickersh, line 18: Context for pinning these versions:  
- `pubtools-marketplacesvm` was previously bumped by dependency automation (`6259b44`, bot-authored) from `1.7.0` to `1.9.0`.
- In Python 3.9 runtime, `starmap-client>=2.4` triggers import/runtime failure due to newer type-hint syntax (`|`) in `starmap_client`:
  `TypeError: unsupported operand type(s) for |: 'type' and '_GenericAlias'`.
- Repro on py3.9:
  - `starmap-client 2.4.1` -> import fails
  - `starmap-client 2.3.0` (`<2.4`) -> import succeeds
 
This issue surfaced during runtime validation of this PR image (not during image build), specifically in the py3.9 `pubtools-pulp-push` execution path.
- `release-service-utils` ships a shared dependency set; the `starmap-client>=2.4` py3.9 incompatibility can be triggered even when testing non-marketplace flows.
- To unblock runtime validation of the Pulp/CGW changes, we pinned:
  - `pubtools-marketplacesvm==1.8.1`
  - `starmap-client<2.4`

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -138,6 +143,65 @@ def parse_args():
...
+            session=session,
+            product_name=product_name,
+            product_code=product_code,
+        )
+        product_version_id = cgw_idempotency.get_version_id(
+            host=cgw_hostname,
+            session=session,
+            product_id=product_id,
+            version_name=product_version_name,
+        )
```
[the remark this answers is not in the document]
swickersh: updated help text

pubtools-pulp-wrapper/pulp_push_wrapper.py
```diff
@@ -114,6 +131,178 @@ def get_source_url(stagedirs):
...
+    deadline = time.time() + POLL_TIMEOUT_SECONDS
+    while time.time() < deadline:
+        task = pulp_request(task_url, context=context)
+        state = task.get("state")
+        if state == "finished":
+            if task.get("error") or task.get("exception") or task.get("traceback"):
+                raise RuntimeError(f"Pulp task failed: {task_url}: {task}")
+            return
+        time.sleep(POLL_INTERVAL_SECONDS)
+    raise TimeoutError(f"Timed out waiting for Pulp task: {task_url}")
```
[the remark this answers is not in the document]
swickersh: fixed.

swickersh:
Added successful test run log showing idempotent workflow succeeds to the jira: https://redhat.atlassian.net/browse/RELEASE-2367?focusedCommentId=16843151

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -1,19 +1,31 @@
...
 import argparse
 import logging
+import importlib
+from pathlib import Path
+
+import requests
+from requests.auth import HTTPBasicAuth
+
+
+def _load_cgw_idempotency():
```
mmalina: This looks complex and I asked Claude if there was a better way. It suggested a couple of options, but recommended this:

```
  Replace the entire _load_cgw_idempotency() function with this at the top of the
  file:

  import sys
  from pathlib import Path

  # Add repo root to sys.path when running as a standalone script
  if __name__ == "__main__":
      repo_root = Path(__file__).resolve().parent.parent
      sys.path.insert(0, str(repo_root))

  from utils import cgw_idempotency

  Why this is better:
  - Only modifies sys.path when running as a script (not when imported)
  - Uses standard import syntax instead of importlib.import_module()
  - Works in both the container (PYTHONPATH set) and locally (sys.path modified)
  - Much more readable and maintainable
```

WDYT?

pyproject.toml
mmalina: I think we should fix this properly. It turns out right now the python version is misaligned. This file has:
```
requires-python = ">=3.10"
```
But in fact the utils image only has python 3.9. Here's where it was set here in February:

https://github.com/konflux-ci/release-service-utils/commit/ef0fabec67ba2dade16ccf24481f7924163e7471

(Cc: @querti )

That seems wrong. Here's Claude's explanation:
```
  How It Actually Worked (Until It Didn't)

  Here's what happened:

  1. February 2024: Someone bumped requires-python = ">=3.10" and regenerated
  uv.lock on a machine with Python 3.10+
  2. The lock file contained specific package versions resolved for Python 3.10+
  3. In the container (Python 3.9):
    - uv pip install reads uv.lock and installs those exact versions
    - Most packages work fine because they don't actually use Python 3.10+ features
   - they just allow Python 3.10+
    - Everything seemed to work... until...
  4. Renovate updated starmap-client to 2.4+, which actually uses Python 3.10+
  syntax (type | type union syntax)
  5. Runtime failure: Python 3.9 can't parse the | operator for types → TypeError

  TL;DR: It worked by accident because most packages claiming "Python 3.10+" still
  had Python 3.9-compatible code. It broke when a dependency actually started using
   Python 3.10+ syntax.
```

So what we should do is move to a base image that has newer python - e.g. https://catalog.redhat.com/en/software/containers/ubi9/python-312/657b08d023df896ebfacf402
(and remove the python install in the dockerfile). That way we can properly use the newest packages.

```
  What to Do

  1. Change the Dockerfile base image from ubi9/ubi:9.7-* to ubi9/python-312
  2. Remove the Python installation steps (already included in the image)
  3. Remove the workaround pins for pubtools-marketplacesvm and starmap-client
  4. Regenerate uv.lock with Python 3.12 to ensure all dependencies are properly
  resolved

  This will fix the contradiction and prevent future "it works but shouldn't"
  scenarios.
```

pubtools-pulp-wrapper/pulp_push_wrapper.py
```diff
@@ -179,6 +373,7 @@ def main():
         LOG.info("Would have run: %s", cmd_str)
     else:
         try:
+            prune_matching_content_before_push(args)
```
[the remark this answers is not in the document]
swickersh: If a team is re-releasing anyway the existing content should be removed regardless.

developer-portal-wrapper/developer_portal_wrapper.py
swickersh: I actually had what Claude proposed in a previous diff: https://github.com/konflux-ci/release-service-utils/compare/8e7acfb0a2dfd15b67e0df2a88fae22a5134a415..3f9056e4acb4d4615367ae6fa6f362922ffef644

Battle of the ai's here :smile:  : 


CI flake8 E402

The old pattern had executable code (ROOT_DIR = ..., sys.path.insert(...)) before from utils....
That triggered E402 module level import not at top of file.
We explicitly changed it to avoid # noqa and fix it “properly.”
Import-context reliability

The suggested if __name__ == "__main__": ... from utils import cgw_idempotency only adjusts sys.path in one execution mode.
In some non-__main__ contexts (imported module, certain test/runtime invocations), it can still miss utils if PYTHONPATH is not already set.
So the current _load_cgw_idempotency() approach is a bit less pretty, but it was chosen to be:

lint-compliant,
fallback-safe,
and consistent across wrapper execution contexts.


I'm happy to change it and suppress the errors with that annotation if you prefer. It works both ways in testing :man_shrugging:

pyproject.toml
swickersh: Created a Jira for this as it's outside the scope of this PR. https://redhat.atlassian.net/browse/RELEASE-2442

developer-portal-wrapper/developer_portal_wrapper.py
mmalina: OK, fine with me then 😀

pyproject.toml
mmalina: Well, then at least change requires-python to 3.9 to reflect reality? And that way you will likely not have to add "starmap-client<2.4" either, because it will just get the version compatible with python 3.9.

swickersh: As I understand it, changing requires-python to >=3.9 would require regenerating uv.lock, which risks pulling in unintended package changes.  And it wouldn't actually remove the need for the starmap-client<2.4 pin anyway: starmap-client doesn't declare requires_python in its PyPI metadata, so uv has no way to know that 2.4+ is incompatible with Python 3.9 and would still resolve it. 

It feels like proper process is to create bug Jira (RELEASE-2442) and fix it as a separate commit/PR.

mmalina: Hmm, but I think you should definitely update uv.lock actually when changing this file, it's actually good you mention it. Don't you think? I will ask others on the team to take a look - maybe somebody has more python knowledge...

swickersh: Yea you're right. I think it would have failed somewhere if I didn't update the lock file after updating the toml file. I regenerated uv lock and tests are running now.

mmalina: ok. I am fine with dealing with the rest in the separate jiras - the one to use ubi10 is probably the best to start with. feel free to resolve this once you're done.

mmalina (dismissed):

happybhati (dismissed):

developer-portal-wrapper/developer_portal_wrapper.py
```diff
@@ -138,6 +146,65 @@ def parse_args():
     return parser.parse_args()
 
 
+def publish_metadata(
```
davidmogar, line 136: Can we add docstrings for all the functions?

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
```diff
@@ -25,9 +25,26 @@
...
 import hashlib
 import logging
+import importlib
+import sys
+from pathlib import Path
 import requests
 from requests.auth import HTTPBasicAuth
 
+
+def _load_cgw_idempotency():
```
davidmogar: Would it make sense to extract this function as it's used in other places?

swickersh: Now that I look at it more - it seems that was only needed for running the wrappers standalone locally. When running as part of the test suite or in production the imports work without this. I made the change to remove it and will let ci build the container to confirm tests still pass.

developer-portal-wrapper/developer_portal_wrapper.py
swickersh: yes, done.

publish-to-cgw-wrapper/publish_to_cgw_wrapper.py
swickersh: Ok - looks like the imports work fine without it. I ran a test on cluster.

davidmogar (dismissed):

utils/cgw_idempotency.py
```diff
@@ -0,0 +1,230 @@
+import logging
+import re
+from urllib.parse import urlsplit, urlunsplit
+
+import requests
+
+TIMESTAMP_TOKEN_RE = re.compile(r"-(\d{10})(?=-)")
+
+
+def call_cgw_api(*, host, method, endpoint, session, data=None):
```
johnbieren, line 10: This is fine for now, but nothing about this seems cgw specific. Pyxis makes http calls too - we should probably combine in the future

utils/cgw_idempotency.py
```diff
@@ -0,0 +1,230 @@
...
+def rollback_files(*, host, session, product_id, version_id, created_file_ids):
+    """Rollback created files by listing and deleting them."""
+    if created_file_ids:
+        logging.warning(
+            "Rolling back created files due to failure (productId: %s, productVersionId: %s)",
+            product_id,
+            version_id,
+        )
+
+    for file_id in created_file_ids:
```
johnbieren, line 123: I think it'd be nice to log something like "deleting file $file_id"

johnbieren (dismissed):

utils/cgw_idempotency.py
swickersh: We can if you feel strongly about it.  This is existing code that is just moved to a common script so both the developer_portal_wrapper and publish_to_cgw_wrapper can be idempotent and share these functions without duplicate code
https://github.com/konflux-ci/release-service-utils/pull/739/changes/BASE..f9cd0ecbd59ebe4c19c5573d3ce1be506864ee87#diff-211a5181f683316b7ae07ea11f6a624d998ff6e9fd61cddd9368d6ec8757d078L285

utils/cgw_idempotency.py
swickersh: same as below comment. This is existing code just moved into a common file: https://github.com/konflux-ci/release-service-utils/pull/739/changes/BASE..f9cd0ecbd59ebe4c19c5573d3ce1be506864ee87#diff-211a5181f683316b7ae07ea11f6a624d998ff6e9fd61cddd9368d6ec8757d078L123

ach912 (dismissed):

pubtools-pulp-wrapper/pulp_push_wrapper.py
```diff
@@ -115,6 +131,180 @@ def get_source_url(stagedirs):
...
+
+def wait_for_task(task_href, context):
+    task_url = task_href
+    deadline = time.time() + POLL_TIMEOUT_SECONDS
+    while time.time() < deadline:
+        task = pulp_request(task_url, context=context)
+        if task is None:
+            raise RuntimeError(f"Empty response while polling Pulp task status: {task_url}")
+        state = task.get("state")
+        if state == "finished":
```
jangel97, line 224: `wait_for_task` only checks for `state == "finished"`, so if a Pulp task reaches
`"error"` or `"canceled"`, it will poll for the full 120s and then raise a `TimeoutError` instead
of surfacing the actual Pulp error.

According to the [Pulp 2 docs](https://docs.pulpproject.org/en/2.21/dev-guide/integration/rest-api/tasks.html),
the possible task states are: `waiting`, `skipped`, `running`, `suspended`, `finished`, `error`, and `canceled`.

Would something like this help?

```python
if state == "finished":
    return
if state in ("error", "canceled", "skipped"):
    raise RuntimeError(f"Pulp task {state}: {task_url}: {task}")
```
WDYT @swickersh ?

swickersh: Yea that's a good call out. Thanks @jangel97 
Implemented.

davidmogar (dismissed):

johnbieren (dismissed):

swickersh:
rebased - need approvals again please @davidmogar @johnbieren

johnbieren (dismissed):

happybhati (dismissed):

mmalina (dismissed):

davidmogar (dismissed):

swickersh:
rebased on https://github.com/konflux-ci/release-service-utils/pull/770 now that ubi10 and newer python are merged.

johnbieren (approved):

davidmogar (approved):

ach912 (approved):

twaugh (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-05-28T11:18:16.203Z |

Approved

happybhati (approved):

PR #745, opened 2026-04-29
feat(RELEASE-2209): imported functions trigger tests

johnbieren:
This commit updates the new utils e2e tests to prepare for the addition of python task scripts to this repo. We will have helper modules that multiple tasks can import. If a task script uses a function from a helper module, we need to be sure that the e2e tests relevant to that task are still triggered. This commit adds that piece. So, for example, if you have a function in scripts/tasks/helpers/file.py and it is used in scripts/tasks/internal/check_embargoed_cves.py, any update to the function in file.py should be treated like an update in the embargoed script too.

Assisted-By: Cursor

### DIFF
--- a/integration-tests/lib/helper_task_import_graph.py
+++ b/integration-tests/lib/helper_task_import_graph.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python3
+"""Map ``scripts/python/helpers/*.py`` to task scripts that import them.
+
+Task scripts live under ``scripts/python/tasks/**/*.py``. Imports are resolved with
+the ``ast`` module: bare imports like ``import file`` match ``helpers/file.py``;
+``from helpers.foo import ...`` is also recognized when present.
+
+Used by `find_catalog_suite_from_utils_diff` so a helper-only diff still
+propagates to dependent ``tasks/`` paths for Dockerfile token generation.
+"""
+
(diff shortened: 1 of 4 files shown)


johnbieren:
No user facing code (ever), no spreadsheet reporting needed

integration-tests/lib/helper_task_import_graph.py
```diff
@@ -0,0 +1,128 @@
...
+            continue
+        try:
+            text = py.read_text(encoding="utf-8", errors="replace")
+            tree = ast.parse(text, filename=str(py))
+        except (OSError, SyntaxError):
+            # Broken files cannot be analyzed; omit rather than failing the pipeline.
+            continue
+        rel = py.relative_to(repo_root).as_posix()
+        for stem in _collect_imported_helper_names(tree, stems):
+            reverse.setdefault(stem, set()).add(rel)
```
mmalina: Claude says that the setdefault is redundant here? Because everything is already initialized on line 70. So this could be changed to just `reverse[stem].add(rel)`. But it's a detail - it doesn't harm anything to have that.

johnbieren: fixed

mmalina (approved):

PR #765, opened 2026-05-12
feat(RELEASE-1984): add check-fbc-opt-in python script

johnbieren:
This commit adds a python script to replicate the functionality of the inline bash script in the check-fbc-opt-in internal task in the catalog repo. The task will be updated to use this python module instead.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/image_ref.py
+++ b/scripts/python/helpers/image_ref.py
@@ -1,0 +1,25 @@
+"""Helpers for parsing and normalizing container image references."""
+
+from __future__ import annotations
+
+
+def pyxis_url_for_pull_spec(pyxis_url: str, pull_spec: str) -> str:
+    """
+    Build the Pyxis repository/tag API URL for `pull_spec`.
+
+    `registry.redhat.io` is rewritten to `registry.access.redhat.com` to
+    match Pyxis lookups.
+    """
+    normalized = pull_spec.replace("registry.redhat.io", "registry.access.redhat.com", 1)
+    parts = normalized.split("/", 2)
+    if len(parts) < 3:
+        raise ValueError(f"invalid pull spec: {pull_spec!r}")
+    registry, repo, image_and_tag = parts
+    image, sep, tag = image_and_tag.partition(":")
+    base = (
+        f"{pyxis_url.rstrip('/')}/repositories/registry/{registry}/repository/"
+        f"{repo}/{image}"
+    )
+    if sep and tag:
+        return f"{base}/tag/{tag}"
+    return base
--- a/scripts/python/helpers/logger.py
+++ b/scripts/python/helpers/logger.py
@@ -1,0 +1,15 @@
+"""Configured logger for task scripts — writes to stderr."""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger("release")
+logger.setLevel(logging.DEBUG)
+
+# Clear any existing handlers first
+logger.handlers.clear()
+_handler = logging.StreamHandler()
+_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
+logger.addHandler(_handler)
+logger.propagate = False
--- a/scripts/python/helpers/test_image_ref.py
+++ b/scripts/python/helpers/test_image_ref.py
@@ -1,0 +1,30 @@
+"""Tests for the `image_ref` helper module."""
+
+from __future__ import annotations
+
+import image_ref
+import pytest
+
+
+def test_pyxis_url_for_pull_spec_with_tag_and_registry_rewrite() -> None:
+    """Tagged refs map to `.../tag/<tag>` and rewrite registry.redhat.io host."""
+    out = image_ref.pyxis_url_for_pull_spec(
+        "https://pyxis.engineering.redhat.com/v1",
+        "registry.redhat.io/repo/image:1.2",
+    )
+    assert out.endswith(
+        "/repositories/registry/registry.access.redhat.com/repository/repo/image/tag/1.2"
+    )
+
+
+def test_pyxis_url_for_pull_spec_without_tag() -> None:
+    """Untyped pull specs omit the trailing `/tag` path segment."""
+    out = image_ref.pyxis_url_for_pull_spec("https://pyxis/v1", "r.io/repo/image")
+    assert out.endswith("/repositories/registry/r.io/repository/repo/image")
+    assert "/tag/" not in out
+
+
+def test_pyxis_url_for_pull_spec_invalid() -> None:
+    """Invalid pull specs raise `ValueError`."""
+    with pytest.raises(ValueError, match="invalid pull spec"):
+        image_ref.pyxis_url_for_pull_spec("https://pyxis/v1", "not/a-pullspec")
--- a/scripts/python/tasks/internal/check_fbc_opt_in.py
+++ b/scripts/python/tasks/internal/check_fbc_opt_in.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+"""Check FBC opt-in status in Pyxis for container images."""
+
+from __future__ import annotations
+
+import base64
+import binascii
+import json
+import os
+import subprocess
+import tempfile
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any
+
+import requests
+from requests.auth import AuthBase
+from requests_kerberos import HTTPKerberosAuth, OPTIONAL
+
+import authentication
+import file
+import http_client
+import image_ref
+import tekton
+from logger import logger
+
+PROG = "check_fbc_opt_in.py"
+
+
+def parse_container_images(value: str) -> list[str]:
+    """Parse `value` as a JSON array of non-empty image strings."""
+    # Input is a JSON array string; keep validation strict so malformed input fails fast.
+    data = json.loads(value)
+    if not isinstance(data, list):
+        raise ValueError("container images must be a JSON array")
+    out: list[str] = []
+    for item in data:
+        if not isinstance(item, str):
+            raise ValueError("container images must be strings")
+        stripped = item.strip()
+        if not stripped:
+            raise ValueError("container images must be non-empty")
+        out.append(stripped)
+    return out
+
+
+def get_fbc_opt_in(
+    pyxis_url: str,
+    pull_spec: str,
+    auth: AuthBase | None,
+    *,
+    warn_on_query_failure: bool = True,
+) -> bool:
+    """
+    Return `True` only when the Pyxis JSON body sets `fbc_opt_in` to JSON boolean
+    true.
+
+    Query failures, non-JSON responses, and missing fields are treated as
+    opt-out (`False`).
+    """
+    try:
+        # Build the Pyxis endpoint for this image pull spec.
+        body = http_client.get_text(
+            image_ref.pyxis_url_for_pull_spec(pyxis_url, pull_spec),
+            auth=auth,
+        )
+        data: dict[str, Any] = json.loads(body)
+    except (
+        OSError,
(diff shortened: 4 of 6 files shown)


scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+
+def parse_container_images(value: str) -> list[str]:
+    """Parse ``value`` as a JSON array of non-empty image strings."""
+    # Task param is a JSON string; keep validation strict so malformed input fails fast.
+    data = json.loads(value)
+    if not isinstance(data, list):
+        raise ValueError("container images must be a JSON array")
+    out: list[str] = []
+    for item in data:
+        if not isinstance(item, str) or not item.strip():
```
FilipNikolovski: nit - we do item.strip() twice here, we can capture it in a variable.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+            raise ValueError("container images must be non-empty strings")
+        out.append(item.strip())
+    return out
+
+
+def get_fbc_opt_in(
+    pyxis_url: str,
+    pull_spec: str,
+    auth: Any,
```
FilipNikolovski: Can we add a more specific type for auth instead of using Any?

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+        # requests-kerberos reads Kerberos env from this process, not the kinit child.
+        os.environ.update(kenv)
+
+        # Reuse one Kerberos auth object while querying each image's Pyxis record.
+        auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL)
+        logger._info("Checking FBC opt-in status for provided container images...")
+        opt_in_rows: list[dict[str, Any]] = []
+        for image in container_images:
+            print(f"INFO: Checking opt-in status for: {image}", flush=True)
+            opted_in = bool(get_opt_in(pyxis_url, image, auth))
```
FilipNikolovski: No need to cast this to bool since get_opt_in returns a boolean.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+            raise tekton.CheckStepError("logging in with Kerberos (kinit)", e) from e
+        # requests-kerberos reads Kerberos env from this process, not the kinit child.
+        os.environ.update(kenv)
+
+        # Reuse one Kerberos auth object while querying each image's Pyxis record.
+        auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL)
+        logger._info("Checking FBC opt-in status for provided container images...")
+        opt_in_rows: list[dict[str, Any]] = []
+        for image in container_images:
+            print(f"INFO: Checking opt-in status for: {image}", flush=True)
```
FilipNikolovski: We should replace print call with logger._info.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+            opted_in = bool(get_opt_in(pyxis_url, image, auth))
+            print(
```
FilipNikolovski: We should replace print call with logger._info.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def run_check(
+    container_images: list[str],
+    pyxis_url: str,
+    service_account_mount: Path,
+    iib_services_config_mount: Path,
+    *,
+    kinit: Any = authentication.kinit_with_retry,
+    get_opt_in: Any = get_fbc_opt_in,
```
FilipNikolovski: We could use the Callable type here for kinit and get_opt_in:
```suggestion
    kinit: Callable[[str, Path, dict[str, str]], None] = authentication.kinit_with_retry,
    get_opt_in: Callable[[str, str, Any], bool] = get_fbc_opt_in,
```

scripts/python/helpers/logger.py
```diff
@@ -0,0 +1,13 @@
+"""Prefixed INFO and WARNING lines on stderr for task scripts."""
```
FilipNikolovski: I'd suggest to use the logging library here instead of plain print, it's more extensible and we can be consistent with our message format etc:
```python
  """Configured logger for task scripts — writes to stderr."""

  from __future__ import annotations

  import logging

  logger = logging.getLogger("release")
  logger.setLevel(logging.DEBUG)

  _handler = logging.StreamHandler()  # stderr by default
  _handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
  logger.addHandler(_handler)
```

then we can log messages like so:
```python
  from logger import logger

  logger.info("Setting up Kerberos authentication for Pyxis...")
  logger.warning("Failed to query Pyxis...")
  logger.error("Failed to do something...")
```

scripts/python/helpers/logger.py
```diff
@@ -0,0 +1,13 @@
...
+
+from __future__ import annotations
+
+import sys
+
+
+def _info(message: str) -> None:
+    print(f"INFO: {message}", file=sys.stderr, flush=True)
```
FilipNikolovski: Why do info messages go to stderr?

johnbieren: No longer relevant due to the logger change

scripts/python/helpers/logger.py
johnbieren: changed

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: updated

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: done

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: done

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: done

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: yep, done

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: it was because we can't call `.strip()` until we know it is a string, but I modified this and improved the error message

FilipNikolovski (dismissed):
LGTM

davidmogar (dismissed):

scripts/python/helpers/image_ref.py
```diff
@@ -0,0 +1,25 @@
...
+    if len(parts) < 3:
+        raise ValueError(f"invalid pull spec: {pull_spec!r}")
+    registry, repo, image_and_tag = parts
+    image, sep, tag = image_and_tag.partition(":")
+    base = (
+        f"{pyxis_url.rstrip('/')}/repositories/registry/{registry}/repository/"
+        f"{repo}/{image}"
+    )
+    if sep and tag:
+        return f"{base}/tag/{tag}"
```
theflockers, line 24: You can completely remove parts regarding the tag as for https://redhat.atlassian.net/browse/KONFLUX-13739. It never will return the opt-in information.

scripts/python/helpers/test_image_ref.py
```diff
@@ -0,0 +1,30 @@
+"""Tests for the ``image_ref`` helper module."""
+
+from __future__ import annotations
+
+import image_ref
+import pytest
+
+
+def test_pyxis_url_for_pull_spec_with_tag_and_registry_rewrite() -> None:
```
theflockers, line 9: Same as above and for the next subsequent tests where `tag` is used.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,199 @@
...
+            service_account_mount,
+            iib_services_config_mount,
+        )
+    except (ValueError, tekton.CheckStepError) as e:
+        # Task failures should surface as a clear one-line SystemExit message.
+        raise SystemExit(f"{PROG}: {e}") from e
+
+    logger.info("FBC opt-in check completed")
+    logger.info("Results:")
+    print(json.dumps(results, indent=2), flush=True)
```
theflockers: Not sure, but it seems you sent "Results"  to logger, but only printed out the results, instead of sending the results to logger as well.

scripts/python/helpers/logger.py
```diff
@@ -0,0 +1,14 @@
+"""Configured logger for task scripts — writes to stderr."""
+
+from __future__ import annotations
+
+import logging
+
+logger = logging.getLogger("release")
+logger.setLevel(logging.DEBUG)
+
+if not logger.handlers:
```
theflockers: From `Claude`:
```
Problem: The guard if not logger.handlers: is good, but it's checking before adding.
If this module is reloaded or imported multiple times in edge cases, handlers  could accumulate.
```
and suggests:
```
# Clear any existing handlers first
logger.handlers.clear()
_handler = logging.StreamHandler()
_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
logger.addHandler(_handler)
logger.propagate = False
```

Not sure we would have a similar situation, as each script would import logger and memory would clean up after the pod is dead, but it worth it taking a look.

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -0,0 +1,199 @@
...
+            raise tekton.CheckStepError("logging in with Kerberos (kinit)", e) from e
+        # requests-kerberos reads Kerberos env from this process, not the kinit child.
+        os.environ.update(kenv)
+
+        # Reuse one Kerberos auth object while querying each image's Pyxis record.
+        auth = HTTPKerberosAuth(mutual_authentication=OPTIONAL)
+        logger.info("Checking FBC opt-in status for provided container images...")
+        opt_in_rows: list[dict[str, Any]] = []
+        for image in container_images:
+            logger.info("Checking opt-in status for: %s", image)
```
theflockers: It could be helpful logging also the pyxis URL being called, for debugging purposes.

scripts/python/helpers/image_ref.py
johnbieren: It looks like that ticket is in to do - I'd rather keep functional parity with this conversion and let the tag part be addressed as part of that jira

theflockers: as you wish!

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: nice catch, fixed

scripts/python/helpers/logger.py
johnbieren: sure, went for it

scripts/python/tasks/internal/check_fbc_opt_in.py
johnbieren: added

ach912 (approved):

theflockers (approved):

davidmogar (approved):

twaugh (approved):

PR #767, opened 2026-05-13
feat(RELEASE-2314): Support PIPELINE_TEST_SUITE_VARS in ITS

midnightercz:
Add support for PIPELINE_TEST_SUITE_VARS for ITS test suites in release-service-catalog

### DIFF
--- a/integration-tests/lib/run_single_catalog_e2e_suite.py
+++ b/integration-tests/lib/run_single_catalog_e2e_suite.py
@@ -19,6 +19,9 @@ Required env:
   ``$(context.pipelineRun.uid)``); child PLR name is ``utils-e2e-catalog-<uid>``
   (same suffix as the temp GitHub fork).
 
+Optional env:
+  PIPELINE_TEST_SUITE_VARS
+
 ``KUBECONFIG_SECRET_NAME`` is the Secret **name** passed to the child catalog
 ``PipelineRun`` as pipeline param ``KUBECONFIG_SECRET_NAME`` (stage/test cluster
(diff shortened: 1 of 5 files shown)


integration-tests/lib/run_single_catalog_e2e_suite.py
```diff
@@ -19,6 +19,9 @@
   ``$(context.pipelineRun.uid)``); child PLR name is ``utils-e2e-catalog-<uid>``
   (same suffix as the temp GitHub fork).
 
+Optional env:
+  PIPELINE_TEST_SUITE
```
elenagerman: typo?

```suggestion
  PIPELINE_TEST_SUITE_VARS
```

elenagerman:
probably need to be mentioned in integration-tests/README.md

midnightercz:
>probably need to be mentioned in integration-tests/README.md

done

elenagerman (dismissed):

johnbieren:
This isn't user facing so it doesn't need the label, but I don't understand why we want this change. Can you provide context?

midnightercz:
> This isn't user facing so it doesn't need the label, but I don't understand why we want this change. Can you provide context?

Right, sorry I just automatically added it

johnbieren:
Can you update the commit to say why we want this?

midnightercz:
> Can you update the commit to say why we want this?

Updated

johnbieren:
> > Can you update the commit to say why we want this?
> 
> Updated

thanks! makes sense to me now

johnbieren (dismissed):

PR #769, opened 2026-05-14
feat(RELEASE-1991): add push-artifacts-to-cdn as standalone python sc…

swickersh:
Convert the push-artifacts-to-cdn internal task's embedded bash
steps into standalone Python scripts under
scripts/python/tasks/internal/push_artifacts_to_cdn/.

Will eventually coincide with https://github.com/konflux-ci/release-service-catalog/pull/2228
For now, this one will not be invoked by the catalog until e2e is implemented as described here:
https://redhat.atlassian.net/browse/RELEASE-2543

Unit tests achieve 98% coverage across all 8 scripts.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,6 +36,8 @@ pythonpath = [
   "scripts/python/helpers",
   "scripts/python/tasks/internal",
   "utils",
+  "pubtools-pulp-wrapper",
+  "publish-to-cgw-wrapper",
 ]
 
 [dependency-groups]
--- a/scripts/python/helpers/authentication.py
+++ b/scripts/python/helpers/authentication.py
@@ -8,8 +8,10 @@ service-account / secret file layouts.
 from __future__ import annotations
 
 import base64
+import json
 import os
 import subprocess
+import tempfile
 from collections.abc import Sequence
 from pathlib import Path
 
@@ -17,10 +19,11 @@ import retry
 
 
 def read_mounted_text(mount: Path, filename: str) -> str:
-    """
-    Read a UTF-8 file (``mount / filename``) and return the text with leading
-    and trailing whitespace removed. Use for one-line secret values (API base URL,
-    principal name) when a pod mounts a volume as a directory of small files.
+    """Read a UTF-8 file (``mount / filename``) and return stripped text.
+
+    Returns the text with leading and trailing whitespace removed.
+    Use for one-line secret values (API base URL, principal name) when
+    a pod mounts a volume as a directory of small files.
     """
     return (mount / filename).read_text(encoding="utf-8").strip()
 
@@ -31,8 +34,7 @@ def load_keytab_from_mount(
     principal_file: str,
     keytab_b64_file: str,
 ) -> tuple[str, bytes]:
-    """
-    Load a Kerberos principal and raw keytab from files under *mount*.
+    """Load a Kerberos principal and raw keytab from files under *mount*.
 
     *principal_file* is a one-line text file (principal). *keytab_b64_file*
     is the same for a base64-encoded keytab. Returns the principal string and
@@ -50,16 +52,16 @@ def load_service_account(
     principal_file: str,
     keytab_b64_file: str,
 ) -> tuple[str, bytes, dict[str, str]]:
-    """
-    Load a mounted Kubernetes-style service account: principal, keytab, and extra
-    one-value string files (API base URL, usernames, etc.).
+    """Load a mounted Kubernetes-style service account from the filesystem.
+
+    Reads the principal, keytab, and extra one-value string files
+    (API base URL, usernames, etc.) from *mount*.
 
     *principal_file* and *keytab_b64_file* are passed through to
     ``load_keytab_from_mount``. It then reads each additional filename in
-    ``text_files`` and returns a
-    dict mapping that filename to stripped text (for example an API base URL
-    in one file). You can list several files; each name becomes a key in the
-    returned dict.
+    ``text_files`` and returns a dict mapping that filename to stripped text
+    (for example an API base URL in one file). You can list several files;
+    each name becomes a key in the returned dict.
     """
     princ, keytab = load_keytab_from_mount(
         mount, principal_file=principal_file, keytab_b64_file=keytab_b64_file
@@ -69,8 +71,7 @@ def load_service_account(
 
 
 def patch_krb5_config(source: str) -> str:
-    """
-    Return ``source`` with one line added immediately after ``[libdefaults]``.
+    """Return ``source`` with one line added immediately after ``[libdefaults]``.
 
     Inserts ``dns_canonicalize_hostname = false`` to avoid
     "Invalid UID in persistent keyring name" / hostname canonicalization
@@ -92,8 +93,7 @@ def patch_krb5_config(source: str) -> str:
 def kinit_with_retry(
     princ: str, keytab: Path, extra_env: dict[str, str], *, max_attempts: int = 5
 ) -> None:
-    """
-    Run ``kinit`` with a keytab, retrying a few times on non-zero exit.
+    """Run ``kinit`` with a keytab, retrying a few times on non-zero exit.
 
     Merges ``os.environ`` with ``extra_env`` for the ``kinit`` child only.
     If every attempt fails, the last ``CalledProcessError`` is raised. **princ** is
@@ -123,3 +123,57 @@ def kinit_with_retry(
         retry_on=subprocess.CalledProcessError,
         base_sleep_seconds=5,
     )
+
+
+def write_docker_config(config_json: str) -> None:
+    """Write *config_json* to ``~/.docker/config.json``, creating the directory if needed.
+
+    Shared by helpers that authenticate to container registries (Quay, Red Hat workloads
+    registry) before pulling or pushing OCI artifacts.  The caller is responsible for
+    obtaining the JSON string (reading a mounted secret file, stripping noise, etc.).
+
+    The directory is created with mode 0700 and the file is written atomically with
+    mode 0600 to prevent other processes from reading registry credentials.
+    """
+    docker_dir = Path.home() / ".docker"
+    docker_dir.mkdir(mode=0o700, parents=True, exist_ok=True)
+    config_path = docker_dir / "config.json"
+    fd, tmp_path = tempfile.mkstemp(dir=docker_dir)
+    try:
+        with os.fdopen(fd, "w") as f:
+            f.write(config_json)
+        os.chmod(tmp_path, 0o600)
+        os.replace(tmp_path, config_path)
+    except Exception:
+        try:
+            os.unlink(tmp_path)
+        except OSError:
+            pass
+        raise
+
+
+def setup_docker_config(
+    path: Path,
+    *,
+    strip_noise: bool = False,
+    optional: bool = False,
+) -> None:
+    """Read a dockerconfigjson file and write it to ``~/.docker/config.json``.
+
+    *path* is the full path to a ``.dockerconfigjson`` file (typically a mounted
+    Kubernetes secret).  When *optional* is ``True`` and the file is absent or empty,
+    the function returns without writing anything — useful for mounts that may not be
+    present in all environments.  When *strip_noise* is ``True``, any leading/trailing
+    non-JSON characters (e.g. outer quotes added by some k8s secret encodings) are
+    stripped before writing.
+    """
+    if optional and (not path.is_file() or path.stat().st_size == 0):
+        return
+    raw = path.read_text(encoding="utf-8")
+    if strip_noise:
+        first = raw.find("{")
+        last = raw.rfind("}")
+        if first == -1 or last == -1 or first > last:
+            raise ValueError(f"No valid JSON object found in {path}")
+        raw = json.dumps(json.loads(raw[first : last + 1]))
+    write_docker_config(raw)
--- a/scripts/python/helpers/build_checksum_map.py
+++ b/scripts/python/helpers/build_checksum_map.py
@@ -1,0 +1,125 @@
+#!/usr/bin/env python3
+"""Build an OCI artifact containing a checksum map of published files.
+
+For each component's ``ready_for_distribution`` directory, computes sha256 checksums
+for all files (excluding ``sha256sum.txt*``) and assembles them into a JSON manifest.
+The manifest is packaged as a tar archive and pushed to OCI using ``oras push``.
+The resulting ``store@digest`` reference is returned by ``run()`` for the caller to record.
+The manifest is stored as an OCI artifact so that downstream advisory tooling has a stable,
+addressable pointer it can pull independently to construct PURLs for released files.
+
+Secret mounts:
+  ``TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT``  (default: ``/mnt/trusted_artifacts_dockerconfig``)
+
+Other env vars:
+  ``SNAPSHOT_JSON``        – JSON string of the Snapshot spec
+  ``CONTENT_DIR``          – override base directory (default: ``/shared/artifacts``)
+  ``SHARED_DIR``           – override shared volume root (default: ``/shared``)
+  ``OCI_STORE``            – OCI repository for checksum map artifacts
+                             (default:
+                             ``quay.io/konflux-ci/release-service-trusted-artifacts``)
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import shutil
+import tarfile
+import tempfile
+from pathlib import Path
+
+import authentication
+import file as file_utils
+import oras_utils
+
+PROG = "build_checksum_map.py"
+
+TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT = Path(
+    os.environ.get(
+        "TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT", "/mnt/trusted_artifacts_dockerconfig"
+    )
+)
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+SHARED_DIR = Path(os.environ.get("SHARED_DIR", "/shared"))
+
+OCI_STORE = os.environ.get("OCI_STORE", "quay.io/konflux-ci/release-service-trusted-artifacts")
+
+logger = logging.getLogger(__name__)
+
+
+def _setup_docker_config() -> None:
+    """Copy the mounted dockerconfig into ~/.docker/config.json if present and non-empty."""
+    authentication.setup_docker_config(
+        TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT / ".dockerconfigjson",
+        optional=True,
+    )
+    logger.info("Docker config loaded for OCI push")
+
+
+def run() -> str:
+    """Build a checksum manifest, push as an OCI artifact, and return the store@digest ref."""
+    shared_snapshot = SHARED_DIR / "snapshot.json"
+    if shared_snapshot.exists():
+        snapshot = json.loads(shared_snapshot.read_text())
+    else:
+        snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    _setup_docker_config()
+
+    checksum_manifest: list[dict] = []
+
+    for component in snapshot.get("components", []):
+        name = component.get("name", "")
+        ready_dir = CONTENT_DIR / name / "ready_for_distribution"
+
+        logger.info("Generating checksum manifest for component: %s", name)
+
+        if not ready_dir.is_dir():
+            logger.warning("ready_for_distribution directory not found for %s", name)
+            continue
+
+        checksum_files: dict[str, str] = {}
+        for f in sorted(ready_dir.rglob("*")):
+            if f.is_file() and not f.name.startswith("sha256sum.txt"):
+                checksum = file_utils.sha256(f)
+                checksum_files[f.name] = f"sha256:{checksum}"
+
+        checksum_manifest.append({"component": name, "files": checksum_files})
+
+    archive_dir = Path(tempfile.mkdtemp())
+    try:
+        manifest_path = archive_dir / "checksum_map.json"
+        manifest_path.write_text(json.dumps(checksum_manifest, separators=(",", ":")))
+        logger.info("Checksum manifest contents:")
+        logger.info("%s", manifest_path.read_text())
+
+        archive_path = archive_dir / "checksum_map"
+        with tarfile.open(str(archive_path), "w:gz", compresslevel=9) as tf:
+            tf.add(str(manifest_path), arcname="checksum_map.json")
+
+        pushed_digest = oras_utils.oras_push(
+            OCI_STORE, archive_dir, "checksum_map", "checksum_map"
+        )
+        ref = f"{OCI_STORE}@{pushed_digest}"
+        logger.info("Checksum map pushed to OCI: %s", ref)
+        return ref
+    finally:
+        shutil.rmtree(str(archive_dir), ignore_errors=True)
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Run checksum map build and push; return exit code."""
+    logging.basicConfig(level=logging.INFO)
+    try:
+        ref = run()
+        logger.info("Checksum map pushed to: %s", ref)
+    except Exception as exc:
+        logger.error("ERROR: %s", exc)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/scripts/python/helpers/compress_artifacts.py
+++ b/scripts/python/helpers/compress_artifacts.py
@@ -1,0 +1,245 @@
+#!/usr/bin/env python3
+"""Pull signed binaries from Quay, restore supplementary files, and compress artifacts.
+
+For each component:
+* Pulls signed macOS and Windows OCI artifacts from Quay into a ``signed/`` directory.
+* Restores supplementary files (readme, license, changelog) that were held during signing.
+* Compresses each file entry into the final deliverable format:
+  - macOS / Linux → ``.tar.gz`` (from ``os/arch/`` directory)
+  - Windows → ``.zip`` (from ``os/arch/`` directory, extension corrected from
+    ``.tar.gz``/``.tar``)
+* Updates ``SNAPSHOT_JSON`` to reflect corrected Windows filenames in ``files[]``.
+* Saves the modified snapshot to ``/shared/snapshot.json`` for downstream use.
+
+CLI arguments:
+  ``--quay-url``
+
+Secret mounts:
+  ``QUAY_SECRET_MOUNT``  (default: ``/mnt/quaySecret``)
+
+Other env vars:
+  ``SNAPSHOT_JSON``   – JSON string of the Snapshot spec
+  ``CONTENT_DIR``     – override base directory (default: ``/shared/artifacts``)
+  ``SHARED_DIR``      – override shared volume root (default: ``/shared``)
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import os
+import shutil
+import subprocess
+import tarfile
+import zipfile
+from pathlib import Path
+
+import oras_utils
+
+PROG = "compress_artifacts.py"
+
+QUAY_SECRET_MOUNT = Path(os.environ.get("QUAY_SECRET_MOUNT", "/mnt/quaySecret"))
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+SHARED_DIR = Path(os.environ.get("SHARED_DIR", "/shared"))
+
+logger = logging.getLogger(__name__)
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    """Parse and return CLI arguments."""
+    p = argparse.ArgumentParser(prog=PROG)
+    p.add_argument("--quay-url", required=True, help="Quay repository URL base")
+    return p.parse_args(argv)
+
+
+def _pull_signed_content(quay_url: str, component_name: str, component_dir: Path) -> None:
+    """Pull signed Mac and Windows OCI artifacts from Quay into the component's signed/ dir."""
+    signed_dir = component_dir / "signed"
+    signed_dir.mkdir(parents=True, exist_ok=True)
+
+    if (component_dir / "has_mac").exists():
+        signed_mac_digest = (component_dir / "signed_mac_digest.txt").read_text().strip()
+        subprocess.check_call(
+            ["oras", "pull", f"{quay_url}/signed/{component_name}@{signed_mac_digest}"],
+            cwd=str(signed_dir),
+        )
+
+    if (component_dir / "has_windows").exists():
+        signed_windows_digest = (
+            (component_dir / "signed_windows_digest.txt").read_text().strip()
+        )
+        signed_windows_digest = signed_windows_digest.strip()
+        subprocess.check_call(
+            ["oras", "pull", f"{quay_url}/signed/{component_name}@{signed_windows_digest}"],
+            cwd=str(signed_dir),
+        )
+
+
+def _restore_supplementary(component_dir: Path) -> None:
+    """Move supplementary files from the hold directory back into the signed content tree."""
+    signed_dir = component_dir / "signed"
+    supp_hold = component_dir / "supplementary"
+    for os_name in ("macos", "windows"):
+        supp_os_dir = supp_hold / os_name
+        if not supp_os_dir.is_dir():
+            continue
+        for file in supp_os_dir.rglob("*"):
+            if file.is_file():
+                rel = file.relative_to(supp_os_dir)
+                dest = signed_dir / os_name / rel
+                dest.parent.mkdir(parents=True, exist_ok=True)
+                shutil.move(str(file), str(dest))
+                logger.info("  Restored supplementary file: %s/%s", os_name, rel)
+
+
+def _windows_filename(source_filename: str) -> str:
+    """Replace .tar.gz or .tar extension with .zip for Windows archives."""
+    if source_filename.endswith(".tar.gz"):
+        return source_filename[: -len(".tar.gz")] + ".zip"
+    if source_filename.endswith(".tar"):
+        return source_filename[: -len(".tar")] + ".zip"
+    return source_filename
+
+
+def _compress_file_entry(
+    entry: dict, array_name: str, component_dir: Path, ready_dir: Path
+) -> str:
+    """Compress one file entry into ready_dir and return the (possibly normalized) source path.
+
+    For macOS and Linux entries the source path is returned unchanged. For Windows entries
+    the archive is created as a ``.zip`` instead of ``.tar.gz``/``.tar``, and the returned
+    source path reflects the corrected filename so the snapshot can be updated accordingly.
+
+    Raises RuntimeError on failure (missing source, unknown OS, or empty arch directory).
+
+    Note: all files are currently compressed regardless of type. ISOs should be
+    passed through as-is rather than wrapped in a tarball — this will need to be
+    addressed before ISO delivery is supported.
+    """
+    source = entry.get("source")
+    if not source:
+        raise RuntimeError(f"Missing source field in {array_name}[] entry: {entry}")
+
+    source_filename = Path(source).name
+    os_name = entry.get("os", "")
+    arch = entry.get("arch", "")
+
+    arch_dir = oras_utils.os_arch_dir(
+        os_name,
+        arch,
+        mac_windows_base=component_dir / "signed",
+        linux_base=component_dir / "linux",
+    )
+    if arch_dir is None:
+        raise RuntimeError(f"Unknown OS '{os_name}' in {array_name}[] entry (arch: {arch})")
+
+    if not arch_dir.is_dir() or not any(arch_dir.iterdir()):
+        raise RuntimeError(f"Architecture directory is empty or not found: {arch_dir}")
+
+    # macOS and Linux follow the Unix convention of tar.gz archives; Windows uses zip
+    # because that is the standard expected by Windows users and Developer Portal tooling.
+    if os_name in ("darwin", "linux"):
+        out_path = ready_dir / source_filename
+        with tarfile.open(str(out_path), "w:gz") as tf:
+            for item in sorted(arch_dir.rglob("*")):
+                if item.is_file():
+                    tf.add(str(item), arcname=str(item.relative_to(arch_dir)))
+        logger.info("  Created (%s): %s", array_name, source_filename)
+        return source
+
+    win_filename = _windows_filename(source_filename)
+    out_path = ready_dir / win_filename
+    with zipfile.ZipFile(str(out_path), "w", zipfile.ZIP_DEFLATED) as zf:
+        for item in sorted(arch_dir.rglob("*")):
+            if item.is_file():
+                zf.write(str(item), arcname=str(item.relative_to(arch_dir)))
+    logger.info("  Created (%s): %s", array_name, win_filename)
+    return str(Path(source).parent / win_filename)
+
+
+def compress_component(component: dict, snapshot: dict) -> dict:
+    """Compress all file entries for one component. Returns updated component dict."""
+    name = component.get("name", "")
+    component_dir = CONTENT_DIR / name
+    ready_dir = component_dir / "ready_for_distribution"
+    ready_dir.mkdir(parents=True, exist_ok=True)
+
+    logger.info("Compressing artifacts for component: %s", name)
+
+    # files[] entries are for Developer Portal delivery; staged.files[] entries are
+    # for Customer Portal (Pulp/CDN) delivery.  Both produce archives in
+    # ready_for_distribution/, but only files[] source paths are updated in the snapshot
+    # (Windows .zip correction).
+    files_entries = list(component.get("files") or [])
+    staged_entries = list((component.get("staged") or {}).get("files") or [])
+
+    normalized_files = []
+    if files_entries:
+        logger.info(
+            "  Processing %d files from files[] (Developer Portal):", len(files_entries)
+        )
+        for entry in files_entries:
+            normalized_source = _compress_file_entry(entry, "files", component_dir, ready_dir)
+            normalized_entry = dict(entry)
+            # no-op for mac/linux, .zip correction for windows
+            normalized_entry["source"] = normalized_source
+            normalized_files.append(normalized_entry)
+
+    if staged_entries:
+        logger.info(
+            "  Processing %d files from staged.files[] (Customer Portal):", len(staged_entries)
+        )
+        for entry in staged_entries:
+            _compress_file_entry(entry, "staged.files", component_dir, ready_dir)
+
+    updated_component = dict(component)
+    if files_entries:
+        updated_component["files"] = normalized_files
+
+    return updated_component
+
+
+def run(quay_url: str) -> None:
+    """Pull signed artifacts, restore supplementary files, and compress all components."""
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    quay_user = (QUAY_SECRET_MOUNT / "username").read_text().strip()
+    quay_pass = (QUAY_SECRET_MOUNT / "password").read_text().strip()
+
+    # Log in to the quay.io registry using the hostname only; quay_url contains the full
+    # repository path (e.g. quay.io/org/repo) and is passed to oras pull commands below.
+    logger.info("Logging into Quay...")
+    oras_utils.oras_login("quay.io", quay_user, quay_pass)
+
+    updated_components = []
+    for component in snapshot.get("components", []):
+        name = component.get("name", "")
+        component_dir = CONTENT_DIR / name
+
+        _pull_signed_content(quay_url, name, component_dir)
+        _restore_supplementary(component_dir)
+
+        updated = compress_component(component, snapshot)
+        updated_components.append(updated)
+
+    snapshot["components"] = updated_components
+    snapshot_path = SHARED_DIR / "snapshot.json"
+    snapshot_path.write_text(json.dumps(snapshot), encoding="utf-8")
+    logger.info("Saved modified snapshot to %s", snapshot_path)
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Parse arguments and run artifact compression; return exit code."""
+    logging.basicConfig(level=logging.INFO)
+    args = parse_args(argv[1:] if argv is not None else None)
+    try:
+        run(args.quay_url)
+    except Exception as exc:
+        logger.error("ERROR: %s", exc)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/scripts/python/helpers/extract_artifacts.py
+++ b/scripts/python/helpers/extract_artifacts.py
@@ -1,0 +1,284 @@
+#!/usr/bin/env python3
+"""Extract artifacts from container images.
+
+For each component in SNAPSHOT_JSON that has ``files`` or ``staged.files`` entries:
+* Pulls the container image with ``skopeo copy`` (authenticated via ``select-oci-auth``).
+* Identifies the specific files listed in the RPA and extracts them from the container layers.
+* Creates OS flag files (``has_mac``, ``has_windows``, ``has_linux``) to indicate
+  which signing paths are needed.
+
+Components are processed in parallel, bounded by ``--concurrent-limit``.
+
+CLI arguments:
+  ``--concurrent-limit``
+
+Secret mounts (paths can be overridden via env vars for testing):
+  ``REDHAT_WORKLOADS_TOKEN_MOUNT``  (default: ``/mnt/redhat-workloads-token``)
+
+Other env vars:
+  ``SNAPSHOT_JSON``   – JSON string of the Snapshot spec (set by the task)
+  ``CONTENT_DIR``     – override base directory (default: ``/shared/artifacts``)
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import os
+import shutil
+import subprocess
+import tarfile
+import tempfile
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from pathlib import Path
+
+import authentication
+
+PROG = "extract_artifacts.py"
+
+REDHAT_WORKLOADS_TOKEN_MOUNT = Path(
+    os.environ.get("REDHAT_WORKLOADS_TOKEN_MOUNT", "/mnt/redhat-workloads-token")
+)
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+
+logger = logging.getLogger(__name__)
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    """Parse and return CLI arguments."""
+    p = argparse.ArgumentParser(prog=PROG)
+    p.add_argument(
+        "--concurrent-limit",
+        type=int,
+        default=3,
+        help="Maximum number of components to process in parallel",
+    )
+    return p.parse_args(argv)
+
+
+def _setup_docker_config() -> None:
+    """Write ~/.docker/config.json from the mounted dockerconfig secret."""
+    authentication.setup_docker_config(
+        REDHAT_WORKLOADS_TOKEN_MOUNT / ".dockerconfigjson",
+        strip_noise=True,
+    )
+
+
+def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
+    """Return (wanted_files, layer_extract_dirs) from files[] and staged.files[] entries."""
+    wanted: list[str] = []
+    # Parent directories to extract from container image layers via `tar`. We extract
+    # whole directories rather than individual files because tar requires the full path
+    # prefix to be present in the layer for selective extraction to work.
+    layer_extract_dirs: list[str] = []
+
+    for entry in list(component.get("files") or []) + list(
+        (component.get("staged") or {}).get("files") or []
+    ):
+        source = entry.get("source", "")
+        if not source:
+            continue
+        stripped = source.lstrip("/")
+        wanted.append(stripped)
+        parent = str(Path(stripped).parent)
+        if parent and parent != ".":
+            layer_extract_dirs.append(parent)
+
+    if not layer_extract_dirs:
+        # No explicit parent directories from the RPA entries (e.g. source paths are bare
+        # filenames with no directory component).  Fall back to "releases/", which is the
+        # conventional top-level directory used in Red Hat release container images.
+        layer_extract_dirs = ["releases"]
+
+    unique_files = sorted(set(wanted))
+    unique_dirs = sorted(set(layer_extract_dirs))
+    return unique_files, unique_dirs
+
+
+def _safe_extract_layer(
+    tf: tarfile.TarFile, image_path: str, target_dir: Path, layer_name: str
+) -> bool:
+    """Extract members under image_path/ from a layer tarfile with path safety checks.
+
+    Returns True if any matching members were found, False otherwise.
+    Raises RuntimeError for unsafe entries (path traversal, symlinks, hardlinks, devices).
+    """
+    target_real = target_dir.resolve()
+    found = False
+    for member in tf.getmembers():
+        if not (member.name == image_path or member.name.startswith(f"{image_path}/")):
+            continue
+        found = True
+        if member.issym() or member.islnk() or member.isdev():
+            raise RuntimeError(
+                f"Layer {layer_name} contains unsupported entry type: {member.name}"
+            )
+        member_real = (target_dir / member.name).resolve()
+        if member_real != target_real and target_real not in member_real.parents:
+            raise RuntimeError(f"Layer {layer_name} contains unsafe path: {member.name}")
+        tf.extract(member, path=str(target_dir), filter="data")
+    return found
+
+
+def process_component(component: dict) -> None:
+    """Pull and extract one component's artifacts into CONTENT_DIR/<name>/."""
+    name = component.get("name")
+
+    num_files = len(component.get("files") or [])
+    # staged files are for Customer Portal (Pulp)
+    num_staged = len((component.get("staged") or {}).get("files") or [])
+    if num_files == 0 and num_staged == 0:
+        logger.info(
+            "Skipping component '%s' - no files defined (not a binary artifact component)",
+            name,
+        )
+        return
+
+    pullspec = component.get("containerImage")
+    if not pullspec:
+        raise ValueError(f"Component '{name}' is missing 'containerImage'")
+
+    destination = CONTENT_DIR / name
+    logger.info("Extracting component '%s' to: %s", name, destination)
+    destination.mkdir(parents=True, exist_ok=True)
+
+    tmp_dir = Path(tempfile.mkdtemp())
+    try:
+        auth_file: Path | None = None
+        try:
+            with tempfile.NamedTemporaryFile(mode="wb", delete=False) as auth_fp:
+                auth_file = Path(auth_fp.name)
+                auth_data = subprocess.check_output(
+                    ["select-oci-auth", pullspec], stderr=subprocess.PIPE
+                )
+                auth_fp.write(auth_data)
+
+            subprocess.check_call(
+                [
+                    "skopeo",
+                    "copy",
+                    "--retry-times",
+                    "3",
+                    "--authfile",
+                    str(auth_file),
+                    f"docker://{pullspec}",
+                    f"dir:{tmp_dir}",
+                ]
+            )
+        finally:
+            if auth_file is not None:
+                auth_file.unlink(missing_ok=True)
+
+        wanted_files, extract_dirs = _get_source_paths(component)
+        logger.info("Files to extract from RPA: %s", wanted_files)
+
+        manifest = json.loads((tmp_dir / "manifest.json").read_text())
+        layer_digests = [layer["digest"] for layer in manifest.get("layers", [])]
+
+        for digest in layer_digests:
+            layer_file = tmp_dir / digest.removeprefix("sha256:")
+            if not layer_file.exists():
+                continue
+            with tarfile.open(str(layer_file)) as tf:
+                for image_path in extract_dirs:
+                    if _safe_extract_layer(tf, image_path, tmp_dir, layer_file.name):
+                        logger.info("Extracting %s/ from %s...", image_path, layer_file.name)
+                    else:
+                        logger.info(
+                            "skipping %s. It doesn't contain the %s dir",
+                            layer_file.name,
+                            image_path,
+                        )
+
+        for wanted in wanted_files:
+            src = tmp_dir / wanted
+            if src.is_file():
+                shutil.copy2(str(src), str(destination / src.name))
+            else:
+                logger.error("Expected file not found in container: %s", wanted)
+                raise RuntimeError(
+                    f"File '{wanted}' declared in RPA was not found in any container layer"
+                )
+    finally:
+        shutil.rmtree(tmp_dir, ignore_errors=True)
+
+
+def _create_os_flag_files(snapshot: dict) -> None:
+    """Create has_mac / has_windows / has_linux flag files based on RPA entries."""
+    for component in snapshot.get("components", []):
+        name = component.get("name", "")
+        component_dir = CONTENT_DIR / name
+        if not component_dir.is_dir():
+            continue
+
+        logger.info("Checking configured OS types for component: %s", name)
+
+        all_file_entries = list(component.get("files") or []) + list(
+            (component.get("staged") or {}).get("files") or []
+        )
+
+        def _matches(entry: dict, os_name: str) -> bool:
+            """Return True if the file entry is associated with the given OS."""
+            if entry.get("os") == os_name:
+                return True
+            for field in ("source", "filename"):
+                if os_name in (entry.get(field) or ""):
+                    return True
+            return False
+
+        if any(_matches(e, "darwin") for e in all_file_entries):
+            (component_dir / "has_mac").touch()
+            logger.info("  - macOS content detected")
+
+        if any(_matches(e, "windows") for e in all_file_entries):
+            (component_dir / "has_windows").touch()
+            logger.info("  - Windows content detected")
+
+        if any(_matches(e, "linux") for e in all_file_entries):
+            (component_dir / "has_linux").touch()
+            logger.info("  - Linux content detected")
+
+
+def run(concurrent_limit: int) -> None:
+    """Extract artifacts from all snapshot components and write OS flag files."""
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    _setup_docker_config()
+    CONTENT_DIR.mkdir(parents=True, exist_ok=True)
+
+    components = snapshot.get("components", [])
+    errors: list[str] = []
+
+    with ThreadPoolExecutor(max_workers=concurrent_limit) as executor:
+        futures = {
+            executor.submit(process_component, comp): comp.get("name", f"index {idx}")
+            for idx, comp in enumerate(components)
+        }
+        for future in as_completed(futures):
+            comp_name = futures[future]
+            try:
+                future.result()
+            except Exception as exc:
+                errors.append(f"Component '{comp_name}': {exc}")
+
+    if errors:
+        raise RuntimeError("\n".join(errors))
+
+    _create_os_flag_files(snapshot)
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Parse arguments and run artifact extraction; return exit code."""
+    logging.basicConfig(level=logging.INFO)
+    args = parse_args(argv[1:] if argv is not None else None)
+    try:
+        run(args.concurrent_limit)
+    except Exception as exc:
+        logger.error("ERROR: %s", exc)
+        return 1
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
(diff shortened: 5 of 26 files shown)


pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "pubtools-pulp==1.33.2",
     "pubtools-exodus==1.5.2",
     "pubtools-marketplacesvm==1.9.0",
+    "starmap-client==2.4.0",  # 2.4.1+ uses PEP604 unions; UBI9 image still runs Python 3.9
```
swickersh: this and the uv.lock won't be necessary when the ubi10 and python upgrade change lands. I'll rebase at that time.

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT = Path(
+    os.environ.get(
+        "TRUSTED_ARTIFACTS_DOCKERCONFIG_MOUNT", "/mnt/trusted_artifacts_dockerconfig"
+    )
+)
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+SHARED_DIR = Path(os.environ.get("SHARED_DIR", "/shared"))
+
+OCI_STORE = "quay.io/konflux-ci/release-service-trusted-artifacts"
```
querti: I could be mistaken, but isn't it possible to customize it for the pipeline? Should this be hardcoded?

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def _sha256(path: Path) -> str:
```
querti: Have we decided on guidelines for the Python scripts? Will docstrings not be enforced?

querti: This function seems duplicated in _push_component_to_cdn and _sha256sum functions

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+    _setup_docker_config()
+
+    checksum_manifest: list[dict] = []
+
+    for component in snapshot.get("components", []):
+        name = component.get("name", "")
+        ready_dir = CONTENT_DIR / name / "ready_for_distribution"
+
+        print(f"Generating checksum manifest for component: {name}")
```
querti: do we want to use print versus a logger?

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+        print(f"Generating checksum manifest for component: {name}")
+
+        if not ready_dir.is_dir():
+            print(f"Warning: ready_for_distribution directory not found for {name}")
+            continue
+
+        checksum_files: dict[str, str] = {}
+        for f in sorted(ready_dir.rglob("*")):
+            if f.is_file() and not f.name.startswith("sha256sum.txt"):
```
querti, line 85: could this file have a suffix?

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+                checksum_files[f.name] = f"sha256:{checksum}"
+
+        checksum_manifest.append({"component": name, "files": checksum_files})
+
+    archive_dir = Path(tempfile.mkdtemp())
+    try:
+        manifest_path = archive_dir / "checksum_map.json"
+        manifest_path.write_text(json.dumps(checksum_manifest, separators=(",", ":")))
+        print("Checksum manifest contents:")
+        print(manifest_path.read_text())
```
querti: Could this output be way too long?

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+                stderr=subprocess.STDOUT,
+                text=True,
+            )
+        finally:
+            auth_file.unlink(missing_ok=True)
+
+        print(oras_output)
+
+        match = re.search(r"^Digest:\s+(sha256:[a-f0-9]{64})", oras_output, re.MULTILINE)
+        if not match:
```
querti: When could this happen?

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+        print(f"Checksum map pushed to OCI: {ref}")
+        return ref
+    finally:
+        shutil.rmtree(str(archive_dir), ignore_errors=True)
+
+
+def main(argv: list[str] | None = None) -> int:
+    rpath, cmap_path = tekton.result_paths("RESULT_RESULT", "RESULT_CHECKSUM_MAP")
+
+    if rpath.read_text(encoding="utf-8").strip() != "Success":
```
querti: I may be misunderstanding this, but if the previous step failed, shouldn't have this step not be triggered at all?

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+        print("Previous step failed. Exiting...")
+        cmap_path.write_text("", encoding="utf-8")
+        return 0
```
querti: return 0?

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
```diff
@@ -0,0 +1,168 @@
...
+
+    checksum_map_ref = ""
+    try:
+        checksum_map_ref = run()
+    except Exception as exc:
+        rpath.write_text(f"{PROG}: ERROR {exc}", encoding="utf-8")
+        cmap_path.write_text(checksum_map_ref, encoding="utf-8")
```
querti: why are we writing this? Is it a tekton thing?

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env python3
+"""Pull signed binaries from Quay, restore supplementary files, and compress artifacts.
+
+For each component:
+* Pulls signed macOS and Windows OCI artifacts from Quay into a ``signed/`` directory.
+* Restores supplementary files (readme, license, changelog) that were held during signing.
+* Compresses each file entry into the final deliverable format:
+  - macOS / Linux → ``.tar.gz`` (from ``os/arch/`` directory)
```
querti, line 8: this is off topic, but why not use zip for all?

scripts/python/tasks/internal/push_artifacts_to_cdn/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+* Restores supplementary files (readme, license, changelog) that were held during signing.
+* Compresses each file entry into the final deliverable format:
+  - macOS / Linux → ``.tar.gz`` (from ``os/arch/`` directory)
+  - Windows → ``.zip`` (from ``os/arch/`` directory, extension corrected from
+    ``.tar.gz``/``.tar``)
+* Updates ``SNAPSHOT_JSON`` to reflect corrected Windows filenames in ``files[]``.
+* Saves the modified snapshot to ``/shared/snapshot.json`` for later steps.
+
+Tekton parameters:
+  ``quayURL`` → ``--quay-url``
```
querti: why is some info passed using args, and other using env vars?

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+
+def run(quay_url: str) -> None:
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    quay_user = (QUAY_SECRET_MOUNT / "username").read_text().strip()
+    quay_pass = (QUAY_SECRET_MOUNT / "password").read_text().strip()
+
+    print("Logging into Quay...")
+    subprocess.check_call(
+        ["oras", "login", "quay.io", "-u", quay_user, "-p", quay_pass],
```
querti: quay_url is not used here?

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+        return None
+
+    if not arch_dir.is_dir() or not any(arch_dir.iterdir()):
+        print(
+            f"  Warning: Architecture directory is empty or not found: {arch_dir}",
+            file=sys.stderr,
+        )
+        return None
+
+    if os_name in ("darwin", "linux"):
```
querti, line 142: Reading these scripts, it's kind of hard to figure out what the file structure is. Maybe this could be added in the header, so we can more easily understand what the scripts are trying to accomplish?

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+def compress_component(component: dict, snapshot: dict) -> dict:
+    """Compress all file entries for one component. Returns updated component dict."""
+    name = component.get("name", "")
+    component_dir = CONTENT_DIR / name
+    ready_dir = component_dir / "ready_for_distribution"
+    ready_dir.mkdir(parents=True, exist_ok=True)
+
+    print(f"Compressing artifacts for component: {name}")
+
+    files_entries = list(component.get("files") or [])
```
querti, line 174: It could be helpful to document what the snapshot format is, it's pretty hard to imagine what the script does without having proper context.

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+    staged_entries = list((component.get("staged") or {}).get("files") or [])
+
+    updated_files = []
+    if files_entries:
+        print(f"  Processing {len(files_entries)} files from files[] (Developer Portal):")
+        for entry in files_entries:
+            corrected = _process_file_entry(entry, "files", component_dir, ready_dir)
```
querti: so this return value is returned for all OSes but only used for Windows? Seems a bit confusing, especially if this is not documented.

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+            if corrected is not None:
+                updated_entry = dict(entry)
+                if entry.get("os") == "windows":
+                    updated_entry["source"] = corrected
+                updated_files.append(updated_entry)
+            else:
```
querti: Is it possible that WIndows path is not corrected? is this a valid state?

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,253 @@
...
+                updated_files.append(entry)
+
+    if staged_entries:
```
querti, line 189: what is the difference between file entries and staged entries?

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+    p.add_argument(
+        "--concurrent-limit",
+        type=int,
+        default=3,
+        help="Maximum number of components to process in parallel",
+    )
+    return p.parse_args(argv)
+
+
+def _setup_docker_config() -> None:
```
querti, line 60: This function is a bit different, but seems duplicated

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+
+    unique_files = sorted(set(wanted))
+    unique_dirs = sorted(set(dirs))
+    return unique_files, unique_dirs
+
+
+def process_component(component: dict, component_index: int) -> None:
+    """Pull and extract one component's artifacts into CONTENT_DIR/<name>/."""
+    num = component_index + 1
+    name = component.get("name")
```
querti, line 126: Is this a possible scenario? Do we need to pass the index, will it help us with debugging?

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+        source = entry.get("source", "")
+        if not source:
+            continue
+        stripped = source.lstrip("/")
+        wanted.append(stripped)
+        parent = str(Path(stripped).parent)
+        if parent and parent != ".":
+            dirs.append(parent)
+
+    if not dirs:
```
querti: what does this mean?

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+        manifest = json.loads((tmp_dir / "manifest.json").read_text())
+        layer_digests = [layer["digest"] for layer in manifest.get("layers", [])]
+
+        for image_path in extract_dirs:
+            print(f"Looking for directory: {image_path}")
+            for digest in layer_digests:
+                layer_file = tmp_dir / digest.removeprefix("sha256:")
+                if not layer_file.exists():
+                    continue
+                result = subprocess.run(
```
querti: Wouldn't it be better to flip the nested fors, so that each layer's contents are only listed once? IDK, maybe it's not a big performance difference

scripts/python/tasks/internal/push_artifacts_to_cdn/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+
+        def _matches(entry: dict, os_name: str) -> bool:
+            if entry.get("os") == os_name:
+                return True
+            for field in ("source", "filename"):
+                if os_name in (entry.get(field) or ""):
+                    return True
+            return False
+
+        darwin_key = "darwin"
```
querti: why do we need this in separate var?

scripts/python/helpers/generate_checksums.py
```diff
@@ -0,0 +1,280 @@
...
+CHECKSUM_CREDENTIALS_MOUNT = Path(
+    os.environ.get("CHECKSUM_CREDENTIALS_MOUNT", "/mnt/checksum_credentials")
+)
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+SHARED_DIR = Path(os.environ.get("SHARED_DIR", "/shared"))
+
+
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+    p = argparse.ArgumentParser(prog=PROG)
+    p.add_argument(
```
querti, line 61: Does this need to be an arg? Is there an expectation of a different realm?

scripts/python/helpers/generate_checksums.py
```diff
@@ -0,0 +1,280 @@
...
+        subprocess.check_call(
+            ["kinit", "-kt", str(keytab), f"{checksum_user}@{kerberos_realm}"]
+        )
+    finally:
+        keytab.unlink(missing_ok=True)
+
+
+def run(kerberos_realm: str, pipeline_run_uid: str) -> None:
+    author = os.environ.get("AUTHOR", "").strip()
+    if not author:
```
querti, line 93: author is checked for existence, but not other env vars? are they optional?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+        staged_files = (component.get("staged") or {}).get("files") or []
+        if not staged_files:
+            print(
+                f"Warning: No staged.files specified in RPA for Customer Portal push "
+                f"for {component_name}",
+                file=sys.stderr,
+            )
+            return
+
+        for i, sf in enumerate(staged_files):
```
querti: is indexing used anywhere?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+            source_path = sf.get("source", "")
+            dest_filename = sf.get("filename", "")
+            if not source_path:
+                continue
+
+            source_filename = Path(source_path).name
+            # Handle windows .tar.gz/.tar → .zip conversion
+            if "windows" in source_filename:
+                for old_ext, new_ext in [(".tar.gz", ".zip"), (".tar", ".zip")]:
+                    if source_filename.endswith(old_ext):
```
querti, line 193: wasn't the extension correction done in a different step?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+                            if dest_filename.endswith(old_ext):
+                                dest_filename = dest_filename[: -len(old_ext)] + new_ext
+                        break
+
+            if not dest_filename:
+                dest_filename = source_filename
+
+            src = component_dir / source_filename
+            if src.is_file():
+                import shutil
```
querti: why add the import here?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+            if f.is_file():
+                staged_json["payload"]["files"].append(
+                    {
+                        "filename": f.name,
+                        "relative_path": str(f.relative_to(staging_dir)),
+                        "version": version,
+                    }
+                )
+
+        import yaml  # type: ignore
```
querti: why import here?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+                "--pulp-cert",
+                str(pulp_cert),
+                "--pulp-key",
+                str(pulp_key),
+                "--udcache-url",
+                udc_url,
+            ]
+        )
+    finally:
+        import shutil
```
querti: why import here?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+
+    result_published_path = Path(os.environ.get("RESULT_PUBLISHED_FILES", "/dev/null"))
+    published_files = []
+    for ready_dir in sorted(CONTENT_DIR.glob("*/ready_for_distribution")):
+        for f in sorted(ready_dir.iterdir()):
+            if f.is_file():
+                published_files.append(f.name)
+
+    published_str = "\n".join(published_files)
+    if len(published_str) > 512:
```
querti: what does this mean?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+    os.environ["PULP_CERT_FILE"] = str(pulp_cert)
+    os.environ["PULP_KEY_FILE"] = str(pulp_key)
+    os.environ["UDCACHE_CERT"] = str(udc_cert)
+    os.environ["UDCACHE_KEY"] = str(udc_key)
+    os.environ["EXODUS_GW_ENV"] = exodus_gw_env
+    os.environ["EXODUS_GW_URL"] = exodus_url
+    os.environ["EXODUS_PULP_HOOK_ENABLED"] = "True"
+    os.environ["EXODUS_GW_TIMEOUT"] = "7200"
+
+    result_published_path = Path(os.environ.get("RESULT_PUBLISHED_FILES", "/dev/null"))
```
querti: /dev/null? what will this do?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+    for component in snapshot.get("components", []):
+        name = component.get("name", "")
+        has_staged = bool(component.get("staged"))
+        has_cgw = bool(component.get("contentGateway"))
+
+        print(f"Processing component: {name} (staged={has_staged}, cgw={has_cgw})")
+
+        if has_staged:
+            _push_component_to_pulp(name, snapshot, pulp_url, pulp_cert, pulp_key, udc_url)
+        elif has_cgw:
```
querti: if I understand correctly, cgw needs to publish to two places?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+            _create_exodus_conf(
+                exodus_conf_path, exodus_gw_cert, exodus_gw_key, exodus_url, exodus_gw_env
+            )
+            _push_component_to_cdn(name, exodus_conf_path)
+
+        if has_cgw:
```
querti, line 363: can this be merged with previous condition?

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,250 @@
...
+    lower = path.name.lower()
+    if "." in lower:
+        base, ext = lower.rsplit(".", 1)
+        ext = f".{ext}"
+    else:
+        base, ext = lower, ""
+    if base in SUPPLEMENTARY_NAMES:
+        if not ext:
+            return True
+        if ext in SUPPLEMENTARY_EXTS:
```
querti, line 65: can there be supplementary names with nonsuplementary extensions?

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,250 @@
...
+
+def run(quay_url: str, pipeline_run_uid: str) -> None:
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    quay_user = (QUAY_SECRET_MOUNT / "username").read_text().strip()
+    quay_pass = (QUAY_SECRET_MOUNT / "password").read_text().strip()
+
+    print("Logging into Quay...")
+    subprocess.check_call(
+        ["oras", "login", "quay.io", "-u", quay_user, "-p", quay_pass],
```
querti: hardcoded quay.io?

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,250 @@
...
+        if num_files == 0 and num_staged == 0:
+            print(f"Skipping component '{name}' - no files or staged.files defined")
+            continue
+
+        component_dir = CONTENT_DIR / name
+        unsigned_dir = component_dir / "unsigned"
+
+        has_mac = (component_dir / "has_mac").exists()
+        has_windows = (component_dir / "has_windows").exists()
+        has_linux = (component_dir / "has_linux").exists()
```
querti, line 151: is this used anywhere?

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,258 @@
...
+) -> str:
+    return f"""#!/bin/bash
+set -eux
+
+mkdir -p {temp_dir}
+mkdir -p {binary_path}
+
+cd {temp_dir}
+set +x
+/usr/local/bin/oras login quay.io -u {quay_user} -p {quay_pass}
```
querti: quay url?

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,258 @@
...
+    binary_path: str,
+    zip_path: str,
+    digest_file: str,
+    keychain_password: str,
+    signing_identity: str,
+    apple_id: str,
+    team_id: str,
+    app_specific_password: str,
+) -> str:
+    return f"""#!/bin/bash
```
querti, line 72: maybe this script could be in its own file?

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,258 @@
...
+                    + [
+                        f"{mac_user}@{mac_host}:{temp_dir}/push_digest.txt",
+                        str(component_dir / "signed_mac_digest.txt"),
+                    ]
+                )
+                if scp_result.returncode != 0:
+                    ssh_exit = scp_result.returncode
+                    failed_op = "scp of signed digest"
+
+            subprocess.run(
```
querti: if this fails, maybe at least print a warning?

scripts/python/helpers/sign_windows.py
```diff
@@ -0,0 +1,254 @@
...
+    quay_pass: str,
+    component_name: str,
+    unsigned_digest: str,
+    pipeline_run_uid: str,
+    windows_temp_dir: str,
+) -> str:
+    return f"""
+mkdir %TEMP%\\{windows_temp_dir} && cd /d %TEMP%\\{windows_temp_dir}
+@echo off
+oras login quay.io -u {quay_user} -p {quay_pass}
```
querti, line 70: non custom quay

scripts/python/helpers/sign_windows.py
```diff
@@ -0,0 +1,254 @@
...
+    *,
+    quay_url: str,
+    quay_user: str,
+    quay_pass: str,
+    component_name: str,
+    unsigned_digest: str,
+    pipeline_run_uid: str,
+    windows_temp_dir: str,
+) -> str:
+    return f"""
```
querti, line 67: maybe add the script to its own file?

scripts/python/helpers/sign_windows.py
```diff
@@ -0,0 +1,254 @@
...
+            unsigned_digest=unsigned_digest,
+            pipeline_run_uid=pipeline_run_uid,
+            windows_temp_dir=windows_temp_dir,
+        )
+
+        local_script = Path(tempfile.mktemp(suffix=".bat"))
+        local_script.write_text(script_content)
+
+        try:
+            subprocess.check_call(
```
querti, line 185: this result is not captured

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,258 @@
...
+MAC_SSH_KEY_MOUNT = Path(os.environ.get("MAC_SSH_KEY_MOUNT", "/mnt/secrets"))
+MAC_HOST_CREDS_MOUNT = Path(os.environ.get("MAC_HOST_CREDS_MOUNT", "/mnt/macHostCredentials"))
+MAC_SIGNING_CREDS_MOUNT = Path(
+    os.environ.get("MAC_SIGNING_CREDS_MOUNT", "/mnt/macSigningCredentials")
+)
+QUAY_SECRET_MOUNT = Path(os.environ.get("QUAY_SECRET_MOUNT", "/mnt/quaySecret"))
+CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
+
+
+def _build_signing_script(
```
querti, line 53: Is code injection a concern? Should there be some sanitation of the input?

scripts/python/helpers/build_checksum_map.py
swickersh: pep8 says docstrings [aren't required](https://peps.python.org/pep-0008/#documentation-strings) for non-public functions.

I went ahead and updated all functions to have them though. Easy enough if it's helpful

scripts/python/helpers/build_checksum_map.py
swickersh: Followed the pre-existing code: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L1868

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
swickersh: Updated to logger instead of print.
But I will point out a lot of this repo is using print for python

scripts/python/helpers/build_checksum_map.py
swickersh: pre-existing logic: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L1882

scripts/python/helpers/build_checksum_map.py
swickersh: there can also be a sha256sum.txt.sig and sha256sum.txt.gpg
The original code handles that: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L1847

And this should be functionally equivalent.

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
swickersh: usually it's fairly short

It's also pre-existing: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L1858

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
swickersh: We do this with tekton in bash: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L1783

IIUC, we want the error to land in the results file and this pattern allows any later cleanup steps to run.

```
Task: push-artifacts-to-cdn-task
  Step 1: python3 extract_artifacts.py     → writes "Success" or "ERROR ..." to result file
  Step 2: python3 push_unsigned.py         → checks result file first, exits if not "Success"
  Step 3: python3 sign_mac.py              → checks result file first, exits if not "Success"
  Step 4: python3 sign_windows.py          → checks result file first, exits if not "Success"
  Step 5: python3 compress_artifacts.py    → checks result file first, exits if not "Success"
  Step 6: python3 generate_checksums.py    → checks result file first, exits if not "Success"
  Step 7: python3 push_artifacts.py        → checks result file first, exits if not "Success"
  Step 8: python3 build_checksum_map.py    → checks result file first, exits if not "Success"
```


edit - this is moot now.
Now we are using a single step in the catalog to call a wrapper (push-artifacts-to-cdn.py) which in turn calls the helpers.

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
swickersh: Same as previous comment. Success/failure will be determined from contents of results file.

scripts/python/tasks/internal/push_artifacts_to_cdn/build_checksum_map.py
swickersh: Yes. Writes to the results file

scripts/python/helpers/compress_artifacts.py
swickersh: This is carried on from the legacy pipeline. 
The reasoning is basically that zip is natively supported on Windows whereas tarballs are not like with linux/mac

scripts/python/tasks/internal/push_artifacts_to_cdn/compress_artifacts.py
swickersh: CLI args — Tekton task parameters that are small scalar values (strings, ints):

- --quay-url, --pipeline-run-uid, --kerberos-realm etc.
- These map directly from params in the task YAML via the args: field
- Easy to see in the task definition, easy to pass in unit tests

Environment variables — three categories:

- Large data — SNAPSHOT_JSON can be 10s of KB. CLI args have OS-level length limits (typically ~2MB total, but individual args get unwieldy). Tekton sets env vars from params too, but for large blobs env var is the safer route. Also pre-existing — the original bash always got SNAPSHOT_JSON from env.

- Tekton result paths — RESULT_RESULT, RESULT_CHECKSUM_MAP etc. These are file paths injected by Tekton at runtime pointing to /tekton/results/.... They aren't really "task parameters" — Tekton sets them internally and the task YAML just passes them through as env vars.

- Secret mount paths — QUAY_SECRET_MOUNT, CONTENT_DIR etc. These are overridable defaults (mainly for testing) where you want to be able to say "in prod use /mnt/quaySecret, in tests use /tmp/fake-secret". Env vars are the natural fit for this.

scripts/python/helpers/compress_artifacts.py
swickersh: Sure - I can add something. Usually the unit tests are a good indicator of what we expect

scripts/python/helpers/compress_artifacts.py
swickersh: I'll generate a sample snapshot

scripts/python/helpers/compress_artifacts.py
swickersh: staged entries are for customer portal and files are developer portal.  We have plans to use better field name for that. I can't update it here though because it originates in the RPA like that.

scripts/python/helpers/compress_artifacts.py
swickersh: So yea the quay url would be something like quay.io/my-org/my-repo but oras login only expects the registry hostname (quay.io).  I agree it would be better to extract that from the variable.

This is pre-existing however, https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/push-artifacts-to-cdn-task/push-artifacts-to-cdn-task.yaml#L506

I'll compile a list of small tech debt and put it in a separate jira.  You're welcome to leave this comment unresolved until I paste the Jira here.

scripts/python/helpers/extract_artifacts.py
swickersh: I'll add refactoring common functions to helpers to the follow up jira.

scripts/python/helpers/extract_artifacts.py
swickersh: renamed variable and added a comment.

scripts/python/helpers/extract_artifacts.py
swickersh: Good point. Yea this seems unnecessary. Removed.

scripts/python/helpers/extract_artifacts.py
swickersh: Sure. Implemented, thanks.

scripts/python/helpers/generate_checksums.py
swickersh: It's pre-existing. And this is a public project... (although much of our code seems like it would be too specific to us to use without forking and changing)

scripts/python/helpers/generate_checksums.py
swickersh: The env vars are required tekton params anway, so this is unnecessary here. But it is pre-existing.
I'll go ahead and remove it.

scripts/python/helpers/push_artifacts.py
swickersh: no, this appears leftover from a previous version. removed. Thanks

scripts/python/helpers/push_artifacts.py
swickersh: compress_artifacts.py converts Windows .tar.gz/.tar archives to .zip and writes the
corrected filenames back into files[] in the shared snapshot. However, staged.files[]
is not updated in the snapshot — the .zip files exist on disk with the corrected names,
but staged.files[] still references the original .tar.gz/.tar filenames. We detect and
apply the same rename here so pulp_push_wrapper gets the actual filename on disk.

There's no fundamental technical reason why one path updates the snapshot and the other doesn't — both use the same _process_file_entry function and the corrected filename is available. It's just an inconsistency that accumulated over time.

This is pre-existing as well. I'll add it to the tech-debt jira. I'm hesitant to fix so much pre-existing logic with this change.

scripts/python/helpers/push_artifacts.py
swickersh: fixed.

scripts/python/helpers/push_artifacts.py
swickersh: fixed

scripts/python/helpers/push_artifacts.py
swickersh: fixed

scripts/python/helpers/push_artifacts.py
swickersh: it's helpful for testing outside of tekton when result_published_files isn't set. I'll remove it though. thanks

scripts/python/helpers/push_artifacts.py
swickersh: This is almost certainly a leftover guard from when RESULT_PUBLISHED_FILES was written directly to a Tekton result (which has a 4KB limit).

removed

scripts/python/helpers/push_artifacts.py
swickersh: has_staged means RPA speciies staged.files[] which are intended for pulp/customer portal
has_cgw means RPA specifies files[] which are intended for CGW.

Any combination of those is allowed by this pipeline

scripts/python/helpers/push_artifacts.py
swickersh: no, it is intentional for the separate destination mentioned above

scripts/python/helpers/push_unsigned.py
swickersh: no, something like README.html would intentionally be excluded.
I think we wanted to keep a narrow allow-list and it can be revisted or maybe eventually passed from konflux-release-data after prodsec gating.

scripts/python/helpers/push_unsigned.py
swickersh: addressed in another comment

scripts/python/helpers/push_unsigned.py
swickersh: https://github.com/konflux-ci/release-service-utils/pull/769/changes#diff-b6a2d808f0523bd351e3fb059db1fceaabbae6fa8c513b4334f3d520c15b6fa8R100

scripts/python/helpers/sign_mac.py
swickersh: All inputs come from Kubernetes secret mounts or Tekton params controlled by the release team, not from user-supplied data or external sources

It's also pre-existing. 

If I'm overlooking a vulnerability here please push back.

scripts/python/helpers/sign_mac.py
swickersh: Yea it could. It's also pre-existing.
That would be better in a follow up IMO

scripts/python/helpers/sign_mac.py
swickersh: addressed in another comment

scripts/python/helpers/sign_mac.py
swickersh: Good point. done

scripts/python/helpers/sign_windows.py
swickersh: same comment as above

scripts/python/helpers/sign_windows.py
swickersh: same comment as earlier

scripts/python/helpers/sign_windows.py
swickersh: We didn't capture the output of SCP'ing the script to the hosts when this was bash either.          
I believe the concern was not accidentally capturing the script with sensitive info in the logs

check_call already raises CalledProcessError which includes the return code. The exception propagates up through the try block and ultimately causes the signing to fail with an error.

scripts/python/helpers/compress_artifacts.py
swickersh: I agree that's confusing. Updated var name, added a comment here and also more detailed comment inside _process_file_entry

scripts/python/helpers/compress_artifacts.py
swickersh: It seems pretty unlikely. Removed

swickersh:
@querti thanks for the review so far.

I refactored since your first review and made a single wrapper script (push-artifacts-to-cdn.py) which calls these various helper scripts. I've tried to add more comments and also the order of the scripts called from the wrapper should help make the flow of this code more understandable.

scripts/python/helpers/compress_artifacts.py
swickersh: https://github.com/konflux-ci/release-service-utils/pull/769/changes#diff-45040a0dd8c40b24a339696a5b878d668a28f61b45ff05d0c88f69ffe7f0f236R19

scripts/python/helpers/compress_artifacts.py
swickersh: https://github.com/konflux-ci/release-service-utils/pull/769/changes#diff-b65f63567b9b14270e8789de567d94421de24ad604e368781e3226b4baa2fa34R17-R56

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+            (component_dir / "has_windows").touch()
+            print("  - Windows content detected")
+
+        if any(_matches(e, "linux") for e in all_file_entries):
+            (component_dir / "has_linux").touch()
+            print("  - Linux content detected")
+
+
+def run(concurrent_limit: int) -> None:
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Disagree. These are library modules invoked from a single orchestrator entry point (`push_artifacts_to_cdn.py`), not standalone CLI scripts. `SNAPSHOT_JSON` is a Tekton-injected environment variable that is legitimately consumed by every module in this PR — the same pattern is used across `sign_windows.py`, `push_artifacts.py`, `generate_checksums.py`, etc. Adding duplicate `argparse` SNAPSHOT_JSON handling to each module would add boilerplate without improving testability (tests already use `monkeypatch.setenv`) or standalone usability. No change made.

scripts/python/helpers/test_extract_artifacts.py
```diff
@@ -0,0 +1,405 @@
...
+
+import json
+import tarfile
+from pathlib import Path
+from unittest import mock
+
+import pytest
+
+from push_artifacts_to_cdn import extract_artifacts
+
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Disagree. The PR description explicitly calls out "Unit tests achieve 98% coverage across all 8 scripts." The tests are intentional and valuable. There is no compliance checklist prohibiting tests in this PR — this appears to be a hallucinated constraint. No change made.

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,258 @@
...
+    --password "{app_specific_password}"
+set -x
+
+SIGNED_TAG="{pipeline_run_uid}-mac"
+PUSH_OUTPUT=$(/usr/local/bin/oras push --annotation=quay.expires-after=1d \\
+  "{quay_url}/signed/{component_name}:$SIGNED_TAG" macos)
+SIGNED_DIGEST=$(echo "$PUSH_OUTPUT" | grep 'Digest:' | awk '{{print $2}}')
+echo -n "$SIGNED_DIGEST" >> "{digest_file}"
+echo "Process completed successfully."
+"""
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Fixed. The SSH cleanup commands in `sign_mac.py` and `sign_windows.py` now use proper quoting, consistent with the existing pattern in `generate_checksums.py`. In `sign_mac.py`, `shlex.quote()` is now applied to `temp_dir` and `mac_script_path` in the `rm -rf` cleanup command. In `sign_windows.py`, the PowerShell `Remove-Item -LiteralPath` arguments are now wrapped in single quotes. The bash script template body was not changed — `component_name` follows Kubernetes naming rules (alphanumeric + hyphens only) and `pipeline_run_uid` is a UUID, so neither can contain shell metacharacters.

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,384 @@
...
+    exodus_key: str,
+    pulp_cert: str,
+    pulp_key: str,
+    udc_cert: str,
+    udc_key: str,
+) -> tuple[Path, Path, Path, Path, Path, Path]:
+    def _write(name: str, content: str) -> Path:
+        p = Path(f"/tmp/{name}")
+        p.write_text(content.strip() + "\n")
+        return p
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Partially fixed. `_write_cert_files` in `push_artifacts.py` now calls `chmod(0o600)` on each written cert/key file, restricting access to the owner only. Using `tempfile.TemporaryDirectory()` for unpredictable paths was not done — these paths are set in environment variables (`EXODUS_GW_CERT`, `PULP_CERT_FILE`, etc.) and referenced throughout the rest of the function; a temp directory would require tracking lifetime across the full `run()` scope. The risk is also lower than in a shared system context since this runs inside an isolated Kubernetes pod. The permission fix addresses the most practical concern.

scripts/python/helpers/generate_checksums.py
```diff
@@ -0,0 +1,273 @@
...
+
+    keytab = Path(tempfile.mktemp(suffix=".keytab"))
+    try:
+        keytab.write_bytes(base64.b64decode(keytab_b64))
+        subprocess.check_call(
+            ["kinit", "-kt", str(keytab), f"{checksum_user}@{kerberos_realm}"]
+        )
+    finally:
+        keytab.unlink(missing_ok=True)
+
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Fixed. `generate_checksums._kinit()` now uses `tempfile.mkstemp()` instead of `tempfile.mktemp()`. `mkstemp()` creates the keytab file atomically with `0o600` permissions and returns an open fd, eliminating the TOCTOU window. The bytes are written via `os.write(fd, ...)` before closing the descriptor, and the existing `finally` cleanup is unchanged.

scripts/python/tasks/internal/push_artifacts_to_cdn.py
```diff
@@ -0,0 +1,112 @@
...
+    rpath, cmap_path, published_path = tekton.result_paths(
+        "RESULT_RESULT",
+        "RESULT_CHECKSUM_MAP",
+        "RESULT_PUBLISHED_FILES",
+    )
+
+    # push_artifacts.run() reads RESULT_PUBLISHED_FILES to know where to
+    # write the published-files list.
+    os.environ["RESULT_PUBLISHED_FILES"] = str(published_path)
+
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Disagree. `push_artifacts_to_cdn.py` is the Tekton task entry point — it is explicitly designed to run inside Tekton. `tekton.result_paths()` and `RESULT_PUBLISHED_FILES` are how this task writes its Tekton results; that is their purpose. Accepting result paths as CLI flags would make this script no more standalone while adding meaningless complexity. No change made.

scripts/python/helpers/sign_mac.py
```diff
@@ -0,0 +1,273 @@
...
+            digest_file=digest_file,
+            keychain_password=keychain_password,
+            signing_identity=signing_identity,
+            apple_id=apple_id,
+            team_id=team_id,
+            app_specific_password=app_specific_password,
+        )
+
+        local_script = Path(tempfile.mktemp(suffix=".sh"))
+        local_script.write_text(script_content)
```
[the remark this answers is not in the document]
swickersh: ***Cursor on behalf of Scott*** — Fixed. Both `sign_mac.run()` and `sign_windows.run()` now use `tempfile.mkstemp()` for the local temp script files. The content is written via `os.write(fd, script_content.encode())` before closing the descriptor, so the file is created atomically with `0o600` permissions. The existing `finally: local_script.unlink(missing_ok=True)` cleanup is unchanged.

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,246 @@
...
+def run(quay_url: str) -> None:
+    """Pull signed artifacts, restore supplementary files, and compress all components."""
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    quay_user = (QUAY_SECRET_MOUNT / "username").read_text().strip()
+    quay_pass = (QUAY_SECRET_MOUNT / "password").read_text().strip()
+
+    logger.info("Logging into Quay...")
+    subprocess.check_call(
+        ["oras", "login", "quay.io", "-u", quay_user, "-p", quay_pass],
```
happybhati: Minor thing — the password ends up in the process's command line args which could be visible via /proc/PID/cmdline for the lifetime of the subprocess. `oras login` supports `--password-stdin` which avoids this. Probably low risk inside a k8s pod but easy to fix if it makes sense.

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,259 @@
...
+                        layer_file.name,
+                        image_path,
+                    )
+
+        for wanted in wanted_files:
+            src = tmp_dir / wanted
+            if src.is_file():
+                shutil.copy2(str(src), str(destination / src.name))
+            else:
+                logger.warning("Expected file not found in container: %s", wanted)
```
happybhati: Curious about this one — if a file listed in the RPA isn't found in any container layer, we warn and keep going. Is that intentional? I'm imagining a scenario where the artifact directory ends up incomplete and the failure only shows up later in compress/push with a confusing error. Or is there a case where it's valid to continue without that file?

scripts/python/helpers/compress_artifacts.py
swickersh: thanks, that's a good suggestion. Implemented it.

scripts/python/helpers/extract_artifacts.py
swickersh: This was pre-existing behavior. But it's a good call-out so I went ahead and fixed it. The script will now fail if a team declares a file for release in an RPA but it doesn't exist in the image. I think that's a reasonable thing to fail on because it suggests the release isn't consistent with what they are declaring.

scripts/python/helpers/compress_artifacts.py
happybhati: Great, LGTM!

happybhati (dismissed):

querti (dismissed):

swickersh:
rebased on https://github.com/konflux-ci/release-service-utils/pull/770 now that ubi10 and newer python are merged.

scripts/python/helpers/extract_artifacts.py
johnbieren, line 1: Should we make the name somehow cdn related? Or do you think other extraction functions will be in here eventually

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,262 @@
...
+    p.add_argument(
+        "--concurrent-limit",
+        type=int,
+        default=3,
+        help="Maximum number of components to process in parallel",
+    )
+    return p.parse_args(argv)
+
+
+def _setup_docker_config() -> None:
```
johnbieren, line 60: Isn't this same function in build_checksum_map (but implemented a bit differently)? Maybe it can go in authentication.py

scripts/python/helpers/generate_checksums.py
```diff
@@ -0,0 +1,275 @@
...
+def _sha256sum(path: Path) -> str:
+    """Return the hex SHA-256 digest of the file at path."""
+    h = hashlib.sha256()
+    with open(path, "rb") as f:
+        for chunk in iter(lambda: f.read(65536), b""):
+            h.update(chunk)
+    return h.hexdigest()
+
+
+def _kinit(checksum_user: str, kerberos_realm: str, keytab_b64: bytes) -> None:
```
johnbieren, line 71: We already have a kinit (with retry) in authentication.py

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,249 @@
...
+        for file in supp_os_dir.rglob("*"):
+            if file.is_file():
+                rel = file.relative_to(supp_os_dir)
+                dest = signed_dir / os_name / rel
+                dest.parent.mkdir(parents=True, exist_ok=True)
+                shutil.move(str(file), str(dest))
+                logger.info("  Restored supplementary file: %s/%s", os_name, rel)
+
+
+def _arch_dir(os_name: str, arch: str, component_dir: Path) -> Path | None:
```
johnbieren: This is almost identical to _os_target_dir in push_unsigned.py

scripts/python/helpers/compress_artifacts.py
```diff
@@ -0,0 +1,249 @@
...
+def _windows_filename(source_filename: str) -> str:
+    """Replace .tar.gz or .tar extension with .zip for Windows archives."""
+    if source_filename.endswith(".tar.gz"):
+        return source_filename[: -len(".tar.gz")] + ".zip"
+    if source_filename.endswith(".tar"):
+        return source_filename[: -len(".tar")] + ".zip"
+    return source_filename
+
+
+def _process_file_entry(
```
johnbieren: Maybe it is just me, and I know the imports and file separation make it okay, but it is a bit confusing to have _process_file_entr[y,ies] in two related files (both for cdn) - push_unsigned.py and compress_artifacts.py

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,245 @@
...
+                f"Archive {archive_name} contains unsupported entry type: {member.name}"
+            )
+        try:
+            tf.extract(member, path=str(target_dir), filter="data")
+        except TypeError:
+            # Python <3.12 does not support the filter argument.
+            tf.extract(member, path=str(target_dir))
+
+
+def _push_oras(tag: str, directory: Path, subdirectory: str, component_name: str) -> str:
```
johnbieren: Isn't this just pushing a string with oras? Maybe it can be made a little more generic so other tasks that oras push can use it

scripts/python/helpers/authentication.py
```diff
@@ -123,3 +123,15 @@ def _run_once() -> None:
         retry_on=subprocess.CalledProcessError,
         base_sleep_seconds=5,
     )
+
+
+def write_docker_config(config_json: str) -> None:
```
johnbieren, line 128: I think this one is missing tests

scripts/python/helpers/extract_artifacts.py
swickersh: Good point on the naming. I think keeping it generic is the right call — `release-to-github` will likely reuse this, as the original `extract-binaries` Tekton task it replaces was written by/for that team. Keeping the name broad leaves the door open for that reuse without a rename later.

***Cursor on behalf of Scott***

scripts/python/helpers/extract_artifacts.py
swickersh: Good catch. Done — I've extracted a shared `write_docker_config(config_json: str)` function into `authentication.py`. Both `extract_artifacts` and `build_checksum_map` now call it; each still handles its own pre-processing (noise stripping and the exists/non-empty guard respectively) before delegating the write.

***Cursor on behalf of Scott***

scripts/python/helpers/generate_checksums.py
swickersh: Agreed. Updated `_kinit` in `generate_checksums.py` to delegate to `authentication.kinit_with_retry` — it still owns the keytab decode, temp file lifecycle, and `KRB5CCNAME` env setup, but the actual `kinit` call with retry logic now lives only in `authentication`.

***Cursor on behalf of Scott***

scripts/python/helpers/compress_artifacts.py
swickersh: Fixed. Consolidated into `oras_utils.os_arch_dir(os_name, arch, *, mac_windows_base, linux_base)` — callers pass the appropriate base directory (e.g. `component_dir / "unsigned"` vs `component_dir / "signed"`), and the OS-to-path logic lives in one place. Both `push_unsigned` and `compress_artifacts` now call it.

***Cursor on behalf of Scott***

scripts/python/helpers/compress_artifacts.py
swickersh: Renamed to make the intent clear: `_process_file_entries` in `push_unsigned.py` → `_unpack_file_entries`, and `_process_file_entry` in `compress_artifacts.py` → `_compress_file_entry`. The names now reflect what each actually does rather than the generic "process".

***Cursor on behalf of Scott***

scripts/python/helpers/push_unsigned.py
swickersh: Good idea. Moved to a new shared `oras_utils.py` as `oras_push(tag, directory, subdirectory, component_name)`. `push_unsigned` now calls `oras_utils.oras_push` directly, and the module is available for `sign_mac`, `sign_windows`, or anything else that needs to push OCI artifacts going forward.

***Cursor on behalf of Scott***

scripts/python/helpers/authentication.py
swickersh: Added three tests to `test_authentication.py` covering: file creation with correct content, directory creation when `~/.docker` doesn't exist, and overwriting an existing `config.json`.

***Cursor on behalf of Scott***

davidmogar (dismissed):

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,169 @@
...
+
+OCI_STORE = os.environ.get("OCI_STORE", "quay.io/konflux-ci/release-service-trusted-artifacts")
+
+logger = logging.getLogger(__name__)
+
+
+def _sha256(path: Path) -> str:
+    """Return the hex SHA-256 digest of the file at path.
+
+    Note: an equivalent ``_sha256sum`` function exists in ``generate_checksums.py``.
```
johnbieren: I mean doesn't this just scream to import it instead?

scripts/python/helpers/build_checksum_map.py
```diff
@@ -0,0 +1,169 @@
...
+            tf.add(str(manifest_path), arcname="checksum_map.json")
+
+        auth_data = subprocess.check_output(
+            ["select-oci-auth", OCI_STORE], stderr=subprocess.PIPE
+        )
+        auth_file = Path(tempfile.mktemp())
+        auth_file.write_bytes(auth_data)
+
+        try:
+            oras_output = subprocess.check_output(
```
johnbieren: why isn't this using oras_utils.oras_push?

scripts/python/helpers/extract_artifacts.py
johnbieren: I'm confused. Why does only one care about checking if the file exists before reading it and the other care about cleaning json noise but not the first?

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,400 @@
...
+    p.add_argument(
+        "--cert-expiration-warn-days",
+        type=int,
+        default=7,
+        help="Days before cert expiry to warn",
+    )
+    return p.parse_args(argv)
+
+
+def _check_cert_expiration(cert_path: str, warn_days: int) -> None:
```
johnbieren, line 81: this sounds like something we want in all tasks that use certs. definitely out of scope of this PR, but we should probably convert the check_cert_expiration bash utility into python and just import and use it. so let's leave this as is for now

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,400 @@
...
+                        "filename": f.name,
+                        "relative_path": str(f.relative_to(staging_dir)),
+                        "version": version,
+                    }
+                )
+
+        staged_yaml_path = staging_dir / "staged.yaml"
+        staged_yaml_path.write_text(yaml.dump(staged_json, indent=4, default_flow_style=False))
+
+        subprocess.check_call(
```
johnbieren: this is weird to me. you are calling a python script with subprocess instead of just calling a python module natively. I think it would be better to make it so pulp_push_wrapper can just be called. I did a similar thing with apply_template.py in https://github.com/konflux-ci/release-service-utils/pull/778

scripts/python/helpers/push_artifacts.py
```diff
@@ -0,0 +1,400 @@
...
+
+    cgw_push = any(bool(c.get("contentGateway")) for c in snapshot.get("components", []))
+    if cgw_push:
+        logger.info("Publishing all components to CGW...")
+        if "developers.qa.redhat.com" in cgw_hostname:
+            os.environ["HTTP_PROXY"] = "http://squid.corp.redhat.com:3128"
+            os.environ["HTTPS_PROXY"] = "http://squid.corp.redhat.com:3128"
+            logger.info("Using squid proxy for preprod CGW access")
+
+        subprocess.check_call(
```
johnbieren: same comment about using subprocess to call python from inside a python script

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,211 @@
...
+        if member_real != target_real and target_real not in member_real.parents:
+            raise RuntimeError(f"Archive {archive_name} contains unsafe path: {member.name}")
+        if member.issym() or member.islnk() or member.isdev():
+            raise RuntimeError(
+                f"Archive {archive_name} contains unsupported entry type: {member.name}"
+            )
+        try:
+            tf.extract(member, path=str(target_dir), filter="data")
+        except TypeError:
+            # Python <3.12 does not support the filter argument.
```
johnbieren: https://github.com/konflux-ci/release-service-utils/blob/main/pyproject.toml#L6 seems we're on python 3.12 now

scripts/python/helpers/push_unsigned.py
```diff
@@ -0,0 +1,211 @@
...
+
+def run(quay_url: str, pipeline_run_uid: str) -> None:
+    """Organise extracted binaries and push unsigned Mac/Windows content to Quay."""
+    snapshot = json.loads(os.environ["SNAPSHOT_JSON"])
+
+    quay_user = (QUAY_SECRET_MOUNT / "username").read_text().strip()
+    quay_pass = (QUAY_SECRET_MOUNT / "password").read_text().strip()
+
+    logger.info("Logging into Quay...")
+    subprocess.check_call(
```
johnbieren: This and compress_artifacts.py both do oras login to quay. I think it would be better to make a function in oras_utils.py that does lines 138-143 here

scripts/python/helpers/sign_windows.py
```diff
@@ -0,0 +1,263 @@
...
+                    f"{windows_user}@{windows_host}",
+                    f"Remove-Item -LiteralPath '{cleanup_path}' -Force -Recurse; "
+                    f"Remove-Item -LiteralPath '{windows_script_path}' -Force",
+                ],
+                check=False,
+            )
+            if cleanup.returncode != 0:
+                logger.warning(
+                    "Remote cleanup failed for %s (exit code: %d) — "
+                    "%s and %s may remain on the Windows host",
```
johnbieren, line 235: we cleanup always, why would it remain on the windows host?

scripts/python/tasks/internal/test_push_artifacts_to_cdn.py
johnbieren, line 1: push_artifacts_to_cdn.py is in tasks/internal. Why is this in helpers?

scripts/python/helpers/sign_windows.py
swickersh: We always attempt cleanup, but that doesn't mean it succeeds. This just logs if it doesn't, without failing the pipeline over it.

scripts/python/tasks/internal/test_push_artifacts_to_cdn.py
swickersh: it shouldn't be. Moved. Thanks

scripts/python/helpers/push_artifacts.py
swickersh: I agree. Implemented this suggestion.

scripts/python/helpers/push_artifacts.py
swickersh: Implemented

scripts/python/helpers/push_unsigned.py
swickersh: oh yea, that's leftover from before I rebased on ubi10. 
fixed, thanks.

scripts/python/helpers/push_unsigned.py
swickersh: done.

scripts/python/helpers/build_checksum_map.py
swickersh: Moved _sha256 / _sha256sum into file_utils.sha256 in file.py and updated both build_checksum_map.py and generate_checksums.py to use it.

scripts/python/helpers/build_checksum_map.py
swickersh: fixed

scripts/python/helpers/extract_artifacts.py
swickersh: Both now go through authentication.setup_docker_config

swickersh (commented):
Moving back to draft so I can retest this set of changes.
Will mark ready again after pipelines succeed.

happybhati (dismissed):

scripts/python/helpers/authentication.py
```diff
@@ -123,3 +123,40 @@ def _run_once() -> None:
...
+    """Write *config_json* to ``~/.docker/config.json``, creating the directory if needed.
+
+    Shared by helpers that authenticate to container registries (Quay, Red Hat workloads
+    registry) before pulling or pushing OCI artifacts.  The caller is responsible for
+    obtaining the JSON string (reading a mounted secret file, stripping noise, etc.).
+    """
+    docker_dir = Path.home() / ".docker"
+    docker_dir.mkdir(parents=True, exist_ok=True)
+    (docker_dir / "config.json").write_text(config_json)
```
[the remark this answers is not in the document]
swickersh: Fixed: rewrote `write_docker_config` to create the `.docker` directory with `mode=0o700` and write `config.json` atomically using `tempfile.mkstemp` + `os.fdopen` + `os.replace`, ensuring the file is created with `0o600` permissions and never exists world-readable at any point during the write.

Cursor, on behalf of Scott

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,264 @@
...
+                text=True,
+            )
+            layer_entries = result.stdout.splitlines()
+            for image_path in extract_dirs:
+                if any(line.startswith(f"{image_path}/") for line in layer_entries):
+                    logger.info("Extracting %s/ from %s...", image_path, layer_file.name)
+                    subprocess.check_call(
+                        ["tar", "-xzvf", str(layer_file), image_path],
+                        cwd=str(tmp_dir),
+                    )
```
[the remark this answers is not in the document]
swickersh: Fixed: replaced `subprocess.check_call(["tar", "-xzvf", ...])` with a pure-Python `tarfile`-based helper (`_safe_extract_layer`). Each member is validated before extraction: path traversal is blocked by resolving the final path and confirming it falls under the target directory; symlinks, hardlinks, and device/block files are rejected outright; and `filter="data"` is applied to the extractor. The pattern mirrors the existing `_safe_extract_archive` helper already in `push_unsigned.py`.

Cursor, on behalf of Scott

scripts/python/helpers/authentication.py
```diff
@@ -123,3 +123,40 @@ def _run_once() -> None:
...
+    stripped before writing.
+    """
+    import re
+
+    if optional and (not path.is_file() or path.stat().st_size == 0):
+        return
+    raw = path.read_text(encoding="utf-8")
+    if strip_noise:
+        raw = re.sub(r"(^|\})[^{}]+(\{|$)", r"\1\2", raw)
+    write_docker_config(raw)
```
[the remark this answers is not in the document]
swickersh: Fixed: replaced the fragile `re.sub(r"(^|\})[^{}]+(\{|$)", r"\1\2", raw)` approach with explicit JSON boundary detection — find the first `{` and last `}`, slice, then validate and normalise via a `json.loads` + `json.dumps` round-trip. This correctly handles nested braces (e.g. `{"auths": {"registry": {...}}}`) which the original regex would corrupt. A test covering that case was added.

Cursor, on behalf of Scott

scripts/python/helpers/extract_artifacts.py
```diff
@@ -0,0 +1,264 @@
...
+                    "--retry-times",
+                    "3",
+                    "--authfile",
+                    str(auth_file),
+                    f"docker://{pullspec}",
+                    f"dir:{tmp_dir}",
+                ]
+            )
+        finally:
+            auth_file.unlink(missing_ok=True)
```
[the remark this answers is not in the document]
swickersh: Fixed: replaced `tempfile.mktemp()` with `tempfile.NamedTemporaryFile(mode="wb", delete=False)`. This atomically creates the temp file (internally via `mkstemp`) with `0o600` permissions and handles the file descriptor lifecycle cleanly through the context manager — no sentinel values or manual fd management needed. The file is unlinked in the `finally` block regardless of whether `select-oci-auth` or `skopeo` fails.

Cursor, on behalf of Scott

scripts/python/helpers/sign_windows.py
```diff
@@ -0,0 +1,263 @@
...
+    pipeline_run_uid: str,
+    windows_temp_dir: str,
+) -> str:
+    """Build the remote batch script that pulls, signs, verifies, and pushes binaries."""
+    return f"""
+mkdir %TEMP%\\{windows_temp_dir} && cd /d %TEMP%\\{windows_temp_dir}
+@echo off
+oras login quay.io -u {quay_user} -p {quay_pass}
+@echo on
+oras pull {quay_url}/unsigned/{component_name}@{unsigned_digest} -o unsigned
```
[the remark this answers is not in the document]
swickersh: The `mktemp()` TOCTOU concern is already addressed — the branch uses `mkstemp()` to create the local script file with `0o600` permissions, and it is deleted unconditionally in the `finally` block.

The broader concern about credentials embedded in the batch script body is acknowledged but not changed here: the credentials are required on the remote Windows host for `oras` authentication, and there is no lightweight mechanism to pass secrets out-of-band to a batch script over SSH/SCP without architectural rework out of scope for this PR. Mitigations already in place: the signing host is behind our internal network, `@echo off` precedes the `oras login` line so the password is never echoed to logs, and both the local temp file and the remote script are deleted after each run.

Cursor, on behalf of Scott

happybhati (dismissed):

johnbieren (dismissed):

ach912 (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-08T10:13:58.990Z |

Approved

swickersh:
Python ruff docstring test was failing in merge queue. Only change since approvals was to docstrings.
Please re-approve

johnbieren (approved):

PR #775, opened 2026-05-20
feat(RELEASE-1988): add standalone process-file-updates Python script

elenagerman:
## Describe your changes
Adds a standalone python script for the process-file-updates internal task logic so it can run from the utils image instead of an inline catalog script.

## Relevant Jira

Assisted-by: Claude

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     "pulp-cli==0.36.3",
     "diffused-lib==0.3.0",
     "confluent-kafka",
+    "python-gitlab>=4.0",
 ]
 
 [tool.black]
--- a/scripts/python/helpers/vcs/git.py
+++ b/scripts/python/helpers/vcs/git.py
@@ -11,6 +11,8 @@ from typing import IO, Any
 
 import retry
 
+from redact import redact_secrets
+
 
 def repository_workdir_name(repository_url: str) -> str:
     """Directory name for *repository_url* (strip `.git` suffix)."""
@@ -20,21 +22,11 @@ def repository_workdir_name(repository_url: str) -> str:
     return base
 
 
-def _redact_credential_urls(text: str) -> str:
-    """Redact oauth2/token credentials embedded in HTTPS URLs in *text*."""
-    return re.sub(
-        r"https://([^/@\s:]+):([^@\s]+)@",
-        r"https://\1:[REDACTED]@",
-        text,
-        flags=re.IGNORECASE,
-    )
-
-
 def _append_cmd_stderr(stderr_path: Path | None, message: str) -> None:
     """Append redacted *message* to *stderr_path* when configured."""
     if stderr_path is None or not message.strip():
         return
-    safe_text = _redact_credential_urls(str(message))
+    safe_text = redact_secrets(str(message))
     with open(
         stderr_path,
         "a",
@@ -256,14 +248,23 @@ def sync_to_origin_main(
 def working_tree_diff(
     repo_dir: Path,
     *,
+    cached: bool = False,
+    other_ref: str | None = None,
     stderr_path: Path | None = None,
 ) -> str:
-    """Return unstaged working-tree diff text for *repo_dir*."""
-    return _run_git_cmd(
-        ["git", "diff"],
-        cwd=repo_dir,
-        stderr_path=stderr_path,
-    ).stdout
+    """Return diff text for *repo_dir*.
+
+    By default return unstaged working-tree diff. With *cached* true, return
+    staged diff. When *other_ref* is set with *cached*, compare the index
+    against that ref.
+    """
+    cmd = ["git", "diff"]
+    if cached:
+        cmd.append("--cached")
+    if other_ref is not None:
+        cmd.append(other_ref)
+    result = _run_git_cmd(cmd, cwd=repo_dir, stderr_path=stderr_path)
+    return result.stdout or ""
 
 
 def changed_paths_from_status(
@@ -329,27 +330,82 @@ def origin_main_has_path_matching(
     return False
 
 
+def commit_staged(
+    repo_root: Path,
+    message: str,
+    *,
+    stderr_path: Path | None = None,
+) -> None:
+    """Commit staged changes via the git CLI."""
+    _run_git_cmd(
+        ["git", "commit", "-m", message],
+        cwd=repo_root,
+        stderr_path=stderr_path,
+    )
+
+
 def index_add_commit(
     repo_root: Path,
     relative_paths: Sequence[str],
     message: str,
     *,
+    commit: bool = True,
     stderr_path: Path | None = None,
 ) -> None:
-    """Stage *relative_paths* and commit with *message* via the git CLI.
+    """Stage *relative_paths* and, when *commit* is true, commit with *message*.
 
-    Call `configure_git_global_user` first so git can identify the committer.
+    Uses the git CLI. Call `configure_git_global_user` first so git can identify
+    the committer when committing.
     """
     _run_git_cmd(
         ["git", "add", *relative_paths],
         cwd=repo_root,
         stderr_path=stderr_path,
     )
+    if commit:
+        commit_staged(repo_root, message, stderr_path=stderr_path)
+
+
+def rebase_onto_remote(
+    repo_root: Path,
+    *,
+    remote_name: str,
+    remote_repository: str,
+    revision: str,
+    stderr_path: Path | None = None,
+) -> None:
+    """Add *remote_name*, fetch *revision*, and rebase onto it."""
+    remotes = _run_git_cmd(
+        ["git", "remote"],
+        cwd=repo_root,
+        stderr_path=stderr_path,
+    ).stdout.splitlines()
+    if remote_name not in remotes:
+        _run_git_cmd(
+            ["git", "remote", "add", remote_name, remote_repository],
+            cwd=repo_root,
+            stderr_path=stderr_path,
+        )
+    fetch(repo_root, remote_name, revision, stderr_path=stderr_path)
     _run_git_cmd(
-        ["git", "commit", "-m", message],
+        ["git", "rebase", f"{remote_name}/{revision}"],
         cwd=repo_root,
         stderr_path=stderr_path,
     )
+
+
+def working_tree_status(
+    repo_root: Path,
+    *,
+    stderr_path: Path | None = None,
+) -> str:
+    """Return ``git status`` output."""
+    result = _run_git_cmd(
+        ["git", "status"],
+        cwd=repo_root,
+        stderr_path=stderr_path,
+    )
+    return result.stdout or ""
 
 
 def commit_and_push(
--- a/scripts/python/helpers/vcs/gitlab.py
+++ b/scripts/python/helpers/vcs/gitlab.py
@@ -31,8 +31,8 @@ class GitLabCredentials:
 def read_credentials_from_mount(secret_mount: Path) -> GitLabCredentials:
     """Load credentials from *secret_mount*, where each field is a separate file.
 
-    `gitlab_host`, `gitlab_access_token`, `git_author_name`,
-    `git_author_email`, `git_repo`.
+    Expected files: ``gitlab_host``, ``gitlab_access_token``, ``git_author_name``,
+    ``git_author_email``, ``git_repo``.
     """
     return GitLabCredentials(
         gitlab_host=authentication.read_mounted_text(secret_mount, "gitlab_host"),
@@ -66,6 +66,21 @@ def configure_git_oauth2_auth(access_token: str) -> None:
     os.environ["GIT_TERMINAL_PROMPT"] = "0"
     os.environ["GIT_ASKPASS"] = str(askpass)
     os.environ["GITLAB_OAUTH2_TOKEN"] = access_token
+
+
+def gitlab_project_path(repository: str) -> str:
+    """Normalize *repository* to a ``group/project`` path for the GitLab API."""
+    repo = repository.strip()
+    if "://" in repo:
+        path = repo.split("://", 1)[1]
+        if "/" in path:
+            path = path.split("/", 1)[1]
+        path = path.strip("/")
+    else:
+        path = repo.strip("/")
+    if path.endswith(".git"):
+        path = path[:-4]
+    return path
 
 
 def raw_file_url(
--- a/scripts/python/helpers/vcs/test_git.py
+++ b/scripts/python/helpers/vcs/test_git.py
@@ -42,14 +42,6 @@ def test_append_cmd_stderr_redacts_oauth2_url(tmp_path: Path) -> None:
     assert "oauth2:[REDACTED]@" in text
 
 
-def test_redact_credential_urls() -> None:
-    """Mask oauth2 tokens in arbitrary text (stderr or argv)."""
-    text = "fatal: https://oauth2:secret@gitlab.com/g/r.git"
-    redacted = git._redact_credential_urls(text)
-    assert "secret" not in redacted
-    assert "oauth2:[REDACTED]@" in redacted
-
-
 def test_run_git_cmd_redacts_clone_failure_log(tmp_path: Path) -> None:
     """Failed clone must not write credential-bearing URLs to the stderr log."""
     log = tmp_path / "log.txt"
@@ -360,6 +352,80 @@ def test_index_add_commit(tmp_path: Path) -> None:
         git.index_add_commit(tmp_path, ["a.yaml"], "msg", stderr_path=None)
     assert run_cmd.call_count == 2
     assert run_cmd.call_args_list[1].args[0] == ["git", "commit", "-m", "msg"]
+
+
+def test_index_add_commit_stage_only(tmp_path: Path) -> None:
+    """When ``commit=False``, paths are staged but not committed."""
+    with mock.patch.object(git, "_run_git_cmd") as run_cmd:
+        git.index_add_commit(tmp_path, ["a.yaml"], "", commit=False)
+    run_cmd.assert_called_once_with(
+        ["git", "add", "a.yaml"],
+        cwd=tmp_path,
+        stderr_path=None,
+    )
+
+
+def test_working_tree_diff_cached(tmp_path: Path) -> None:
+    """Return cached diff output from the git CLI."""
+    with mock.patch.object(
+        git,
+        "_run_git_cmd",
+        return_value=mock.MagicMock(stdout="diff\n"),
+    ):
+        assert git.working_tree_diff(tmp_path, cached=True) == "diff\n"
+
+
+def test_working_tree_diff_cached_with_other_ref(tmp_path: Path) -> None:
+    """Compare cached diff against another ref when requested."""
+    with mock.patch.object(
+        git,
+        "_run_git_cmd",
+        return_value=mock.MagicMock(stdout="diff\n"),
+    ) as run_cmd:
+        assert git.working_tree_diff(tmp_path, cached=True, other_ref="mr_1") == "diff\n"
+    assert run_cmd.call_args.args[0] == ["git", "diff", "--cached", "mr_1"]
+
+
+def test_rebase_onto_remote_adds_missing_remote(tmp_path: Path) -> None:
+    """Add remote when missing, fetch, and rebase."""
+    calls: list[list[str]] = []
+
+    def _cmd_side_effect(cmd: list[str], **_kwargs: object) -> mock.MagicMock:
+        calls.append(cmd)
+        if cmd[:2] == ["git", "remote"] and len(cmd) == 2:
+            return mock.MagicMock(stdout="origin\n")
+        return mock.MagicMock(stdout="")
+
+    with mock.patch.object(git, "_run_git_cmd", side_effect=_cmd_side_effect):
+        git.rebase_onto_remote(
+            tmp_path,
+            remote_name="glab-base",
+            remote_repository="https://gitlab.com/g/up.git",
+            revision="main",
+        )
+    assert ["git", "remote", "add", "glab-base", "https://gitlab.com/g/up.git"] in calls
+    assert ["git", "fetch", "glab-base", "main"] in calls
+    assert ["git", "rebase", "glab-base/main"] in calls
+
+
+def test_rebase_onto_remote_skips_existing_remote(tmp_path: Path) -> None:
+    """Skip ``git remote add`` when the remote already exists."""
+    calls: list[list[str]] = []
+
+    def _cmd_side_effect(cmd: list[str], **_kwargs: object) -> mock.MagicMock:
+        calls.append(cmd)
+        if cmd[:2] == ["git", "remote"] and len(cmd) == 2:
+            return mock.MagicMock(stdout="glab-base\n")
+        return mock.MagicMock(stdout="")
+
+    with mock.patch.object(git, "_run_git_cmd", side_effect=_cmd_side_effect):
+        git.rebase_onto_remote(
+            tmp_path,
+            remote_name="glab-base",
+            remote_repository="https://gitlab.com/g/up.git",
+            revision="main",
+        )
+    assert not any(cmd[:3] == ["git", "remote", "add"] for cmd in calls)
 
 
 def test_commit_and_push(tmp_path: Path) -> None:
--- a/scripts/python/tasks/internal/process_file_updates.py
+++ b/scripts/python/tasks/internal/process_file_updates.py
@@ -1,0 +1,730 @@
+#!/usr/bin/env python3
+"""Apply file edits to a GitLab repository and open a merge request with the changes.
+
+Intended to be invoked by a release pipeline to automatically propagate release
+artifacts (e.g. updated image digests, version strings) into downstream
+configuration repositories, creating an MR for human review.
+
+* Reads credentials from ``/mnt/file-updates-secret/`` (or
+  ``FILE_UPDATES_SECRET_MOUNT``): ``gitlab_host``, ``gitlab_access_token``,
+  ``git_author_name``, ``git_author_email``.
+* Clones ``--repo`` at ``--ref``, rebases on ``--upstream-repo``, applies path
+  seeds and YAML replacements (``yq``/``sed``), then commits or reuses an MR.
+* Writes ``RESULT_FILE_UPDATES_*`` and internal-request name results.
+* After a valid run with result env vars, almost always exits ``0``; failures use
+  ``RESULT_FILE_UPDATES_STATE``. Invalid YAML on a replacement target exits ``1``.
+  Bad CLI flags exit before result handling (``1``; argparse uses ``2`` for
+  malformed argv).
+
+Pass ``gitlab_client`` to ``run_file_updates`` to inject a client in unit tests.
+Catalog Tekton tests mock ``git`` via ``tests/mocks/git`` on ``PATH``.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import re
+import subprocess
+import sys
+import uuid
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import gitlab
+from gitlab.exceptions import GitlabError
+
+import file
+import redact
+import tekton
+from logger import logger
+from vcs import git as vcs_git
+from vcs import gitlab as vcs_gitlab
+
+PROG = "process_file_updates.py"
+FILE_UPDATES_SECRET_MOUNT_ENV = "FILE_UPDATES_SECRET_MOUNT"
+_REPLACEMENT_EXPRESSION = re.compile(r"^\|([^|\n]*)\|([^|\n]*)\|$")
+
+
+def _usage_text() -> str:
+    """Return the short usage summary printed to stderr on bad CLI usage."""
+    return (
+        f"usage: {PROG} --upstream-repo REPO --repo REPO --ref REF --paths JSON \\\n"
+        f"  --component-group GROUP --internal-request-pipeline-run-name NAME \\\n"
+        f"  --internal-request-task-run-name NAME [--temp-dir DIR]\n"
+    )
+
+
+def parse_args(argv: list[str] | None) -> argparse.Namespace:
+    """Parse required CLI flags; help or missing values exit ``1`` (extras exit ``2``)."""
+    p = argparse.ArgumentParser(prog=PROG, add_help=False, usage=argparse.SUPPRESS)
+    p.add_argument("-h", "--help", action="store_true")
+    p.add_argument("--upstream-repo", metavar="REPO")
+    p.add_argument("--repo", metavar="REPO")
+    p.add_argument("--ref", metavar="REF")
+    p.add_argument("--paths", metavar="JSON")
+    p.add_argument("--component-group", metavar="GROUP")
+    p.add_argument("--internal-request-pipeline-run-name", metavar="NAME")
+    p.add_argument("--internal-request-task-run-name", metavar="NAME")
+    p.add_argument("--temp-dir", metavar="DIR")
+    ns = p.parse_args(argv or [])
+    if ns.help:
+        print(_usage_text(), file=sys.stderr, end="")
+        raise SystemExit(1)
+    required = (
+        "upstream_repo",
+        "repo",
+        "ref",
+        "paths",
+        "component_group",
+        "internal_request_pipeline_run_name",
+        "internal_request_task_run_name",
+    )
+    if any(not getattr(ns, f) or not str(getattr(ns, f)).strip() for f in required):
+        print(_usage_text(), file=sys.stderr, end="")
+        raise SystemExit(1)
+    return ns
+
+
+def load_file_updates_secrets(mount: Path) -> dict[str, str]:
+    """Read git/GitLab secret files under *mount*."""
+    keys = ("gitlab_host", "gitlab_access_token", "git_author_name", "git_author_email")
+    out: dict[str, str] = {}
+    for key in keys:
+        path = mount / key
+        try:
+            out[key] = path.read_text(encoding="utf-8").strip()
+        except OSError as e:
+            raise tekton.CheckStepError(f"reading secret file {key}", e) from e
+    return out
+
+
+def git_functions_init(author_name: str, author_email: str, token: str) -> None:
+    """Validate identity fields and set global git user.name and user.email."""
+    if not author_name or not author_email or not token:
+        raise tekton.CheckStepError(
+            "initializing git",
+            ValueError("GIT_AUTHOR_NAME, GIT_AUTHOR_EMAIL, and ACCESS_TOKEN are required"),
+        )
+    vcs_git.configure_git_global_user(author_name, author_email)
+
+
+def gitlab_create_mr(
+    *,
+    head: str,
+    title: str,
+    target_branch: str,
+    description: str,
+    upstream_repo: str,
+    gitlab_client: gitlab.Gitlab,
+) -> str:
+    """Create a merge request; return JSON ``{"merge_request": url}`` for Tekton results."""
+    project_path = vcs_gitlab.gitlab_project_path(upstream_repo)
+    try:
+        project = gitlab_client.projects.get(project_path)
+        mr = project.mergerequests.create(
+            {
+                "source_branch": head,
+                "target_branch": target_branch,
+                "title": title,
+                "description": description,
+            }
+        )
+    except GitlabError as e:
+        raise tekton.CheckStepError("creating GitLab merge request", e) from e
+
+    url = getattr(mr, "web_url", None) or ""
+    if url:
+        return json.dumps({"merge_request": url})
+
+    raise tekton.CheckStepError(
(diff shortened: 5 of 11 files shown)


mmalina:
The commit message and description would be suitable for the catalog change, not this one. This change is adding the standalone script that will be used in the task, it is not changing the task. Can you fix that?

elenagerman:
> The commit message and description would be suitable for the catalog change, not this one. This change is adding the standalone script that will be used in the task, it is not changing the task. Can you fix that?

done

theflockers:
I think it should have pytests as in this PR for instance  https://github.com/konflux-ci/release-service-utils/pull/765/

elenagerman:
> I think it should have pytests as in this PR for instance #765

according to the acceptance criteria: "Unit tests for the script in the utils repo is not part of this. There will be a separate ticket for that"

swickersh:
> I think it should have pytests as in this PR for instance #765

Yea the Jira said tests would be in a separate ticket. I do think they would make code review a little easier and give us more confidence in merging these changes. 

I guess even after code is merged here in utils, it isn't actually used until the catalog invokes it.  Is the plan that tests will be required before invoking these new scripts from the catalog @johnbieren ?

johnbieren:
> > I think it should have pytests as in this PR for instance #765
> 
> Yea the Jira said tests would be in a separate ticket. I do think they would make code review a little easier and give us more confidence in merging these changes.
> 
> I guess even after code is merged here in utils, it isn't actually used until the catalog invokes it. Is the plan that tests will be required before invoking these new scripts from the catalog @johnbieren ?

yeah we should've updated the tickets. now that we have AI, the tests should be added alongside the script (AI should be able to easily generate you some anyways)

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,634 @@
...
+    git_rebase_upstream("glab-base", upstream_repo, revision, token, repo_cwd)
+
+    replacements_update_error: str | None = None
+    diff_path: Path | None = None
+    replacements_performed: int | None = None
+    key_not_found = False
+
+    print(update_paths_file.read_text(encoding="utf-8"), flush=True)
+
+    for path_index, entry in enumerate(paths_data):
```
swickersh: path_index looks unused

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,634 @@
...
+        check=False,
+    )
+    if check and proc.returncode != 0:
+        raise subprocess.CalledProcessError(
+            proc.returncode, ["git", *args], proc.stdout, proc.stderr
+        )
+    return proc.stdout or ""
+
+
+def run_glab(
```
swickersh: python gitlab library might be cleaner to use: https://python-gitlab.readthedocs.io/en/stable/

this is non-blocking, but it's probably what I would do.

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,634 @@
...
+            "--target-branch",
+            target_branch,
+            "--description",
+            description,
+            "-R",
+            upstream_repo,
+        ],
+        cwd=repo_cwd,
+    )
+    for line in out.splitlines():
```
swickersh: Cursor suggested this to me as I was reviewing your change. This is part of the reason below I was thinking the python gitlab library might be good to import. 

```
This depends on the glab CLI's human-readable output format. If glab changes its output (adds color 
codes, changes the line format, etc.), this silently returns "{}" — which causes a silent success with no MR 
URL recorded. Consider using glab mr create --output json if the glab version supports it, or at minimum
 add a warning log when the fallback "{}" is returned.


Also, building f"merge_request: {url}" as a yq input string without quoting the URL is potentially fragile if 
the URL contains characters yq treats specially.

```

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,634 @@
...
+        )
+        mr_page = [ln for ln in out.splitlines() if ln.startswith("!")]
+        if not mr_page:
+            break
+        items.extend(mr_page)
+        page += 1
+    return items
+
+
+def run_file_updates(
```
swickersh: run_file_updates is ~200 lines and handles setup, per-path processing, and MR creation all in one place. That scope will make the follow-up test ticket significantly harder. Worth extracting a few focused helpers before tests are written rather than after.

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,634 @@
+#!/usr/bin/env python3
+"""Update files in Git repositories and open GitLab merge requests when needed.
```
swickersh: The description is pretty vague. 

Can you please add another sentence about real-world how/why it is used?

swickersh (changes requested):
I'd recommend adding doc strings. I received that feedback on a recent python PR of mine and agree they are helpful.

scripts/python/tasks/internal/process_file_updates.py
elenagerman: done

scripts/python/tasks/internal/process_file_updates.py
elenagerman: Good point!
I agree python-gitlab would be cleaner for the GitLab side than glab.
But IMO to keep glab in this PR to match the old task (/home/utils/gitlab-functions) and keep RELEASE-1988 scoped to “extract the script, same behavior.” 
Downsides of switching now: new dependency and image rebuild, catalog mocks/tests need rework, and we’d still use git subprocess for clone/rebase/commit/push anyway, so it’s only part of the script. 
WDYT?

scripts/python/tasks/internal/process_file_updates.py
elenagerman: good catch!
agree
modified

scripts/python/tasks/internal/process_file_updates.py
elenagerman: removed

scripts/python/tasks/internal/process_file_updates.py
elenagerman: Good point!
I split run_file_updates into helpers so the test ticket can target smaller units.

theflockers:
> > I think it should have pytests as in this PR for instance #765
> 
> Yea the Jira said tests would be in a separate ticket. I do think they would make code review a little easier and give us more confidence in merging these changes.
> 
> I guess even after code is merged here in utils, it isn't actually used until the catalog invokes it. Is the plan that tests will be required before invoking these new scripts from the catalog @johnbieren ?

That is correct, only when the change in the catalog is made that it will be used. But the script itself should contain its tests, that are not directly related to the task. For instance, if my script create a "formatted date", I should attest that that function does what it says. This has nothing to do with the actual usage of the script, meaning, I don't need to process a file update in this test, but I need to make sure the function that replaces A with B works.

elenagerman:
> > > I think it should have pytests as in this PR for instance #765
> > 
> > 
> > Yea the Jira said tests would be in a separate ticket. I do think they would make code review a little easier and give us more confidence in merging these changes.
> > I guess even after code is merged here in utils, it isn't actually used until the catalog invokes it. Is the plan that tests will be required before invoking these new scripts from the catalog @johnbieren ?
> 
> That is correct, only when the change in the catalog is made that it will be used. But the script itself should contain its tests, that are not directly related to the task. For instance, if my script create a "formatted date", I should attest that that function does what it says. This has nothing to do with the actual usage of the script, meaning, I don't need to process a file update in this test, but I need to make sure the function that replaces A with B works.

test was added

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,901 @@
...
+    true and the command fails.
+    """
+    if os.environ.get(USE_SHELL_ENV):
+        cmd = "git " + " ".join(shlex.quote(a) for a in args)
+        proc = _shell_wrap(cmd, cwd)
+        if check and proc.returncode != 0:
+            raise subprocess.CalledProcessError(
+                proc.returncode, ["bash", "-c", cmd], proc.stdout, proc.stderr
+            )
+        return proc.stdout or ""
```
theflockers: It might worth it check if the token won't leak to stdout or stderr in case of success or error.

scripts/python/tasks/internal/process_file_updates.py
swickersh: >“extract the script, same behavior.”

But you didn't just extract the script, you extracted AND converted to Python. I'm not sure importing a library is a behavioral change.

>Downsides of switching now: new dependency and image rebuild

image rebuild? The utils image is rebuilt and added to the task anyway

>catalog mocks/tests need rework

That all happens anyway with this extraction and convert to Python change

>we’d still use git subprocess for clone/rebase/commit/push 

Yes, fair point. But those are `git` actions rather Gitlab api interactions.

I wont block your PR on this... it just crossed my mind as a potentially cleaner approach. Seems like we should use libraries when they exist.  But I could see valid reasons against the suggestion as well. :smile:

scripts/python/tasks/internal/process_file_updates.py
swickersh: Thank you!

scripts/python/tasks/internal/process_file_updates.py
elenagerman: good point!
this check was added (+ test)

swickersh:
This mostly looks good to me. My only remaining concern is less than 50% test coverage. I don't know if we have a team standard, but I'd think we want to aim for 90% +

scripts/python/tasks/internal/process_file_updates.py
elenagerman: I switched to the python-gitlab

elenagerman:
> This mostly looks good to me. My only remaining concern is less than 50% test coverage. I don't know if we have a team standard, but I'd think we want to aim for 90% +

more tests were added; should be ~99% now

swickersh:
edit - I had Cursor generate some sample snippets and comment here.  Let me know if it's not helpful and I can try to explain differently

The fake client classes (`_FakeMergeRequest`, `_FakeMergeRequestManager`, `_FakeProject`, `_FakeProjectsManager`, `_CatalogFakeGitlab`) and the `USE_SHELL_ENV` branch in `connect_gitlab` need to be removed. Test infrastructure has no place in production code — this module should have no knowledge of how it will be tested.

---

Consider accepting the GitLab client as an optional parameter instead of relying on a global and a side-effectful init call. This makes the function trivially testable without any special env vars or fakes in the module:

```python
def run_file_updates(*, upstream_repo, repo, ..., gitlab_client=None):
    if gitlab_client is None:
        gitlab_client = gitlab.Gitlab(gitlab_api_url(host), private_token=token)
    ...
```

The utils unit tests can then just pass `mock.Mock()` directly.

---

When the catalog task is updated to invoke this script, `mocks.sh` can shadow the `gitlab` Python module via PYTHONPATH — same concept as overriding `glab` as a bash function today, just at the Python module level:

```bash
mkdir -p /tmp/fake_gitlab/gitlab
cat > /tmp/fake_gitlab/gitlab/__init__.py << 'EOF2'
from .exceptions import GitlabError

class Gitlab:
    def __init__(self, url, private_token=None):
        self.projects = _Projects()

class _Projects:
    def get(self, path):
        return _Project(path)
# etc.
EOF2
cat > /tmp/fake_gitlab/gitlab/exceptions.py << 'EOF2'
class GitlabError(Exception): pass
EOF2
export PYTHONPATH="/tmp/fake_gitlab:${PYTHONPATH}"
```

`USE_SHELL_ENV` can stay in `run_git` for bash function injection of `git`, but nothing else.

***Cursor on behalf of Scott***

elenagerman:
> edit - I had Cursor generate some sample snippets and comment here. Let me know if it's not helpful and I can try to explain differently
> 
> The fake client classes (`_FakeMergeRequest`, `_FakeMergeRequestManager`, `_FakeProject`, `_FakeProjectsManager`, `_CatalogFakeGitlab`) and the `USE_SHELL_ENV` branch in `connect_gitlab` need to be removed. Test infrastructure has no place in production code — this module should have no knowledge of how it will be tested.
> 
> Consider accepting the GitLab client as an optional parameter instead of relying on a global and a side-effectful init call. This makes the function trivially testable without any special env vars or fakes in the module:
> 
> ```python
> def run_file_updates(*, upstream_repo, repo, ..., gitlab_client=None):
>     if gitlab_client is None:
>         gitlab_client = gitlab.Gitlab(gitlab_api_url(host), private_token=token)
>     ...
> ```
> 
> The utils unit tests can then just pass `mock.Mock()` directly.
> 
> When the catalog task is updated to invoke this script, `mocks.sh` can shadow the `gitlab` Python module via PYTHONPATH — same concept as overriding `glab` as a bash function today, just at the Python module level:
> 
> ```shell
> mkdir -p /tmp/fake_gitlab/gitlab
> cat > /tmp/fake_gitlab/gitlab/__init__.py << 'EOF2'
> from .exceptions import GitlabError
> 
> class Gitlab:
>     def __init__(self, url, private_token=None):
>         self.projects = _Projects()
> 
> class _Projects:
>     def get(self, path):
>         return _Project(path)
> # etc.
> EOF2
> cat > /tmp/fake_gitlab/gitlab/exceptions.py << 'EOF2'
> class GitlabError(Exception): pass
> EOF2
> export PYTHONPATH="/tmp/fake_gitlab:${PYTHONPATH}"
> ```
> 
> `USE_SHELL_ENV` can stay in `run_git` for bash function injection of `git`, but nothing else.
> 
> _**Cursor on behalf of Scott**_

Fixed, if I understand you correctly.
all fake client classes were removed; we use an optional gitlab_client on run_file_updates and related helpers; if none is passed, it builds one with make_gitlab_client(host, token). 
And the utils tests pass mock.Mock() directly.

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,784 @@
...
+import gitlab
+from gitlab.exceptions import GitlabError
+
+import file
+import tekton
+
+PROG = "process_file_updates.py"
+FILE_UPDATES_SECRET_MOUNT_ENV = "FILE_UPDATES_SECRET_MOUNT"
+SHELL_INIT_ENV = "PROCESS_FILE_UPDATES_SHELL_INIT"
+USE_SHELL_ENV = "PROCESS_FILE_UPDATES_USE_SHELL"
```
swickersh: There's one more thing that needs to come out: USE_SHELL_ENV, SHELL_INIT_ENV, and _shell_wrap. These exist so the catalog integration tests can inject bash mock functions for git, but they're the same problem as the fake GitLab classes — production code modified to accommodate a test harness.

The catalog already has the right pattern for this, established by check-embargoed-cves-task. Any file placed under tests/mocks/<name> is automatically written as an executable and prepended to PATH by [render_python_task_mocks_from_yaml.py](https://github.com/konflux-ci/release-service-catalog/blob/3173224fd75c5244d16b33e50a2e07ace2c9e2ba/.github/scripts/render_python_task_mocks_from_yaml.py). So when the catalog task is updated, a tests/mocks/git fake script handles it — Python's subprocess.run(["git", ...]) finds the fake via PATH with no special handling in the production code at all.

And the utils unit tests that currently set USE_SHELL=1 can mock run_git directly with mock.patch instead.

elenagerman: removed

scripts/python/tasks/internal/process_file_updates.py
theflockers: great solution.

swickersh (dismissed):

theflockers (dismissed):

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,760 @@
...
+
+import file
+import tekton
+
+PROG = "process_file_updates.py"
+FILE_UPDATES_SECRET_MOUNT_ENV = "FILE_UPDATES_SECRET_MOUNT"
+_OAUTH_URL_IN_TEXT = re.compile(r"(https?://)oauth2:[^@\s\"']+@", re.IGNORECASE)
+
+
+def redact_secrets(text: str) -> str:
```
johnbieren: I think this would fit nicely in logger.py so (pretty much all) python scripts can use it

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,760 @@
...
+    for key in keys:
+        path = mount / key
+        try:
+            out[key] = path.read_text(encoding="utf-8").strip()
+        except OSError as e:
+            raise tekton.CheckStepError(f"reading secret file {key}", e) from e
+    return out
+
+
+def run_git(args: Sequence[str], *, cwd: Path | None = None, check: bool = True) -> str:
```
johnbieren: idk who will get to merge first but I am adding an entire git library here https://github.com/konflux-ci/release-service-utils/pull/778

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,760 @@
...
+        text=True,
+        capture_output=True,
+        check=False,
+    )
+    if check and proc.returncode != 0:
+        _raise_checked_subprocess(proc, ["git", *args])
+    return proc.stdout or ""
+
+
+def gitlab_api_url(host: str) -> str:
```
johnbieren: gitlab module also in https://github.com/konflux-ci/release-service-utils/pull/778

johnbieren (commented):
I started to review it but there is a lot of git stuff in here. I understand it isn't merged yet so there is no way for you to know, but I am adding a lot of git stuff in https://github.com/konflux-ci/release-service-utils/pull/778 (and I have some more in my update-infra-deployments draft). In any case, I don't think the git stuff fits well in process_file_updates.py since a lot of our tasks use git. So, you can either wait for 778 to merge and then just import the git stuff here, or I suppose you could duplicate it and if you merge first I can resolve the conflict in 778

scripts/python/tasks/internal/process_file_updates.py
elenagerman: sure, moved to the logger.py

scripts/python/tasks/internal/process_file_updates.py
elenagerman: removed, using your solution (copied to my pr)
thank you!

scripts/python/tasks/internal/process_file_updates.py
elenagerman: absolutely!!!
thank you @johnbieren !!!

elenagerman:
> I started to review it but there is a lot of git stuff in here. I understand it isn't merged yet so there is no way for you to know, but I am adding a lot of git stuff in #778 (and I have some more in my update-infra-deployments draft). In any case, I don't think the git stuff fits well in process_file_updates.py since a lot of our tasks use git. So, you can either wait for 778 to merge and then just import the git stuff here, or I suppose you could duplicate it and if you merge first I can resolve the conflict in 778

No doubt I need to wait for you to merge.
In the meantime, I copied your code locally to check the correct usage.
Thank you!
Set this PR as a Draft until #778 is merged

elenagerman:
@johnbieren I removed the prod/needs-approval label as well to avoid confusion

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,720 @@
...
+    replaced_block_lines = len(result_path.read_text(encoding="utf-8").splitlines())
+    if replaced_block_lines != value_size + 1:
+        return "Text block size differs from the original", diff_path
+
+    grep_proc = subprocess.run(
+        ["sed", "-ne", f"{start_block},+{value_size}p", str(target_file)],
+        text=True,
+        capture_output=True,
+        check=True,
+    )
```
swickersh: I think you've already captured this above.
Maybe delete grep_proc and in replaced_count below do: 

```
replaced_count = sum(
    1 for line in result_path.read_text(encoding="utf-8").splitlines()
    if replace_str in line
)
```

swickersh (commented):
This mostly lgtm.

I think the description at the top of process_file_updates.py accurately describes the _mechanism_ of what the script does but not really the _purpose_. 

How would you feel about something like this:

```
Apply file edits to a GitLab repository and open a merge request with the changes.

Intended to be invoked by a release pipeline to automatically propagate release
artifacts (e.g. updated image digests, version strings) into downstream
configuration repositories, creating an MR for human review.
```

elenagerman:
/retest

> This mostly lgtm.
> 
> I think the description at the top of process_file_updates.py accurately describes the _mechanism_ of what the script does but not really the _purpose_.
> 
> How would you feel about something like this:
> 
> ```
> Apply file edits to a GitLab repository and open a merge request with the changes.
> 
> Intended to be invoked by a release pipeline to automatically propagate release
> artifacts (e.g. updated image digests, version strings) into downstream
> configuration repositories, creating an MR for human review.
> ```

sure, added

scripts/python/tasks/internal/process_file_updates.py
elenagerman: good catch! thank you!
I missed it...
fixed

scripts/python/helpers/vcs/git.py
```diff
@@ -98,15 +98,36 @@ def clone(
...
     directory_name: str | None = None,
     revision: str,
-    sparse_dirs: Sequence[str],
+    sparse_dirs: Sequence[str] | None = None,
     stderr_path: Path | None = None,
 ) -> Path:
-    """Shallow clone *clone_url* with sparse checkout of *sparse_dirs* at *revision*.
+    """Shallow clone *clone_url* at *revision*.
+
+    When *sparse_dirs* is set, use sparse checkout of those paths. When ``None``,
```
mmalina: When Johnny was adding this, I asked if maybe there are cases where we need the full history and we weren't sure. So I wonder, is this needed here? Could the task work with sparse checkout as well? If so, that would be preferable.

scripts/python/helpers/vcs/git.py
```diff
@@ -170,21 +191,125 @@ def index_add_commit(
...
     )
+    if commit:
+        _run_git_cmd(
+            ["git", "commit", "-m", message],
+            cwd=repo_root,
+            stderr_path=stderr_path,
+        )
+
+
+def commit(
```
mmalina: Wait, you add a commit function, but then you copy and paste the content above? in the `if commit` block? Why? If we have a function for it now, why not use it?

scripts/python/helpers/vcs/gitlab.py
```diff
@@ -68,6 +69,34 @@ def configure_git_oauth2_auth(access_token: str) -> None:
     os.environ["GITLAB_OAUTH2_TOKEN"] = access_token
 
 
+def gitlab_api_url(host: str) -> str:
```
mmalina: Do we really need this? Isn't it clearly defined what host will look like? And if we don't need this, then we don't need make_gitlab_client either. But we probably don't need that function anyway I'd say.

scripts/python/helpers/logger.py
```diff
@@ -13,3 +15,16 @@
 _handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
 logger.addHandler(_handler)
 logger.propagate = False
+
+_OAUTH_URL_IN_TEXT = re.compile(r"(https?://)oauth2:[^@\s\"']+@", re.IGNORECASE)
+
+
+def redact_secrets(text: str) -> str:
+    """Redact ``ACCESS_TOKEN`` and ``oauth2:<token>@`` URLs in *text*."""
+    if not text:
```
mmalina: What is the purpose of this?

scripts/python/helpers/logger.py
```diff
@@ -13,3 +15,16 @@
...
+        return text
+    out = _OAUTH_URL_IN_TEXT.sub(r"\1oauth2:***@", text)
+    token = os.environ.get("ACCESS_TOKEN", "")
+    if token:
```
mmalina:
```suggestion
    token = os.environ.get("ACCESS_TOKEN")
    if token is not None:
```

scripts/python/helpers/logger.py
```diff
@@ -13,3 +15,16 @@
...
+        out = out.replace(token, "***")
```
mmalina: Claude suggests this:
```
# Escape special regex characters in the token
          escaped_token = re.escape(token)
          out = re.sub(escaped_token, "***", out)
```

swickersh (commented):
lgtm other than Martin's remaining comments

scripts/python/helpers/logger.py
elenagerman: I don’t think we need the re.escape / re.sub change. We’re using .replace(), which already does literal substring replacement, so regex escaping isn’t required.

scripts/python/helpers/logger.py
elenagerman: Partially agree; not the "**is not None**" part.
Dropping the "" default makes sense. 
**if token is not None** is the wrong guard here. 
For redaction we want “is there a real token value?”, not “was the variable defined?”. 
**if token**: matches that.

scripts/python/helpers/logger.py
elenagerman: to skip redaction when there’s nothing to process (empty string, or other falsy input). No regex or ACCESS_TOKEN replacement needed, and we avoid edge cases if callers pass empty error output.

scripts/python/helpers/vcs/gitlab.py
elenagerman: make sense
gitlab_host in the secret is already the API base URL 
both gitlab_api_url and make_gitlab_client dropped

scripts/python/helpers/vcs/git.py
elenagerman: sure
refactored to reuse

scripts/python/helpers/vcs/git.py
elenagerman: good point!
git.py already supports sparse checkout, so I left it unchanged.
But I modified file-updates to build the sparse dir list from the --paths json and use that instead.

scripts/python/helpers/logger.py
mmalina: What does "a real token value" mean and how does "if token" match that? An env var is a string. How can you say what string is ok as access token and which one is not? And a string will evaluate as true as long as it's not empty. So my proposal was to guard against unset env var (hence the `is not None`).

scripts/python/helpers/vcs/git.py
mmalina: ok, so then you don't need to modify this function at all for now?

scripts/python/helpers/logger.py
mmalina: What is "other falsy input"? It's a string. So the only way this condition is true is an empty string. So if we want to skip further processing, maybe just return "" explicitly?
```
if not text:
    return ""
```

scripts/python/helpers/vcs/git.py
elenagerman: I added the "if sparse_dirs is None:" block because clone() used to be sparse-only. file-updates can use sparse when paths are known, but other callers may still need a full tree, so one function handles both.

elenagerman: and this block is in use. 
file-updates normally passes sparse dirs from the paths json and, if none can be derived, it passes None and gets a full shallow clone. 
it’s the fallback and the “full tree” option for callers that need it

scripts/python/helpers/logger.py
elenagerman: Fair point! 
we're not deciding what counts as a valid access token. We only need to know if there's a non-empty string to search for in the error text. 
**if token:** means "skip when there's nothing to replace", that covers unset (None) and empty (""). 
**is not None** would still run .replace("", "***") when the var is set to an empty string, which corrupts the output.

elenagerman: But I dropped the "" default

scripts/python/helpers/logger.py
elenagerman: yea, good idea, it could work as well
modified
thank you!

scripts/python/helpers/vcs/git.py
mmalina: ok, so you need normal clone for the process-file-updates use case? because sometimes the paths are not known? in that case it's fine. but if you're adding this option for somebody else down the line, then let's remove it.

scripts/python/helpers/logger.py
mmalina: I don't quite understand the use case of ACCESS_TOKEN being set, but empty. But ok...

scripts/python/helpers/vcs/git.py
elenagerman: removed

mmalina (dismissed):

scripts/python/helpers/vcs/git.py
```diff
@@ -329,27 +329,110 @@ def origin_main_has_path_matching(
...
     _run_git_cmd(
         ["git", "add", *relative_paths],
         cwd=repo_root,
         stderr_path=stderr_path,
     )
+    if commit:
+        commit_staged(repo_root, message, stderr_path=stderr_path)
+
+
+def checkout_new_branch(
```
johnbieren: `checkout` already does this

scripts/python/helpers/vcs/git.py
```diff
@@ -329,27 +329,110 @@ def origin_main_has_path_matching(
...
+    """Create and switch to a new branch via the git CLI."""
     _run_git_cmd(
-        ["git", "commit", "-m", message],
+        ["git", "checkout", "-b", branch],
+        cwd=repo_root,
+        stderr_path=stderr_path,
+    )
+
+
+def staged_diff(
```
johnbieren: We have an existing diff function - `working_tree_diff` can we amend that one to include the use case this adds?

scripts/python/helpers/logger.py
```diff
@@ -13,3 +15,16 @@
 _handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
 logger.addHandler(_handler)
 logger.propagate = False
+
+_OAUTH_URL_IN_TEXT = re.compile(r"(https?://)oauth2:[^@\s\"']+@", re.IGNORECASE)
+
+
+def redact_secrets(text: str) -> str:
```
johnbieren: Doesn't `_redact_credential_urls` in `git.py` already do this?

pyproject.toml
```diff
@@ -24,6 +24,7 @@ dependencies = [
     "pulp-cli==0.36.3",
     "diffused-lib==0.3.0",
     "confluent-kafka",
+    "python-gitlab>=4.0",
```
johnbieren, line 27: Is this still needed or a leftover from a previous version? It looks like you use git cli throughout now

scripts/python/helpers/logger.py
elenagerman: Partially: _redact_credential_urls covers credential URLs in git stderr, but it’s private to git.py and doesn’t redact a bare ACCESS_TOKEN value in plain error text. 
current method does both (URLs + token substring). 
but I see now that there’s overlap with tekton._redact_for_tekton_result too
happy to consolidate into one shared helper if you prefer; file-updates could also switch to write_failure_result instead of hand-rolling this.

pyproject.toml
elenagerman: yep, still needed. 
git cli handles clone/commit/push/rebase; python-gitlab is for the GitLab REST API (list MRs, create MR). 
Those aren’t covered by the git CLI helpers.
but I probably see a possible leftover: GitPython>=3.1.40 on line 8. I don’t see direct GitPython usage in the Python code anymore (everything goes through vcs/git.py CLI). 
I could remove that one. WDYT?

scripts/python/helpers/vcs/git.py
elenagerman: good catch!
removed + extended working_tree_diff() with cached=True

scripts/python/helpers/vcs/git.py
elenagerman: ah, sorry, I missed we have it already.
sure, removed
thank you!

pyproject.toml
elenagerman: no, you know, I would prefer to open a separate Jira for this cleanup (GitPython>=3.1.40 deletion)
WDYT?

scripts/python/helpers/logger.py
johnbieren: Yeah the goal is to reuse as much as possible. I certainly don't think we should have two versions of a function just because one is currently private. It is less to maintain to have one function to redact tokens in strings

pyproject.toml
johnbieren: thanks, I didn't realize process_file_updates.py was doing gitlab things, I only looked in gitlab.py for it. Resolving

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,731 @@
...
+        text=True,
+        capture_output=True,
+        check=False,
+    )
+    if proc.returncode != 0 or not proc.stdout.strip():
+        return 0
+    return int(proc.stdout.strip().splitlines()[0])
+
+
+_REPLACEMENT_EXPRESSION = re.compile(r"^\|([^|\n]*)\|([^|\n]*)\|$")
```
johnbieren: shouldn't constants be at the top of the file?

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,731 @@
...
+    print(seed, flush=True)
+    if not seed:
+        return
+    print("seed operation to perform", flush=True)
+    target_file.parent.mkdir(parents=True, exist_ok=True)
+    target_file.write_text(
+        seed + ("\n" if not seed.endswith("\n") else ""),
+        encoding="utf-8",
+    )
+    print("-- start targetFile --", flush=True)
```
johnbieren: should we use logger functions instead of print?

elenagerman: yes and no
Most prints I used intentionally mirror the legacy bash script output for Tekton logs (file dumps, FOUND/NOT FOUND, diff banners). 
And switching those to logger.info would add INFO: prefixes and change what operators see in CI. 
But I can change it if we want consistency with other tasks.

scripts/python/tasks/internal/process_file_updates.py
elenagerman: good catch! thank you!
do not have to, but kind of a good practice.
I put it here for better visibility during debugging.
moved now to the top with the other module constants.

scripts/python/helpers/logger.py
elenagerman: perfect!
So all three are now consolidated into helpers/redact.py. git, tekton, and file-updates all use the same redact_secrets() now; removed the private copies from logger.py, git.py, and tekton.py.
WDYT?

scripts/python/tasks/internal/process_file_updates.py
johnbieren: I don't think adding `INFO:` prefixes to the task logs is a reason to not use python logging like the other tasks. I'd vote for changing it

elenagerman: sure, modified

scripts/python/helpers/vcs/test_git.py
```diff
@@ -44,8 +44,10 @@ def test_append_cmd_stderr_redacts_oauth2_url(tmp_path: Path) -> None:
 
 def test_redact_credential_urls() -> None:
```
johnbieren: This can be removed now, right?

scripts/python/helpers/logger.py
johnbieren: Excellent, thanks!

scripts/python/tasks/internal/process_file_updates.py
```diff
@@ -0,0 +1,730 @@
...
+    )
+
+
+def list_konflux_open_mrs(upstream_repo: str, gitlab_client: gitlab.Gitlab) -> list[Any]:
+    """List open MRs in *upstream_repo* matching ``Konflux release`` (paginated)."""
+    project_path = vcs_gitlab.gitlab_project_path(upstream_repo)
+    try:
+        project = gitlab_client.projects.get(project_path)
+    except GitlabError as e:
+        raise tekton.CheckStepError("listing GitLab merge requests", e) from e
```
johnbieren: Can this one be `getting GitLab project` or something instead so we can differentiate it from line 167?

elenagerman: good point!
sure, fixed

scripts/python/helpers/vcs/test_git.py
elenagerman: absolutely!
removed

johnbieren (approved):

twaugh (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-18T14:43:51.209Z |

PR #778, opened 2026-05-21
feat(RELEASE-1985): add create_advisory python script

johnbieren:
This commit adds a python script to replicate the functionality of the inline bash script in the create-advisory internal task in the catalog repo. The task will be updated to use this python module instead.

It also updates some helper modules so that future converted tasks can use them instead of it being inside the create_advisory module. Finally, it reworks apply_template.py a bit from the utils dir so that the create_advisory python script can call it natively instead of using subprocess to call apply_template.py.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/advisory_data.py
+++ b/scripts/python/helpers/advisory_data.py
@@ -1,0 +1,307 @@
+"""Decode advisory task payloads and content filtering (idempotency rules)."""
+
+from __future__ import annotations
+
+import base64
+import gzip
+import json
+import re
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+
+def _strip_checksum_from_purl(purl: str) -> str:
+    """Strip `checksum` query/fragment parts from a package URL for comparison."""
+    # purl checksum may appear as `&checksum=` (extra param), `?checksum=` in
+    # the query, or trailing `?checksum=` before end — strip all three shapes.
+    stripped = re.sub(r"&checksum=[^&]*", "", purl)
+    stripped = re.sub(r"\?checksum=[^&]*&", "?", stripped)
+    stripped = re.sub(r"\?checksum=[^&]*$", "", stripped)
+    return stripped
+
+
+def _filter_image(content: list[Any], existing: list[Any]) -> list[Any]:
+    """Drop image rows that already exist (same image, tags, repository)."""
+    # `tags` is compared as JSON lists (order and length must match).
+    out: list[Any] = []
+    for item in content:
+        if not isinstance(item, dict):
+            continue
+        container_image = item.get("containerImage")
+        tags = item.get("tags")
+        repo = item.get("repository")
+        is_duplicate = False
+        for existing_row in existing:
+            if not isinstance(existing_row, dict):
+                continue
+            if (
+                existing_row.get("containerImage") == container_image
+                and existing_row.get("tags") == tags
+                and existing_row.get("repository") == repo
+            ):
+                is_duplicate = True
+                break
+        if not is_duplicate:
+            out.append(item)
+    return out
+
+
+def _filter_rpm(content: list[Any], existing: list[Any]) -> list[Any]:
+    """Drop artifact rows whose `purl` exactly matches an existing row."""
+    existing_purls = {
+        existing_row.get("purl")
+        for existing_row in existing
+        if isinstance(existing_row, dict) and existing_row.get("purl") is not None
+    }
+    out: list[Any] = []
+    for item in content:
+        if not isinstance(item, dict):
+            continue
+        purl = item.get("purl")
+        # Without `purl` there is nothing to compare; skip the row.
+        if purl is None or purl not in existing_purls:
+            out.append(item)
+    return out
+
+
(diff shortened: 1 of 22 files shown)


scripts/python/helpers/vcs/gitlab.py
```diff
@@ -0,0 +1,90 @@
...
+
+def oauth2_https_clone_url(https_repository_url: str, oauth2_token: str) -> str:
+    """
+    Build an HTTPS clone URL with embedded token for GitLab.
+
+    Uses username `oauth2`, which is the GitLab convention for project tokens.
+    """
+    host_and_path = https_repository_url.split("://", 1)[1]
+    return f"https://oauth2:{oauth2_token}@{host_and_path}"
+
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/helpers/vcs/git.py
```diff
@@ -0,0 +1,156 @@
...
+            # GitPython logs push rejections but does not raise unless we check
+            # the returned PushInfo list.
+            repo.remotes.origin.push().raise_if_error()
+            return
+        except GitCommandError as push_error:
+            _append_git_stderr(stderr_path, push_error)
+            attempt += 1
+            try:
+                repo.git.pull("--rebase", "origin", branch)
+            except GitCommandError as pull_error:
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/tasks/internal/create_advisory.py
```diff
@@ -0,0 +1,603 @@
...
+        [
+            "kubectl",
+            "get",
+            "configmap",
+            config_map_name,
+            "-o",
+            "jsonpath={.data.SIG_KEY_NAME}",
+        ],
+        stderr_path=stderr_path,
+    ).stdout.strip()
```
elenagerman, line 274: If the configmap key is missing, the task succeeds with an invalid advisory.
IMO it should fail fast on an empty signing key

scripts/python/helpers/advisory_data.py
```diff
@@ -0,0 +1,307 @@
...
+        sort_keys=False,
+        allow_unicode=True,
+    )
+
+
+def spec_content_json_pointer(content_type: str) -> str:
+    """Return the dotted *content_list_path* for *content_type* (under `spec` in YAML)."""
+    if content_type == "image":
+        return ".content.images"
+    if content_type in ("binary", "generic", "rpm"):
```
elenagerman, line 240: shouldn't a "disk-image" option be added to the list? wdyt?

johnbieren: I hesitate to say no, but we don't have it here https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/create-advisory-task/create-advisory-task.yaml#L177

scripts/python/tasks/internal/create_advisory.py
johnbieren: fixed

johnbieren:
Addressed comments and reworked the git stuff a bit to have easier reusability in future tasks. Sorry to change it up during review

scripts/python/helpers/advisory_data.py
elenagerman: disk-image isn't listed in the internal task parameter description, but managed create-advisory (https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/internal/create-advisory-task/create-advisory-task.yaml#L241) includes it.
It is definitely a bug in the original code, but I'm not sure what the fix should be: handle it (add to the list) or reject it as it is now.
Just asking... not a blocker
/LGTM

elenagerman (approved):

scripts/python/helpers/advisory_data.py
johnbieren: Ah, nice catch. I'll file a followup PR for that since it isn't really in scope of RELEASE-1985 and this is a large enough change as is

mmalina:
update-fbc-catalog IR repeatedly failed.

Found one: https://console-openshift-console.apps.rosa.kflux-c-stg-i01.qfla.p3.openshiftapps.com/k8s/ns/internal-services/tekton.dev~v1~PipelineRun/internalrequest-2x5v8/
Says logs no longer available.

johnbieren:
> update-fbc-catalog IR repeatedly failed.
> 
> Found one: https://console-openshift-console.apps.rosa.kflux-c-stg-i01.qfla.p3.openshiftapps.com/k8s/ns/internal-services/tekton.dev~v1~PipelineRun/internalrequest-2x5v8/ Says logs no longer available.

https://github.com/konflux-ci/release-service-utils/pull/796 will make it so the logs show, as they do for catalog PRs

davidmogar (approved):

ach912 (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-02T14:55:03.180Z |

Approved

PR #780, opened 2026-05-25
feat(RELEASE-2419): add --append-tags to create_container_image

mmalina:
Add --append-tags parameter to create_container_image script. When set to true, the script preserves existing tags in Pyxis and adds new ones instead of replacing them. This enables releasing the same image for multiple product versions while maintaining all version tags.

- Add construct_tags() function to handle both normal and append modes
- Modify construct_repository() to accept pre-constructed tag dicts
- In append mode, new/re-applied tags get current date, existing-only tags preserve their original dates
- Add comprehensive unit and integration tests

Assisted-by: Claude Code

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python3
-"""
-Python script to create a Container Image object in Pyxis
+"""Python script to create a Container Image object in Pyxis.
 
 NOTE: Any change to this file that changes functionality should be posted for review in
 #forum-metadata-guild on slack. No PR changing functionality should be removed without
@@ -64,11 +63,10 @@ MANIFEST_LIST_TYPES = [
 
 
 def setup_argparser() -> Any:  # pragma: no cover
-    """Setup argument parser
+    """Set up argument parser.
 
     :return: Initialized argument parser
     """
-
     parser = argparse.ArgumentParser(description="ContainerImage resource creator.")
 
     parser.add_argument(
@@ -174,11 +172,10 @@ def proxymap(repository: str) -> str:
 
 
 def find_image(pyxis_url, digest: str) -> Any:
-    """Function to find a containerImage with the given digest.
+    """Find a containerImage with the given digest.
 
     :return: the image, if one exists, else None if none found
     """
-
     raw_filter = f'repositories.manifest_schema2_digest=="{digest}";not(deleted==true)'
     # quote is needed to urlparse the quotation marks
     filter_str = quote(raw_filter)
@@ -202,7 +199,7 @@ def find_image(pyxis_url, digest: str) -> Any:
 
 
 def find_repo_in_image(repository_str: str, image: Dict[str, Any]) -> Optional[int]:
-    """Check if a repository already exists in the ContainerImage
+    """Check if a repository already exists in the ContainerImage.
 
     :return: index of repository if repository_str string is found in the ContainerImage
         repositories, None otherwise
@@ -214,16 +211,16 @@ def find_repo_in_image(repository_str: str, image: Dict[str, Any]) -> Optional[i
 
 
 def prepare_parsed_data(args) -> Dict[str, Any]:
-    """Function to extract the data this script needs to create parsed_data.
-       Processes data from:
-        - Architecture from args
-        - ORAS manifest fetch output
-        - Dockerfile content (if provided)
-        - Metadata file with env_variables and labels (if provided)
+    """Extract the data this script needs to create parsed_data.
+
+    Processes data from:
+     - Architecture from args
+     - ORAS manifest fetch output
+     - Dockerfile content (if provided)
+     - Metadata file with env_variables and labels (if provided)
 
     :return: Dict containing processed data from parsed sources
     """
-
     with open(args.oras_manifest_fetch) as json_file:
         oras_manifest_fetch = json.load(json_file)
 
@@ -276,7 +273,7 @@ def prepare_parsed_data(args) -> Dict[str, Any]:
 
 
 def pyxis_tags(tags, date_now):
-    """Return list of tags formatted for pyxis"""
+    """Return list of tags formatted for pyxis."""
     return [
         {
             "added_date": date_now,
@@ -319,7 +316,7 @@ def construct_tags(
 
 
 def repository_digest_values(args):
-    """Return digest values for the repository entry in the image entity"""
+    """Return digest values for the repository entry in the image entity."""
     result = {"manifest_schema2_digest": args.architecture_digest}
     if args.media_type in MANIFEST_LIST_TYPES:
         result["manifest_list_digest"] = args.digest
@@ -327,8 +324,7 @@ def repository_digest_values(args):
 
 
 def create_container_image(args, parsed_data: Dict[str, Any], tags: List[str]):
-    """Function to create a new containerImage entry in a pyxis instance"""
-
+    """Create a new containerImage entry in a pyxis instance."""
     LOGGER.info("Creating new container image")
 
     constructed_tags = construct_tags(tags)
@@ -372,6 +368,7 @@ def create_container_image(args, parsed_data: Dict[str, Any], tags: List[str]):
 def update_container_image_repositories(
     pyxis_url, image_id: str, repositories: List[Dict[str, Any]]
 ):
+    """Update repositories for a container image."""
     LOGGER.info(f"Updating repositories for container image {image_id}")
 
     patch_url = urljoin(pyxis_url, f"v1/images/id/{image_id}")
@@ -444,8 +441,7 @@ def construct_repository(args, tag_dicts):
 
 
 def main():  # pragma: no cover
-    """Main func"""
-
+    """Execute main function."""
     parser = setup_argparser()
     args = parser.parse_args()
     log_level = logging.DEBUG if args.verbose else logging.INFO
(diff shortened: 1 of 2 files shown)


FilipNikolovski (dismissed):
LGTM

elenagerman:
this change is undocumented?

pyxis/create_container_image.py
```diff
@@ -436,18 +479,25 @@ def main():  # pragma: no cover
...
         )
-        repositories.append(construct_repository(args, tags))
+        constructed_tags = construct_tags(tags)
+        repositories.append(construct_repository(args, constructed_tags))
         update_container_image_repositories(args.pyxis_url, identifier, repositories)
         return
 
     # Then, check if the tags are different. If they are, update them
-    existing_tags = [tag["name"] for tag in repositories[repo_index].get("tags") or []]
+    existing_tags = [tag["name"] for tag in repositories[repo_index].get("tags", [])]
```
elenagerman, line 484: what is the chance that pyxis and the cli can hold the same tags in a different order? 
It could cause an unnecessary patch, no?

mmalina:
> this change is undocumented?

What do you mean by this? There is a description in the arg parser. We don't have anything better for most of the other args either.

If you mean from a Konflux user perspective, this utils change is a prerequisite for a catalog change which will come once this is merged. Without the catalog change, this option is not exposed to users yet.

pyxis/create_container_image.py
mmalina: Yeah, looking at the check (`if existing_tags != tags:`) I agree that if the order is different, we will still call the patch request. But it seems it's always been like that. And in reality, it seems pretty unlikely that this would happen - like somebody would need to reorder the tags in the RPA. And even then, I don't see much harm in that. So probably ok? But if you think it should be fixed, I can do it.

pyxis/create_container_image.py
```diff
@@ -382,7 +425,7 @@ def construct_repository(args, tags):
             "registry": registry,
             "repository": proxymap(image_name),
             "push_date": date_now,
-            "tags": pyxis_tags(tags, date_now),
+            "tags": tags,
```
elenagerman: the callers pass dicts, not a tag strings. Maybe it is worth to rename it to the tag_dicts?

elenagerman:
No tests for order-insensitive skip, append no-op skip, --is-latest + append, or --rh-push + append

elenagerman:
> > this change is undocumented?
> 
> What do you mean by this? There is a description in the arg parser. We don't have anything better for most of the other args either.
> 
> If you mean from a Konflux user perspective, this utils change is a prerequisite for a catalog change which will come once this is merged. Without the catalog change, this option is not exposed to users yet.

Sorry, small misunderstanding here. I didn't mean it has no documentation; I meant only that the in-code comment might be worth adding.
Something like: "append ignored when adding a new repo entry" or "--append-tags only applies when updating tags on an existing repo entry".

mmalina:
> No tests for order-insensitive skip, append no-op skip, --is-latest + append, or --rh-push + append

By order-insensitive skip, you mean that it will not skip if the order of tags is different, right? I don't think we need a test for that - from the perspective of the script, in this scenario the tags are just different.

For the rest, do we really need a test for every combination? Or which function do you mean? If main, it previously didn't have any tests at all and now it has some.

pyxis/create_container_image.py
mmalina: you're right, probably better to rename it. changed.

mmalina:
> > > this change is undocumented?
> > 
> > 
> > What do you mean by this? There is a description in the arg parser. We don't have anything better for most of the other args either.
> > If you mean from a Konflux user perspective, this utils change is a prerequisite for a catalog change which will come once this is merged. Without the catalog change, this option is not exposed to users yet.
> 
> Sorry, small misunderstanding here. I didn't mean it has no documentation; I meant only that the in-code comment might be worth adding. Something like: "append ignored when adding a new repo entry" or "--append-tags only applies when updating tags on an existing repo entry".

ok, but where would this fit? The arg description already says:

> If set to true, when an image with the given repository already exists, "
>        "missing tags will be added to the existing tags instead of replacing them. "

So this implies that if it doesn't exist, there is no effect.

elenagerman:
> By order-insensitive skip, you mean that it will not skip if the order of tags is different, right? I don't think we need a test for that - from the perspective of the script, in this scenario the tags are just different.
> 
> For the rest, do we really need a test for every combination? Or which function do you mean? If main, it previously didn't have any tests at all and now it has some.

I mean the opposite: when tag names match but order differs, we skip the PATCH (replace mode only), because order isn’t meaningful. 
But probably you can skip it since you are sure in the list equality (https://github.com/konflux-ci/release-service-utils/pull/780#discussion_r3304554067)

elenagerman:
> > > > this change is undocumented?
> > > 
> > > 
> > > What do you mean by this? There is a description in the arg parser. We don't have anything better for most of the other args either.
> > > If you mean from a Konflux user perspective, this utils change is a prerequisite for a catalog change which will come once this is merged. Without the catalog change, this option is not exposed to users yet.
> > 
> > 
> > Sorry, small misunderstanding here. I didn't mean it has no documentation; I meant only that the in-code comment might be worth adding. Something like: "append ignored when adding a new repo entry" or "--append-tags only applies when updating tags on an existing repo entry".
> 
> ok, but where would this fit? The arg description already says:
> 
> > If set to true, when an image with the given repository already exists, "
> > "missing tags will be added to the existing tags instead of replacing them. "
> 
> So this implies that if it doesn't exist, there is no effect.

before https://github.com/konflux-ci/release-service-utils/pull/780/changes#diff-5aa26df8e75364957237d2f2a2a2ec79c73df4748ecdb0630423b01de155394cR482
But you're right, the arg help already covers it and the in-code comment is an optional maintainer hint, not a documentation gap; we can drop it if you insist this is unnecessary.
this is not a merge blocker from my side, just proposal...

elenagerman:
I don't see any merge blocker in the code that have to be fixed, so /LGTM

elenagerman (dismissed):

mmalina:
> > By order-insensitive skip, you mean that it will not skip if the order of tags is different, right? I don't think we need a test for that - from the perspective of the script, in this scenario the tags are just different.
> > For the rest, do we really need a test for every combination? Or which function do you mean? If main, it previously didn't have any tests at all and now it has some.
> 
> I mean the opposite: when tag names match but order differs, we skip the PATCH (replace mode only), because order isn’t meaningful. But probably you can skip it since you are sure in the list equality ([#780 (comment)](https://github.com/konflux-ci/release-service-utils/pull/780#discussion_r3304554067))

I am confused. You brought up that because we just check for list equality, order actually does matter. And so if the tag names are the same, but in different order, the result is that it's treated as a new set of tags and the patch request is done. And it seems you were fine with keeping it that way. But now you say we skip the patch in this scenario. I don't think so.

mmalina:
> > > > > this change is undocumented?
> > > > 
> > > > 
> > > > What do you mean by this? There is a description in the arg parser. We don't have anything better for most of the other args either.
> > > > If you mean from a Konflux user perspective, this utils change is a prerequisite for a catalog change which will come once this is merged. Without the catalog change, this option is not exposed to users yet.
> > > 
> > > 
> > > Sorry, small misunderstanding here. I didn't mean it has no documentation; I meant only that the in-code comment might be worth adding. Something like: "append ignored when adding a new repo entry" or "--append-tags only applies when updating tags on an existing repo entry".
> > 
> > 
> > ok, but where would this fit? The arg description already says:
> > > If set to true, when an image with the given repository already exists, "
> > > "missing tags will be added to the existing tags instead of replacing them. "
> > 
> > 
> > So this implies that if it doesn't exist, there is no effect.
> 
> before https://github.com/konflux-ci/release-service-utils/pull/780/changes#diff-5aa26df8e75364957237d2f2a2a2ec79c73df4748ecdb0630423b01de155394cR482 But you're right, the arg help already covers it and the in-code comment is an optional maintainer hint, not a documentation gap; we can drop it if you insist this is unnecessary. this is not a merge blocker from my side, just proposal...

I am fine with adding it, but where? where do you think this would fit?

johnbieren (dismissed):

ach912 (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-08T10:13:49.929Z |

Approved

mmalina:
I needed to fix docstrings because of the new docstring check. It's the second commit - nothing changed about the code change.

johnbieren (approved):

ach912 (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-11T10:06:57.490Z |

Re-approve

mmalina:
omg, this again: 1: T1 Title exceeds max length (75>72): "feat(RELEASE-2419): add --append-tags flag to create_container_image (#780)"

Elena created a bug for this: https://redhat.atlassian.net/browse/RELEASE-2663

mmalina:
But also, all of e2e failed for some reason.

mmalina:
I fixed the lingering gitlint issue. but now it failed on a couple of e2e tests in merge queue.

PR #784, opened 2026-05-27
feat(RELEASE-1989): add Python implementation for publish-index-image

midnightercz:
… tasks

Implements managed and internal publish-index-image tasks with comprehensive skopeo integration and testing infrastructure:

- Add managed_publish_index_image task that processes IR results and spawns parallel internal Tekton requests for image publishing
- Add internal publish_index_image task that copies index images using skopeo
- Implement skopeo wrapper module with inspect, copy, and login operations
- Add internal_request module for creating and managing Tekton internal requests
- Implement rsmodels for internal request and secret data models
- Add fake skopeo implementation for testing without registry access
- Include comprehensive test coverage for all new modules
- Update Dockerfile and dependencies for Python task execution


seanconroy2021:
Can you update the commit to use Assisted-by: instead of Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com> ? So the label AI CI will work.

midnightercz:
> Can you update the commit to use Assisted-by: instead of Co-Authored-By: Claude Sonnet 4.5 [noreply@anthropic.com](mailto:noreply@anthropic.com) ? So the label AI CI will work.

fixed

scripts/python/helpers/skopeo.py
```diff
@@ -0,0 +1,346 @@
...
+        if dest_no_creds:
+            cmd.append("--dest-no-creds")
+        if dest_cert_dir is not None:
+            cmd.extend(["--dest-cert-dir", dest_cert_dir])
+        if dest_authfile is not None:
+            cmd.extend(["--dest-authfile", dest_authfile])
+
+        cmd.extend([source, destination])
+        if self.logger:
+            self.logger.info(f"Running skopeo copy command: {cmd}")
```
seanconroy2021: unveil() is called at lines 321 and 333 so wont the command contain raw credentials here ?

scripts/python/tasks/managed/managed_publish_index_image.py
```diff
@@ -0,0 +1,237 @@
...
+                pipeline_timeout=pipeline_timeout,
+            )
+            ir_futures[ft] = (target_index, source_index, pi)
+
+    # Process all internal request futures
+    for ft in as_completed(ir_futures):
+        target_index, source_index, pi = ir_futures[ft]
+        try:
+            internal_request_id = ft.result()
+            ir = _get_internal_request(internal_request_id)
```
seanconroy2021: This will only creates the CR and returns the name it doesn't wait. The bash IR script waits by default. The get_internal_request call fetches the IR right after creation so it'll always be still Running with no results yet. So do we not need a wait_for_internal_request between create and get else the status/results checks below won't do noting right.

scripts/python/tasks/internal/publish_index_image.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    parser.add_argument(
+        "--target-index",
+        required=True,
+        help="Target image pull spec (e.g., registry.io/image:tag)",
+    )
+    parser.add_argument(
+        "--retries",
+        type=int,
+        default=3,
```
seanconroy2021, line 192: This is never passed to client.inspect() or client.copy() both support retry_times but neither uses it. The bash task passed retry.

scripts/python/tasks/managed/managed_publish_index_image.py
```diff
@@ -0,0 +1,237 @@
+#!/usr/bin/env python3
```
seanconroy2021, line 1: from __future__ import annotations is missing here and in publish_index_image.py. Other task script in the repo has it e.g. check_embargoed_cves.py, check_fbc_opt_in.py.

scripts/python/tasks/internal/publish_index_image.py
```diff
@@ -0,0 +1,218 @@
...
+            logger=logger,
+        )
+
+        # Step 9: Write result
+        write_result(message)
+
+        return 0 if success else 1
+
+    except Exception as e:
+        logger.exception("Failed to publish index image: {e}")
```
seanconroy2021: This is a plain string. Should be f"Failed to publish index image: {e}"

scripts/python/tasks/managed/managed_publish_index_image.py
```diff
@@ -0,0 +1,237 @@
...
+import logging
+import re
+import traceback
+import sys
+from pathlib import Path
+
+import jq
+
+# Add helpers to path
+helpers_path = str(Path(__file__).parent.parent / "helpers")
```
seanconroy2021: I dont really understand why this is needed scripts/python/helpers is already in PYTHONPATH in the Dockerfile and in pythonpath in pyproject.toml so the imports work without it. Same thing in tests/test_managed_publish_index_image.py line 10.

scripts/python/tasks/managed/managed_publish_index_image.py
```diff
@@ -0,0 +1,237 @@
...
+    sys.path.insert(0, helpers_path)
+from internal_request import (  # noqa: E402
+    create_internal_request as _create_internal_request,
+    get_internal_request as _get_internal_request,
+)
+from logger import setup_logger  # noqa: E402
+from decorators import async_in_executor  # noqa: E402
+
+# Create process pool executor for parallel execution
+executor = ProcessPoolExecutor()
```
seanconroy2021: Is there a reason this needs to be ProcessPoolExecutor? The work is all I/O so ThreadPoolExecutor would be simpler here no ?

scripts/python/tasks/managed/managed_publish_index_image.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def _extract(in_query: str, in_file: str):
```
seanconroy2021: No need to use jq here that made sense in the bash version but in Python you can just do json.load() the file once and work with the dict directly then json.dump(). Right now it re-opens and re-parses the same file.

scripts/python/helpers/skopeo.py
midnightercz: yeah, you're right. Client isn't written correctly.

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: That's not needed for python >= 3.10

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: Ah, SYNC is set to true on the top, I overlooked it. But I found this to be little strange in publish-index-image task

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: yeah, maybe, I just chose it by default as it's not limited by GIL. Maybe lets try ThreadPoolExecutor first as there were problems with OOM caused by python scripts

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: this was originally written as helper so user can pass any query string here. It's not needed for this specific use case though. So yeah, lets make it simpler

scripts/python/tasks/internal/publish_index_image.py
midnightercz: good catch

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: it was failing on my local host, but now I realized you can use uv to run pytest and then it respect pyproject. So yeah, I'll remove it

scripts/python/helpers/skopeo.py
midnightercz: fixed

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: fixed

scripts/python/tasks/internal/publish_index_image.py
midnightercz: fixed

scripts/python/tasks/internal/publish_index_image.py
midnightercz: fixed

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: fixed

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: fixed

scripts/python/tasks/managed/managed_publish_index_image.py
midnightercz: fixed

scripts/python/helpers/skopeo.py
```diff
@@ -0,0 +1,354 @@
+"""Python client for skopeo container image operations."""
```
FilipNikolovski: Should we perhaps create a separate PR for the skopeo helper and prioritize the review? This implementation looks the most promising from a first glance, so it would be nice if we could have this sooner rather than later, since there's a lot of tasks which use this tool.

midnightercz: sure

midnightercz: https://github.com/konflux-ci/release-service-utils/pull/812. I'm going to be on PTO for next 2 weeks. So feel free to merge the PR

seanconroy2021 (dismissed):
LGTM - Once Fitz's comment is resolved :)

seanconroy2021:
@midnightercz what is the status of this PR? Can it be moved to draft if you're  not working on it ?

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+    Returns:
+        A tuple containing a boolean indicating success and a message string.
+
+    """
+    logger.info(f"Copying image from {source_index} to {target_index}")
+    kwargs: dict[str, Any] = {
+        "all": True,
+        "preserve_digests": True,
+        "src_tls_verify": False,
+        "dest_creds": target_credential,
```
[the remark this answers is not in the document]
midnightercz: that's expected

src/helpers/skopeo/skopeo.py
```diff
@@ -18,6 +19,8 @@ def inspect(
         cmd.append("--config")
     if raw:
         cmd.append("--raw")
+    if creds:
+        cmd.extend(["--creds", creds])
     cmd.append(f"docker://{image_ref}")
```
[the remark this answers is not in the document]
midnightercz: that's expected

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+
+
+def _wait_for_ir_wrapper(name):
+    """Wrap wait_for_completion for async executor.
+
+    Args:
+        name: Name of the InternalRequest to wait for.
+
+    """
+    _wait_for_completion(name=name)
```
querti: This uses the helper default timeout of 600s. Catalog pipeline timeout is `requestTimeout + 300` (default 660). Pass `timeout=args.request_timeout + SPAWN_OVERHEAD_SECONDS`.

_Created with Cursor._

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+            log.exception(f"{traceback.format_tb(e.__traceback__)}")
+            raise
+
+    # Process wait results and check outcomes
+    for ft in as_completed(wait_futures):
+        ir_name, pi = wait_futures[ft]
+        try:
+            ft.result()
+            results = _fetch_results(ir_name)
+        except InternalRequestWaitError:
```
querti: A failed/timed-out IR often has an empty `requestMessage`, so the `"error" in message` check below treats this as success. Treat `InternalRequestWaitError` as failure regardless of the result body. Bash checks `status.conditions[].status == False` first.

_Created with Cursor._

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+        # Step 4: Inspect source image config to get ocp version
+        source_config = inspect_image(
+            args.source_index,
+            source_creds,
+            config=True,
+            logger=logger,
+            retry_times=args.retries,
+        )
+        source_ocp_version = (
+            source_config["config"]["Labels"].get("com.redhat.component.ocp-version")
```
querti: Bash/IIB use `com.redhat.index.delivery.version` from `skopeo inspect --config`. This label is not on FBC index images, so `source_ocp_version` is always `None`, the check always fails, and we never copy. Tests mock the same wrong label so they stay green. Also prefix the message with `Error:` — the managed task greps `requestMessage` for `"error"`, and this string would look like success.

_Created with Cursor._

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def write_result(message: str) -> None:
+    """Write result message to stdout.
+
+    Args:
+        message: The message to write to stdout.
+
+    """
+    print(message)
```
querti: This needs to write `$(results.requestMessage.path)` (via `tekton.result_paths_from_env("RESULT_REQUEST_MESSAGE")`), not stdout. Bash tees to the result file; IR `status.results.requestMessage` will otherwise be empty.

_Created with Cursor._

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+    )
+    parser.add_argument(
+        "--target-index",
+        required=True,
+        help="Target image pull spec (e.g., registry.io/image:tag)",
+    )
+    parser.add_argument(
+        "--retries",
+        type=int,
+        default=3,
```
querti: Catalog default is `"0"`.

_Created with Cursor._

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+                ("targetOcpVersion", str(ocp_version)),
+            ]
+            labels = [(PIPELINERUN_LABEL, args.pipeline_run_id)]
+
+            log.debug(
+                f"Creating internal request for component {i} "
+                f"with params: {params} and labels: {labels}"
+            )
+
+            ft = create_internal_request(
```
querti, line 190: `internal_request.create()` always runs `cleanup_existing_requests()` for this pipeline + pipelinerun-uid before creating. Firing several of these in parallel (or even sequentially with `sync=False`) will delete sibling IRs. Bash waits for each IR before creating the next, which is why cleanup is safe there. Cleanup once up front, then use `build_payload` + `create_internal_request` (the low-level one), or stay sequential like `request_advisory_creation`.

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
querti, line 219: A failed/timed-out IR often has an empty `requestMessage`, so the `"error" in message` check below treats this as success. Treat `InternalRequestWaitError` as failure regardless of the result body. Bash checks `status.conditions[].status == False` first.

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+    parser.add_argument(
+        "--request-timeout",
+        type=int,
+        default=360,
+        help="Timeout for each individual request in seconds.",
+    )
+    parser.add_argument(
+        "--retries",
+        type=int,
+        default=3,
```
querti: Catalog default for `retries` is `"0"`.

src/tasks/internal/publish_index_image/publish_index_image.py
querti, line 245: Bash/IIB use `com.redhat.index.delivery.version` from `skopeo inspect --config`. This label is not on FBC index images, so `source_ocp_version` is always `None`, the check always fails, and we never copy. Tests mock the same wrong label so they stay green. Also prefix the message with `Error:` — the managed task greps `requestMessage` for `"error"`, and this string would look like success.

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+            args.target_index,
+            target_credential,
+            logger=logger,
+            retry_times=args.retries,
+        )
+        target_digest = target_image_info["Digest"] if target_image_info else None
+
+        # Hotfix and pre-ga targetIndex should skip the next check
+        if target_image_info and re.match(r".*:v[0-9]{1}\.[0-9]{2}$", args.target_index):
+            real_target_ocp_version = target_image_info["Labels"].get(
```
querti, line 269: Same wrong label, and this is a non-`--config` inspect. Bash uses `--config` and `.config.Labels."com.redhat.index.delivery.version"`. Prefix with `Error:` as well.

src/tasks/internal/publish_index_image/publish_index_image.py
querti, line 170: This needs to write `$(results.requestMessage.path)` (via `tekton.result_paths_from_env("RESULT_REQUEST_MESSAGE")`), not stdout. Bash tees to the result file; IR `status.results.requestMessage` will otherwise be empty.

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+        help="Path to source registry credential file",
+    )
+    parser.add_argument(
+        "--target-credential-path",
+        default="/mnt/publishingCredentials/targetIndexCredential",
+        help="Path to target registry credential file",
+    )
+    parser.add_argument(
+        "--target-ocp-version",
+        default="OCP Version this image was built to",
```
querti, line 206: That's the YAML description, not the default. Catalog default is `""`.

src/tasks/internal/publish_index_image/publish_index_image.py
querti, line 191: Catalog default is `"0"`.

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+    """
+    logger.info(f"Inspecting image: {index}")
+    kwargs: dict[str, Any] = {"config": config}
+    if credential:
+        kwargs["creds"] = credential
+    if retry_times is not None:
+        kwargs["retry_times"] = retry_times
+
+    ret = inspect(index, **kwargs)
+    if ret.returncode != 0:
```
querti, line 109: Auth/network failures are logged as “does not exist” and we proceed to copy. Please log `ret.stderr`.

src/helpers/skopeo/skopeo.py
```diff
@@ -13,6 +13,7 @@ def inspect(
     no_tags: bool = False,
     override_os: str | None = None,
     override_arch: str | None = None,
+    creds: str | None = None,
```
querti, line 19: This rewrite drops `is_repo_not_found`, which `apply_mapping` uses on main.

querti (commented):
(some of the comments were generated by an LLM ,so please verify)

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
seanconroy2021, line 190: Every IR created in this loop shares the same labels PR UID Pipeline name. internal_request.create() deletes any existing IR matching that same pipeline, PR UID pair before creating a new one it meant to cleanup previous attempts leftover. Wont this fire within the same run when we call create new IR each time which will delete the existing IRs. This is same issue we had before right ?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+        except InternalRequestWaitError:
+            results = _fetch_results(ir_name)
+        except Exception as e:
+            log.error(f"ERROR: Waiting for {pi} failed: {e}")
+            failed = True
+            continue
+
+        request_message = results.get("requestMessage", "")
+
+        if "error" in request_message.lower():
```
seanconroy2021, line 228: Wont this only catch failures if the message has error in it. The internal task can also return something like "The source index does not match its targetOcpVersion" which dosen't have "error" in it so those get logged as succesfull right ? Other script that use fetch_results and check the result https://github.com/konflux-ci/release-service-utils/blob/76de9191eb979ee7a1a6731b1ce1e23d2579f23d/src/tasks/managed/push_disk_images/push_disk_images.py#L163

src/tasks/internal/publish_index_image/publish_index_image.py
seanconroy2021, line 206: This looks like help= text that ended up in default= by mistake

src/tasks/internal/publish_index_image/publish_index_image.py
```diff
@@ -0,0 +1,318 @@
...
+        **kwargs,
+    )
+    if ret.returncode == 0:
+        return True, "Index Image Published successfully"
+    else:
+        logger.error(f"Error details: {ret.stderr}")
+        return False, "Error: Failed publishing Index Image"
+
+
+def write_result(message: str) -> None:
```
seanconroy2021, line 163: This just prints to stdout every other internal task in this repo writes its result tekton.result_paths_from_env should we not do the same ?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
```diff
@@ -0,0 +1,246 @@
...
+
+    # Process wait results and check outcomes
+    for ft in as_completed(wait_futures):
+        ir_name, pi = wait_futures[ft]
+        try:
+            ft.result()
+            results = _fetch_results(ir_name)
+        except InternalRequestWaitError:
+            results = _fetch_results(ir_name)
+        except Exception as e:
```
seanconroy2021, line 221: Is this intentional? This swallows the exception and just continues so main() returns 1 instead of raising but I thought we agreed that managed tasks shouldn't catch exceptions in main() at all. Same at line 239. 

Something like this instead: https://github.com/konflux-ci/release-service-utils/blob/76de9191eb979ee7a1a6731b1ce1e23d2579f23d/src/tasks/managed/close_advisory_issues/close_advisory_issues.py#L215-L230

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: fixed

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: added error logger

src/helpers/skopeo/skopeo.py
midnightercz: not sure what you mean, is_repo_not_found is defined at line 51

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: fixed

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: improved error catching

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: Now the code catches all exceptions, logs tracebacks

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: Now the code catches all exceptions, logs tracebacks

src/tasks/internal/publish_index_image/publish_index_image.py
seanconroy2021: This is not fixed. It's still the same?

src/tasks/internal/publish_index_image/publish_index_image.py
seanconroy2021: This is not fixed. It's still the same?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
seanconroy2021: This is not fixed?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
seanconroy2021: This is not fixed. It's still the same?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
seanconroy2021: This seems exactly the same as before ?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: It's fixed now, sorry I thought I pushed from the right source, but the changes were made in another

seanconroy2021:
Need to fix the conflicts

seanconroy2021:
I still don't see any changes when I check  https://github.com/konflux-ci/release-service-utils/pull/784#discussion_r4024605498 ?

src/tasks/managed/managed_publish_index_image/managed_publish_index_image.py
midnightercz: I changed internal part to always starts with error:, but now I changed to fail with it's not success

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

src/tasks/internal/publish_index_image/publish_index_image.py
midnightercz: fixed

PR #795, opened 2026-05-29
feat(RELEASE-1986): filter-already-released-advisory-images script

elenagerman:
## Describe your changes
Adds a standalone Python script for the `filter-already-released-advisory-images` internal task so its logic runs from the utils image instead of inline bash in the catalog.

## Relevant Jira
[RELEASE-1986](https://redhat.atlassian.net/browse/RELEASE-1986)

[RELEASE-1986]: https://redhat.atlassian.net/browse/RELEASE-1986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/python/helpers/test_subprocess_cmd.py
+++ b/scripts/python/helpers/test_subprocess_cmd.py
@@ -2,23 +2,30 @@
 
 from __future__ import annotations
 
+import json
+import shutil
 import subprocess
+from pathlib import Path
+from unittest import mock
 
 import pytest
 import subprocess_cmd
 
 
 def test_run_cmd_captures_stdout() -> None:
+    """``run_cmd`` returns captured stdout from a successful subprocess."""
     out = subprocess_cmd.run_cmd(["echo", "hi"], check=True).stdout.strip()
     assert out == "hi"
 
 
 def test_run_cmd_check_false(tmp_path) -> None:
+    """``run_cmd`` with ``check=False`` returns a non-zero exit code without raising."""
     r = subprocess_cmd.run_cmd(["false"], cwd=tmp_path, check=False)
     assert r.returncode != 0
 
 
 def test_run_cmd_stderr_path_logs_failed_command(tmp_path) -> None:
+    """Failed commands append a failure line to *stderr_path*."""
     log = tmp_path / "log.txt"
     with pytest.raises(subprocess.CalledProcessError):
         subprocess_cmd.run_cmd(["false"], stderr_path=log, check=True)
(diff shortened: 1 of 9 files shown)


scripts/python/tasks/internal/filter_already_released_advisory_images.py
```diff
@@ -0,0 +1,393 @@
...
+        f"--internal-request-pipeline-run-name NAME "
+        f"--internal-request-task-run-name NAME\n"
+        f"  --transformed-snapshot  Gzip+base64 JSON array of arch-specific images\n"
+        f"  --origin                Release origin workspace (advisories subdir)\n"
+    )
+
+
+def parse_args(argv: list[str] | None) -> argparse.Namespace:
+    """Parse required CLI flags for Tekton parameters."""
+    p = argparse.ArgumentParser(prog=PROG, add_help=False, usage=argparse.SUPPRESS)
```
mmalina: Why disable the normal help function here and instead add a _usage_text function that does the same? Why not just rely on what argparse offers?

mmalina: It seems you may be copying the approach that Johnny used here: https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/tasks/internal/check_embargoed_cves.py

I will raise this in a team chat.

mmalina: Raised here: https://redhat-internal.slack.com/archives/C04J47GL936/p1780558311783609

scripts/python/tasks/internal/filter_already_released_advisory_images.py
```diff
@@ -0,0 +1,393 @@
...
+) -> Any:
+    """Evaluate a ``yq`` expression against ``path`` and parse JSON output."""
+    runner = run_cmd or _default_run_cmd
+    out = runner(["yq", "-o=json", expression, str(path)])
+    if not str(out).strip():
+        return []
+    return json.loads(out)
+
+
+def _default_run_cmd(args: Sequence[str], *, cwd: Path | None = None) -> str:
```
mmalina: This is the kind of thing that should probably be in a shared library, right?

scripts/python/tasks/internal/filter_already_released_advisory_images.py
```diff
@@ -0,0 +1,393 @@
...
+import sys
+from collections.abc import Callable, Sequence
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import file
+import internal_request
+import tekton
+from git.exc import GitCommandError
```
mmalina: This is a thirdparty lib, so it shouldn't be grouped with the local imports. It seems it's the only thirdparty here, so should have its own section in the middle.

scripts/python/tasks/internal/filter_already_released_advisory_images.py
elenagerman: We use a short custom usage on stderr and exit 1 on --help or missing flags (same as the original bash/Tekton behavior). That pattern started in check_embargoed_cves.py; this task copied it.
Anyway, the shared bits (tekton_argument_parser, exit_with_usage, etc.) now live in scripts/python/helpers/subprocess_cmd.py (refactored).

scripts/python/tasks/internal/filter_already_released_advisory_images.py
elenagerman: Fair point. We already have subprocess_cmd.run_cmd, but it’s not the same shape as this stdout-only helper. I kept it local since only this script uses yq

elenagerman: but I moved it to the helpers anyway: in subprocess_cmd.py now we have two separate methods: run_cmd_text and run_yq_json

scripts/python/tasks/internal/filter_already_released_advisory_images.py
elenagerman: sure, fixed

scripts/python/tasks/internal/filter_already_released_advisory_images.py
mmalina: I still think it's unnecessarily complicated. but we'll need to agree on this as a team - I created a jira for it: https://redhat.atlassian.net/browse/RELEASE-2651

so in the meantime, I will resolve this comment.

scripts/python/tasks/internal/filter_already_released_advisory_images.py
mmalina: ok, thanks

mmalina (dismissed):

davidmogar (dismissed):

ach912 (commented):
lgtm

ach912 (dismissed):

scripts/python/helpers/test_subprocess_cmd.py
johnbieren, line 1: Shouldn't there be a test showing yaml converted to json? Isn't that the whole point of run_yq_json?

elenagerman: sure! good catch! thank you @johnbieren !
test was added

johnbieren (dismissed):

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-10T09:30:56.644Z |

Approved

PR #800, opened 2026-06-02
feat(RELEASE-1993): convert update-fbc-catalog internal task

FilipNikolovski:
Convert the inline bash scripts from the internal update-fbc-catalog-task Tekton task into a standalone Python script.

New shared helpers were added for IIB REST API interaction, container image inspection via skopeo, and a kerberos_login context manager that extracts the common Kerberos temp-file setup pattern.

Assisted-by: Claude Code

### DIFF
--- a/scripts/python/helpers/iib.py
+++ b/scripts/python/helpers/iib.py
@@ -1,0 +1,158 @@
+"""IIB (Index Image Build) REST API client helpers.
+
+Provide functions for querying, submitting, and monitoring IIB builds,
+plus gzip/base64 compression for Tekton result files (4 KB limit).
+"""
+
+from __future__ import annotations
+
+import base64
+import gzip
+import json
+from datetime import datetime
+from typing import TypedDict
+from urllib.parse import urlencode
+
+import requests
+from requests.auth import AuthBase
+
+import http_client
+from logger import logger
+
+
+class IIBBuildLogs(TypedDict, total=False):
+    """Log URLs attached to an IIB build."""
+
+    url: str
+
+
+class IIBBuild(TypedDict, total=False):
+    """Subset of fields returned by the IIB builds endpoint."""
+
+    id: int
+    state: str
+    state_reason: str | None
+    from_index: str | None
+    from_index_resolved: str | None
+    index_image: str | None
+    index_image_resolved: str | None
+    internal_index_image_copy: str | None
+    fbc_fragments: list[str] | None
+    build_tags: list[str] | None
+    distribution_scope: str | None
+    updated: str | None
+    created: str | None
+    logs: IIBBuildLogs
+    state_history: list[dict[str, str]]
+
+
+class IIBQueryResponse(TypedDict):
+    """Paginated response from ``GET /builds``."""
+
+    items: list[IIBBuild]
+
+
+class FBCOperationPayload(TypedDict, total=False):
+    """JSON body for ``POST /builds/fbc-operations``."""
+
+    fbc_fragments: list[str]
+    from_index: str
+    build_tags: list[str]
+    add_arches: list[str]
+    overwrite_from_index: bool
+    overwrite_from_index_token: str
+
+
+def query_builds(
+    iib_url: str,
+    *,
+    user: str,
+    from_index: str,
+    state: str,
+) -> IIBQueryResponse:
+    """Query IIB for builds matching *user*, *from_index*, and *state*.
+
+    Return the parsed JSON response body (contains an ``items`` list).
+    The IIB read endpoint is unauthenticated.
+    """
+    params = urlencode({"user": user, "from_index": from_index, "state": state})
+    url = f"{iib_url}/builds?{params}"
+    body = http_client.get_text(url)
+    return json.loads(body)
+
+
+def get_build(
+    iib_url: str,
+    build_id: int,
+) -> IIBBuild:
+    """Fetch a single IIB build by *build_id*.
+
+    Return the parsed JSON response body.  The endpoint is unauthenticated.
+    """
+    url = f"{iib_url}/builds/{build_id}"
+    body = http_client.get_text(url)
+    return json.loads(body)
+
+
+def submit_fbc_operation(
+    iib_url: str,
+    payload: FBCOperationPayload,
+    *,
+    auth: AuthBase,
+    verify_ssl: bool = False,
+) -> IIBBuild:
+    """POST an FBC operation to IIB with Kerberos negotiate auth.
+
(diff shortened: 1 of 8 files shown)


scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -0,0 +1,857 @@
...
+            state="complete",
+        )
+    except (
+        requests.RequestException,
+        OSError,
+        json.JSONDecodeError,
+    ) as e:
+        logger.warning("Failed to query IIB for completed builds: %s", e)
+        return None
+
```
theflockers, line 301: I am not smart enough to understand what it is happening here. Maybe some comment block explaining the logic (the part inside `matching` plus the lambda function to know when the status is updated ?

theflockers: funny that it highlighted the part before what I am talking about :D

FilipNikolovski: The bash equivalent is these lines: https://github.com/konflux-ci/release-service-catalog/blob/6929647f8c9e52687441a7bcd9a4665c101ec960/tasks/internal/update-fbc-catalog-task/update-fbc-catalog-task.yaml#L148-L153

This filters the IIB query response to find completed builds that have:
1. `fbc_fragment` set
2. same fragments as the `sorted_frags`
3. has a `distribution_scope` of prod, stage or null

FilipNikolovski: I'll add a comment explaining this.

FilipNikolovski: done.

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -0,0 +1,861 @@
...
+        raise ValueError("fbcFragments must be a JSON array")
+    if len(data) == 0:
+        raise ValueError("fbcFragments array is empty")
+    for item in data:
+        if not isinstance(item, str) or not item.strip():
+            raise ValueError("fbcFragments items must be non-empty strings")
+    return sorted(data)
+
+
+def create_container_auth_config(
```
swickersh: Maybe this could be in authentication.py

Also, just an FYI I have a couple functions in authentication.py on PR 769 that write container registry credentials to a config file before running skopeo/container operations.
Looks like the main difference is 
mine:  ~/.docker/config.json, reads pre-formed JSON from a mounted secret file
yours: ~/.config/containers/auth.json, constructs the JSON itself from a username:password credential string

So they are different enough that they could still be separate functions in that file.

scripts/python/helpers/test_authentication.py
```diff
@@ -132,3 +133,113 @@ def _fail(*args, **kwargs) -> object:  # type: ignore[no-untyped-def]
...
+        assert not f.exists()
+
+
+def test_kerberos_login_updates_environ(
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """``os.environ`` is updated with KRB5 variables inside the context."""
+    monkeypatch.delenv("KRB5_CONFIG", raising=False)
+    monkeypatch.delenv("KRB5CCNAME", raising=False)
+    monkeypatch.delenv("KRB5_TRACE", raising=False)
```
swickersh: Any reason to not unset these 3 env vars in the finally block of the kerberos_login function?

FilipNikolovski: Good catch, I've made the update.

scripts/python/tasks/internal/update_fbc_catalog.py
FilipNikolovski: Sure I think it's a good suggestion, I'll move it there. I think we should allow for some duplicate functions for now, and as we see some patterns emerge we can consolidate them, we don't want to create wrong abstractions in the beginning.

scripts/python/helpers/iib.py
```diff
@@ -0,0 +1,158 @@
...
+    error = data.get("error")  # type: ignore[call-overload]
+    if error is not None:
+        raise ValueError(f"IIB service error: {error}")
+    return data
+
+
+def parse_date_to_epoch(date_str: str) -> int:
+    """Parse an ISO 8601 date string to a Unix epoch integer."""
+    return int(datetime.fromisoformat(date_str).timestamp())
+
```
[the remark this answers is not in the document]
FilipNikolovski: This is a false positive, Python 3.12+ handles the Z suffix and nano-second precision dates just fine, have also added tests for these scenarios.

scripts/python/tasks/internal/update_fbc_catalog.py
swickersh: I agree. I think there will be refactoring regardless because we don't know who else has similar/duplicate functions while we convert these tasks simultaneously.

swickersh (dismissed):

scripts/python/tasks/internal/update_fbc_catalog.py
johnbieren, line 1: Have you tested this with a draft catalog PR? This file isn't executable so I am not sure how it would work

FilipNikolovski: Yes here's the catalog PR https://github.com/konflux-ci/release-service-catalog/pull/2268

johnbieren: oh okay you're using `python3` to call it

johnbieren (dismissed):

ach912 (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-08T10:13:39.559Z |

Approved

scripts/python/tasks/internal/update_fbc_catalog.py
FilipNikolovski: @theflockers can we resolve this?

theflockers (approved):

davidmogar (approved):

ach912 (approved):

twaugh (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-11T10:04:18.464Z |

Approved

PR #802, opened 2026-06-03
feat(RELEASE-2499): rewrite make-repo-public logic in Python

querti:
Replace the inline bash script in the make-repo-public Tekton task with a standalone Python module. The new implementation uses the requests library and shared helpers (http_client, file, logger) for cleaner error handling, CA bundle setup, and Quay registry detection. Unit tests cover all code paths including caching, failure modes, and multi-registry validation.

The Dockerfile PYTHONPATH and pyproject.toml pythonpath are extended to include scripts/python/tasks/managed/ so the new module is importable both at runtime and in pytest.

Assisted-by: Cursor

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,6 +35,7 @@ pythonpath = [
   "integration-tests/lib",
   "scripts/python/helpers",
   "scripts/python/tasks/internal",
+  "scripts/python/tasks/managed",
   "utils",
   "pubtools-pulp-wrapper",
   "publish-to-cgw-wrapper",
--- a/scripts/python/tasks/managed/make_repo_public.py
+++ b/scripts/python/tasks/managed/make_repo_public.py
@@ -1,0 +1,228 @@
+#!/usr/bin/env python3
+"""Make Quay repositories public using the Quay API."""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+from pathlib import Path
+from typing import Any
+
+import requests
+
+import file
+import http_client
+from logger import logger
+
+PROG = "make_repo_public.py"
+
+SYSTEM_CA_BUNDLE = "/etc/pki/tls/certs/ca-bundle.crt"
+
+
+def setup_ca_bundle(ca_cert_path: Path) -> None:
+    """Combine system and custom CA bundles if the custom cert exists."""
+    if not ca_cert_path.is_file():
+        return
+
+    system_bundle = Path(SYSTEM_CA_BUNDLE)
+    parts: list[bytes] = []
+    if system_bundle.is_file():
+        parts.append(system_bundle.read_bytes())
+    parts.append(ca_cert_path.read_bytes())
+    combined = file.make_tempfile_path(prefix="combined-ca-bundle-", data=b"\n".join(parts))
+
+    os.environ["SSL_CERT_FILE"] = str(combined)
+    os.environ["CURL_CA_BUNDLE"] = str(combined)
+    os.environ["REQUESTS_CA_BUNDLE"] = str(combined)
+
+
+def is_quay_registry(
+    registry: str,
+    session: requests.Session,
+    cache: dict[str, bool],
+) -> bool:
+    """Return True if ``registry`` exposes a Quay-compatible discovery endpoint.
+
+    Issue a GET to ``https://{registry}/api/v1/discovery``; HTTP 200 means
+    Quay, any other status means not Quay. Results are cached in ``cache``
+    so each registry is probed only once.
+    """
+    if registry in cache:
+        return cache[registry]
+
+    url = f"https://{registry}/api/v1/discovery"
+    try:
+        resp = session.get(url, timeout=30)
+        result = resp.status_code == 200
+    except requests.RequestException as exc:
+        logger.warning(
+            "Failed to probe discovery endpoint for %s: %s",
+            registry,
+            exc,
+        )
+        result = False
+
+    cache[registry] = result
+    return result
+
+
+def make_repo_public(
+    registry: str,
+    repo_path: str,
+    token: str,
+    session: requests.Session,
+) -> None:
+    """POST to the Quay API to change repository visibility to public.
+
+    Raises ``RuntimeError`` on non-2xx responses. When ``REGISTRY_SECRET_NAME``
+    is set in the environment, the error message includes a hint about the
+    expected secret key and permissions.
+    """
+    url = f"https://{registry}/api/v1/repository/{repo_path}/changevisibility"
+    try:
+        resp = session.post(
+            url,
+            headers={
+                "Authorization": f"Bearer {token}",
+                "Content-Type": "application/json",
+            },
+            json={"visibility": "public"},
+            timeout=30,
+        )
+    except requests.RequestException as exc:
+        raise RuntimeError(
+            f"Failed to connect to {registry} to make {repo_path} public: {exc}"
+        ) from exc
+    if resp.ok:
+        logger.info("Repository %s/%s is now public", registry, repo_path)
+        return
+
+    msg = (
+        f"Failed to make repo {registry}/{repo_path} public"
+        f" (HTTP {resp.status_code}: {resp.text})."
+    )
+    secret_name = os.environ.get("REGISTRY_SECRET_NAME", "").strip()
+    if secret_name:
+        msg += (
+            f" Make sure the secret {secret_name} contains"
+            ' the "token" key with token that has permission to'
+            " Administer Repositories."
+        )
+    raise RuntimeError(msg)
+
+
+def run(
+    data_file: Path,
+    snapshot_file: Path,
+    secret_path: Path,
+    ca_cert_path: Path,
+) -> None:
+    """Orchestrate making repositories public based on data and snapshot files.
+
+    Reads the merged data JSON and snapshot JSON, then for each component with
+    ``public: true`` calls the Quay API to change visibility. Only a single
+    Quay registry is supported per invocation; encountering repos on two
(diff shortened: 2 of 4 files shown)


pyproject.toml
```diff
@@ -36,6 +36,7 @@ pythonpath = [
   "scripts/python/helpers",
   "scripts/python/tasks/internal",
   "utils",
+  "scripts/python/tasks/managed",
```
mmalina, line 38: Nit: Maybe move it one line higher to group it with the other scripts/python lines? (And it looks it's actually in alphabetical order as well.)

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+    cache: dict[str, bool],
+) -> bool:
+    """
+    Return True if ``registry`` exposes a Quay-compatible discovery endpoint.
+
+    Issue a GET to ``https://{registry}/api/v1/discovery``; HTTP 200 means
+    Quay, any other status means not Quay. Successful probe results are
+    cached in ``cache`` so each registry is probed only once.  Connection
+    failures are **not** cached so that a later repo on the same registry
+    can retry the probe.
```
mmalina: Can you explain this? Why would we want this? If it's the same registry, we should just use the same result as we got for the previous one, no? Why check again at all?

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+
+            if not is_quay_registry(registry, session, quay_cache):
+                logger.warning("Registry %s is not a Quay instance. Skipping.", registry)
+                continue
+
+            if target_registry is None:
+                target_registry = registry
+            elif target_registry != registry:
+                raise RuntimeError(
+                    f"Multiple Quay registries found ({target_registry} and"
```
mmalina, line 184: The error would only trigger for the second registry - after the first one was already made public. Would it make sense to check this before starting? Like:

```
  # Before the main loop
  quay_registries = set()
  for component in components:
      if component.get("public", default_public):
          for repo_entry in component.get("repositories", []):
              registry = repo_entry.get("url", "").split("/")[0]
              if is_quay_registry(registry, session, quay_cache):
                  quay_registries.add(registry)

  if len(quay_registries) > 1:
      raise RuntimeError(f"Multiple Quay registries found:
  {quay_registries}")

  # Then proceed with making repos public...
```

Feel free to dismiss this - I guess this is likely how the original task works, so improving logic is probably out of scope anyway...

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+    components = snapshot.get("components", [])
+    for component in components:
+        public = component.get("public", default_public)
+        if str(public).lower() != "true":
+            continue
+
+        repositories = component.get("repositories", [])
+        for repo_entry in repositories:
+            repo_url = repo_entry.get("url", "")
+            if not repo_url:
```
mmalina, line 165: What is the use case here? Shouldn't this be an error instead?

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+
+SYSTEM_CA_BUNDLE = "/etc/pki/tls/certs/ca-bundle.crt"
+
+
+def setup_ca_bundle(ca_cert_path: Path) -> None:
+    """Combine system and custom CA bundles if the custom cert exists."""
+    if not ca_cert_path.is_file():
+        return
+
+    combined = Path("/tmp/combined-ca-bundle.crt")
```
mmalina: Would it make sense to use an equivalent of `mktemp` rather than hardcoding a /tmp path? Like `tempfile.NamedTemporaryFile()`?

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+        )
+    except requests.RequestException as exc:
+        raise RuntimeError(
+            f"Failed to connect to {registry} to make {repo_path} public: {exc}"
+        ) from exc
+    if resp.ok:
+        logger.info("Repository %s/%s is now public", registry, repo_path)
+        return
+
+    msg = (
```
FilipNikolovski, line 101: I dont think the logic from this line would run at all, since all exceptions raised by `session.post` inherit from `RequestException` : https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

To catch a specific non 2xx error you could use the HTTPError exception.

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+
+    setup_ca_bundle(ca_cert_path)
+
+    token_path = secret_path / "token"
+    if not token_path.is_file():
+        raise RuntimeError(f"Registry secret token file not found at {token_path}")
+    token = token_path.read_text(encoding="utf-8").strip()
+
+    try:
+        data: dict[str, Any] = json.loads(data_file.read_text(encoding="utf-8"))
```
FilipNikolovski, line 141: Could we make this more strict instead of using `Any`?

scripts/python/tasks/managed/make_repo_public.py
```diff
@@ -0,0 +1,232 @@
...
+    except json.JSONDecodeError as exc:
+        raise RuntimeError(f"Invalid JSON in data file {data_file}: {exc}") from exc
+    try:
+        snapshot: dict[str, Any] = json.loads(snapshot_file.read_text(encoding="utf-8"))
```
FilipNikolovski, line 145: Could we make this more strict instead of using Any?

pyproject.toml
querti: Changed the ordering.

scripts/python/tasks/managed/make_repo_public.py
querti: The idea was that a network failure shouldn't be cached, since it may be transient. But since we have a retry mechanism by default, I think there's not much point to retrying with a different entry in this case. So I have removed the exception to not cache in case of a network failure.

scripts/python/tasks/managed/make_repo_public.py
querti: It was in the original implementation. Your suggestion is an improvement, but I think it's a very niche use-case and it's very unlikely that multiple quay registries will ever be used like this. So I'll keep it as-is if that's OK with you.

scripts/python/tasks/managed/make_repo_public.py
querti: Good point, added an error

scripts/python/tasks/managed/make_repo_public.py
querti: Yeah, it's better from a security standpoint and more consistent across the repo. I changed the implementation.

scripts/python/tasks/managed/make_repo_public.py
querti: Hmm I don't think this is true - I don't think requests raises non-200 exceptions by default. From your link:
`[Response.raise_for_status()](https://requests.readthedocs.io/en/latest/api/#requests.Response.raise_for_status) will raise an [HTTPError](https://requests.readthedocs.io/en/latest/api/#requests.HTTPError) if the HTTP request returned an unsuccessful status code.`  which implies that they have to be raised manually.

scripts/python/tasks/managed/make_repo_public.py
querti: I'm not sure what could be here to be honest. Are there any restrictions? I can imagine a scenario where a different type will be added in the future that would break this task. 

Would it be OK with you to keep it as-is? I don't think this task should care too much about data validation.

scripts/python/tasks/managed/make_repo_public.py
querti: Same comment as above

scripts/python/tasks/managed/make_repo_public.py
FilipNikolovski: You're right we actually need to call the `raise_for_status()` method to raise the HTTPError exception, so this is fine. I'll resolve this.

scripts/python/tasks/managed/make_repo_public.py
FilipNikolovski: Hm yeah, mapping can contain anything I suppose, we could use something like a TypedDict, but I don't think there's any benefit to it here. Ok I think this is fine as-is, I'll resolve this.

FilipNikolovski (commented):
LGTM once @mmalina comments are resolved.

scripts/python/tasks/managed/make_repo_public.py
mmalina: ok, thanks

scripts/python/tasks/managed/make_repo_public.py
mmalina: yeah, that's fine

mmalina (dismissed):

ach912 (dismissed):
lgtm

johnbieren (dismissed):
is_quay_registry will probably end up being moved to a helper, but fine where it is for now IMO

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-10T08:53:17.270Z |

Approved

twaugh (approved):

mmalina (approved):

PR #804, opened 2026-06-03
feat(RELEASE-2509): add python scripts for publish-to-nrrc task

elenagerman:
## Describe your changes
Moves the prepare-repo and upload-npm-archive logic from inline Tekton bash into standalone Python scripts under scripts/python/tasks/managed/, with Tekton parameters passed as CLI flags. 
It adds a small charon_env helper for parsing charon parameter files, reuses the existing subprocess_cmd helper from create_advisory, and bundles the charon CLI into the utils image so both steps can run from release-service-utils. 

## Relevant Jira
[RELEASE-2509](https://redhat.atlassian.net/browse/RELEASE-2509)

Assisted-by: Claude

[RELEASE-2509]: https://redhat.atlassian.net/browse/RELEASE-2509?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/python/helpers/charon_env.py
+++ b/scripts/python/helpers/charon_env.py
@@ -1,0 +1,82 @@
+"""Parse charon parameter files used by NRRC/MRRC publish tasks."""
+
+from __future__ import annotations
+
+import re
+import shutil
+from pathlib import Path
+
+from dotenv import dotenv_values
+
+import file
+
+_REGISTRY_SPLIT = re.compile(r"%")
+
+
+def load_charon_env(path: Path) -> dict[str, str]:
+    """Load charon parameters from a dotenv file (``KEY=value`` lines)."""
+    if not path.is_file():
+        raise FileNotFoundError(f"charon env file not found: {path}")
+    values = dotenv_values(path, encoding="utf-8")
+    return {key: value for key, value in values.items() if value is not None}
+
+
+def split_oci_registries(value: str) -> list[str]:
+    """Split ``CHARON_OCI_REGISTRY`` on ``%`` into non-empty registry references."""
+    return [part.strip() for part in _REGISTRY_SPLIT.split(value) if part.strip()]
+
+
+def short_sha256_prefix(registry: str) -> str:
+    """Return the first six characters of the digest in *registry*."""
+    marker = "@sha256:"
+    if marker not in registry:
+        raise ValueError(f"registry reference missing @sha256: digest: {registry!r}")
+    return registry.split(marker, 1)[1][:6]
+
+
+def source_repo(registry: str) -> str:
+    """Return the repository part of an OCI reference (before ``@sha256:``)."""
+    return registry.split("@sha256:", 1)[0]
+
+
+def require_env_keys(env: dict[str, str], *keys: str) -> None:
+    """Raise ValueError when any *keys* are missing from *env*."""
+    for key in keys:
+        if key not in env:
+            raise ValueError(f"missing required charon env variable: {key}")
+
+
+def require_oci_registries(env: dict[str, str]) -> list[str]:
+    """Return non-empty ``CHARON_OCI_REGISTRY`` entries from *env*."""
+    try:
+        value = env["CHARON_OCI_REGISTRY"]
+    except KeyError as e:
+        raise ValueError("CHARON_OCI_REGISTRY is required in charon env file") from e
+    registries = split_oci_registries(value)
+    if not registries:
+        raise ValueError("CHARON_OCI_REGISTRY must list at least one registry reference")
+    return registries
+
+
+def charon_config_path(*, home: Path | None = None) -> Path:
+    """Return the default charon configuration file path under *home* or ``Path.home()``."""
+    root = home if home is not None else Path.home()
+    return root / ".charon" / "charon.yaml"
+
+
+def install_charon_config(config_source: Path, *, home: Path | None = None) -> Path:
+    """Copy the charon config into ``$HOME/.charon/charon.yaml``."""
+    if not config_source.is_file():
+        raise FileNotFoundError(f"charon config file not found: {config_source}")
+    dest = charon_config_path(home=home)
+    dest.parent.mkdir(parents=True, exist_ok=True)
+    shutil.copy2(config_source, dest)
+    return dest
+
+
+NRRC_WORK_DIR_DEFAULT = Path("/var/workdir/nrrc")
+
+
+def nrrc_work_dir() -> Path:
+    """Return the NRRC staging directory from ``WORK_DIR`` or ``/var/workdir/nrrc``."""
+    return file.path_from_env_variable("WORK_DIR", NRRC_WORK_DIR_DEFAULT)
--- a/scripts/python/helpers/file.py
+++ b/scripts/python/helpers/file.py
@@ -7,7 +7,10 @@ import json
 import gzip
 import io
 import os
+import re
+import subprocess
 import tempfile
+from collections.abc import Callable, Sequence
 from pathlib import Path
 from typing import Any
 
@@ -23,6 +26,7 @@ def load_json_dict(path: Path) -> dict[str, Any]:
 
 
 _GZIP_READ_CHUNK_SIZE = 64 * 1024
+_ARCHIVE_TYPE = re.compile(r"(gzip compressed data|POSIX tar archive)")
 
 
 def sha256(path: Path) -> str:
@@ -55,6 +59,25 @@ def path_from_env_variable(
     if raw is not None and str(raw).strip() != "":
         return Path(str(raw).strip())
     return default if isinstance(default, Path) else Path(default)
+
+
+def resolve_path_under_base(base: Path, relative: str | Path) -> Path:
+    """Resolve *relative* under *base* and ensure the result stays inside *base*.
+
+    Rejects absolute paths and ``..`` traversal after resolution. Typical use:
+    Tekton passes a path relative to a data directory (e.g. charon env/config files).
+    """
+    text = str(relative).strip()
+    if not text:
+        raise ValueError(f"path must be relative to {base}: {relative!r}")
+    rel = Path(text)
+    if rel.is_absolute():
+        raise ValueError(f"path must be relative to {base}: {relative!r}")
+    root = base.resolve()
+    candidate = (root / rel).resolve()
+    if not candidate.is_relative_to(root):
+        raise ValueError(f"path must stay under {base}: {relative!r}")
+    return candidate
 
 
 def make_tempfile_path(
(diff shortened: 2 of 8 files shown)


Dockerfile
```diff
@@ -36,6 +38,10 @@ RUN dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.
 COPY --from=oras /usr/bin/oras /usr/bin/oras
 COPY --from=oras /usr/local/bin/select-oci-auth /usr/local/bin/select-oci-auth
 COPY --from=oras /usr/local/bin/get-reference-base /usr/local/bin/get-reference-base
+COPY --from=charon /usr/local/bin/charon /usr/local/bin/charon
+COPY --from=charon /usr/local/lib/python3.12/site-packages/charon /usr/local/lib/python3.12/site-packages/charon
```
mmalina: Do we really want to copy python libs from another image? If it's a python package, then hopefully we can install it via pypy? Also, this is kind of a special case that maybe we haven't hit before - maybe it's worth bringing up with the team what we want to do here? If we strictly require one step and one script per task, then I guess we'll need charon in the utils image. But maybe it's acceptable to keep that step separate?

elenagerman: Probably the answer is YES; unfortunately
We're moving both publish-to-nrrc bash blocks from the catalogue into utils Python scripts (prepare + upload). The upload step still calls the Konflux charon CLI, and, unfortunatelly, that's not the public pypI package, and it isn't published to an index we can pip install from today. 
Copying the charon binary and its Python dependencies from the charon image is how we run that upload logic from utils without leaving the script in the catalogue.
The alternative is a partial migration: keep upload as inline bash on the charon image and only move prepare-repo to utils.

elenagerman: and utils already does this for other tools: oras, cosign, roxctl. 
and, since charon isn’t on PyPI, so copying from the pinned charon image is the same “borrow from a known-good image” approach (IMO)

mmalina: > and utils already does this for other tools: oras, cosign, roxctl.

yes, but until now we never copied python libs like this. that seems really fragile - notice the python version in the path. we happen to be on python3.12 as well. what if that image moves to another version?

> The alternative is a partial migration: keep upload as inline bash on the charon image and only move prepare-repo to utils.

that's why I suggested opening this topic in slack, because I think this is a special case that we haven't really encountered before. I still think copying the python libraries like this is not a good idea.

mmalina: How about this? https://github.com/Commonjava/charon#from-git

elenagerman: Good call on the fragile site-packages copy! Agreed we shouldn’t tie utils to the charon image’s Python layout.
I dropped the COPY --from=charon lines and install charon via pyproject.toml from [Commonjava/charon](https://github.com/Commonjava/charon) at tag charon-1.4.0 (thanks for the link sharing).

elenagerman: and I open a slack discussion as well

mmalina: Thanks for opening the discussion. Hopefully we can get a few more people to chime in. But right now I think it would be worth trying to reach out to the team about the internal PyPI.

scripts/python/helpers/charon_env.py
```diff
@@ -0,0 +1,84 @@
...
+
+_REGISTRY_SPLIT = re.compile(r"%")
+
+
+def load_charon_env(path: Path) -> dict[str, str]:
+    """Parse a charon ``.env`` file containing ``export KEY=value`` lines."""
+    if not path.is_file():
+        raise FileNotFoundError(f"charon env file not found: {path}")
+    env: dict[str, str] = {}
+    for raw in path.read_text(encoding="utf-8", errors="replace").splitlines():
```
theflockers: Claude signalized this as critical, so I think it makes sense to crosscheck if this should not be handled differently:

```
Silent data corruption in charon_env.py:16 - Using errors="replace" will silently corrupt 
non-UTF8 characters. Should use strict error handling instead
```

I don't have any knowledge about "charon", but I think it worth it taking a look.

mmalina: Yeah, I think `errors="replace"` should be dropped - we expect ASCII text and so if there are some non-standard characters, it's better to fail.

scripts/python/helpers/charon_env.py
```diff
@@ -0,0 +1,84 @@
...
+        line = raw.strip()
+        if not line or line.startswith("#"):
+            continue
+        if line.startswith("export "):
```
mmalina: This looked fragile - why would we want to implement a parser of an env file? So I went and looked how it's done now. The env file is generated in collect-charon-params and it will generate lines in the format `export VAR_NAME=value` with no exception (sometimes quoted, sometimes not). I think this whole thing should really be refactored - there is no sense generate it with `export ` on each line only to parse that back in this python script. I think we should probably change it to generate a normal env file (no `export ` prefixes) and then use python-dotenv to load the file.

I understand such changes are not really in scope of the original jira. But I think it makes sense to increase the scope here, rather than working around the current collect task implementation and fixing it later.

elenagerman: sure
I replaced a custom parser with the python-dotenv

scripts/python/helpers/charon_env.py
elenagerman: good catch! thank you!
modified

Dockerfile
elenagerman: I'm closing this discussion for now.
I dropped charon from the utils image. Utils only prepares the npm files; the catalog still uploads them using the existing charon container on quay.

mmalina: ok, sounds good, thanks

mmalina (dismissed):

scripts/python/helpers/file.py
```diff
@@ -57,6 +61,33 @@ def path_from_env_variable(
...
+    candidate = (root / rel).resolve()
+    if not candidate.is_relative_to(root):
+        raise ValueError(f"path must stay under {base}: {relative!r}")
+    return candidate
+
+
+NRRC_WORK_DIR_DEFAULT = Path("/var/workdir/nrrc")
+
+
+def nrrc_work_dir() -> Path:
```
johnbieren: Why is this in file.py?

scripts/python/helpers/charon_env.py
johnbieren, line 1: `charon_env` isn't a managed task. If this is a helper, can we put it in the helpers dir?

johnbieren (commented):
I'm confused by this PR. We have talked about how when we convert to python, we want to go from multi step to single step in the tekton tasks. Here, I see the `publish-to-nrrc` task split into multi python scripts. Why? Also, if it isn't a managed task name, it shouldn't really be in `tasks/managed` it should be in `helpers`

elenagerman:
> I'm confused by this PR. We have talked about how when we convert to python, we want to go from multi step to single step in the tekton tasks. Here, I see the `publish-to-nrrc` task split into multi python scripts. Why? Also, if it isn't a managed task name, it shouldn't really be in `tasks/managed` it should be in `helpers`

Good questions!
you are right, the layout is confusing.
publish_to_nrrc_upload.py shouldn’t be there. 
It was a leftover from an earlier attempt; removed for now.
+ agree: charon_env.py moved to helpers

scripts/python/helpers/charon_env.py
elenagerman: sure, moved under the helpers

scripts/python/helpers/file.py
elenagerman: Good catch! It had to be moved, and I probably forgot to do so.

Anyway, now it is in the right place; fixed
Thanks a lot for catching!!!

scripts/python/tasks/managed/publish_to_nrrc.py
```diff
@@ -0,0 +1,92 @@
...
+import subprocess_cmd
+import tekton
+from logger import logger
+
+
+def prepare_repo(
+    *,
+    charon_param_file: Path,
+    work_dir: Path,
+    run_cmd: Callable[..., subprocess.CompletedProcess[str]] = subprocess_cmd.run_cmd,
```
johnbieren: Can you hardcode this func? I don't think we need it configurable

scripts/python/tasks/managed/publish_to_nrrc.py
```diff
@@ -0,0 +1,92 @@
...
+    for registry in registries:
+        logger.info("Downloading the npm archive from %s", registry)
+        short_hash = charon_env.short_sha256_prefix(registry)
+        subdir = repo_dir / short_hash
+        subdir.mkdir(parents=True, exist_ok=True)
+
+        source_repo = charon_env.source_repo(registry)
+        auth_file = file.make_tempfile_path("oci-auth-", None)
+        try:
+            auth = run_cmd(
```
johnbieren: Can you use `oras_utils.oras_pull` here?

scripts/python/tasks/managed/publish_to_nrrc.py
johnbieren, line 1: Where is the `charon upload` call?

scripts/python/tasks/managed/publish_to_nrrc.py
elenagerman: good point! thank you @johnbieren !
done

scripts/python/tasks/managed/publish_to_nrrc.py
elenagerman: sure

scripts/python/tasks/managed/publish_to_nrrc.py
elenagerman: This part left for catalog since we cannot use quay image from utils
The task was always two steps, and I only migrated the first one to Python:
- prepare-repo (utils image): only downloads OCI artifacts with oras pull and collects gzip/tar files. No charon CLI here.
- upload-npm-archive (charon image): inline bash in tasks/managed/publish-to-nrrc/publish-to-nrrc.yaml sources charon.env, reads archives from /var/workdir/nrrc/shared, and runs charon upload 

Full migration meant charon in utils via site-packages copy (fragile) or git/GitHub (special case). Upload already belongs on the charon Quay image; I moved only prepare to utils and left upload in the catalog until charon has a proper install path (ideally internal PyPI).

johnbieren: Ah, right, my bad!

johnbieren (approved):

davidmogar (approved):

ach912 (approved):
lgtm

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-01T07:26:44.585Z |

Approved

theflockers (approved):

PR #807, opened 2026-06-03
feat(RELEASE-1990): add pulp-push-disk-images python script

happybhati:
This commit adds a python script to replicate the functionality of the inline bash script in the pulp-push-disk-images internal task in the catalog repo. The task will be updated to use this python module instead.

It also adds a subprocess_cmd helper module for running external commands with optional stderr logging.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/oras_utils.py
+++ b/scripts/python/helpers/oras_utils.py
@@ -7,6 +7,8 @@ import subprocess
 import tempfile
 from pathlib import Path
 
+import file
+import subprocess_cmd
 from subprocess_cmd import run_cmd
 
 
@@ -50,6 +52,37 @@ def oras_login(registry: str, username: str, password: str) -> None:
         stdout=subprocess.DEVNULL,
         stderr=subprocess.DEVNULL,
     )
+
+
+def oras_pull(
+    pull_spec: str,
+    download_dir: Path,
+    *,
+    stderr_path: Path | None = None,
+) -> None:
+    """Pull an OCI artifact into *download_dir* using select-oci-auth and oras."""
+    auth_file = file.make_tempfile_path("oras-auth-")
+    try:
+        auth_out = subprocess_cmd.run_cmd(
+            ["select-oci-auth", str(pull_spec)],
+            check=True,
+        ).stdout
+        auth_file.write_text(auth_out, encoding="utf-8")
+        subprocess_cmd.run_cmd(
+            [
+                "oras",
+                "pull",
+                "--registry-config",
+                str(auth_file),
+                str(pull_spec),
+            ],
+            cwd=download_dir,
+            stderr_path=stderr_path,
+            check=True,
+        )
+    finally:
+        # Always remove the auth file; subprocess failures still propagate to callers.
+        auth_file.unlink(missing_ok=True)
 
 
 def oras_push(tag: str, directory: Path, subdirectory: str, component_name: str) -> str:
--- a/scripts/python/tasks/internal/pulp_push_disk_images.py
+++ b/scripts/python/tasks/internal/pulp_push_disk_images.py
@@ -1,0 +1,423 @@
+#!/usr/bin/env python3
+"""Push disk images with Pulp and publish metadata to the Developer Portal / CGW.
+
+Reads snapshot JSON and credentials from mounted secrets, pulls OCI artifacts with
+oras, stages files, invokes pulp_push_wrapper and developer_portal_wrapper.
+
+Environment variables (set by the pulp-push-disk-images Tekton task):
+  SNAPSHOT_JSON, CERT_EXPIRATION_WARN_DAYS, CONCURRENT_LIMIT, EXODUS_GW_ENV,
+  CGW_HOSTNAME, RESULT_RESULT
+
+Mount paths default to /mnt/exodusGwSecret, /mnt/pulpSecret, /mnt/udcacheSecret,
+/mnt/redhat-workloads-token, /mnt/cgwSecret and can be overridden in tests.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from collections.abc import Callable, Mapping
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from pathlib import Path
+from typing import Any
+
+import authentication
+import file
+import oras_utils
+import push_artifacts
+import subprocess_cmd
+import tekton
+from logger import logger
+
+PROG = "pulp_push_disk_images.py"
+
+DEFAULT_EXODUS_MOUNT = Path("/mnt/exodusGwSecret")
+DEFAULT_PULP_MOUNT = Path("/mnt/pulpSecret")
+DEFAULT_UDCACHE_MOUNT = Path("/mnt/udcacheSecret")
+DEFAULT_WORKLOADS_MOUNT = Path("/mnt/redhat-workloads-token")
+DEFAULT_CGW_MOUNT = Path("/mnt/cgwSecret")
+
+RunCmd = Callable[..., subprocess.CompletedProcess[str]]
+
+
+def _validate_certificates(
+    warn_days: int,
+    *,
+    exodus_mount: Path,
+    pulp_mount: Path,
+    udcache_mount: Path,
+) -> None:
+    """Validate Exodus, Pulp, and UDCache certificate expiration."""
+    logger.info("=== Checking certificate expiration ===")
+    for label, cert_path in (
+        ("Exodus Gateway", exodus_mount / "cert"),
+        ("Pulp", pulp_mount / "konflux-release-rhsm-pulp.crt"),
+        ("UDCache", udcache_mount / "cert"),
+    ):
+        logger.info("Checking %s certificate", label)
+        push_artifacts._check_cert_expiration(str(cert_path), warn_days)
+    logger.info("=== All certificates are valid ===")
+
+
+def normalize_docker_config(raw: str) -> str:
+    """Strip extra quoted fields from a dockerconfigjson secret payload."""
+    return re.sub(r"(^|\})[^{}]+(\{|$)", r"\1\2", raw)
+
+
+def build_staged_payload(
+    disk_image_dir: Path,
+    version: str,
+) -> dict[str, Any]:
+    """Build the staged.json payload listing every file under *disk_image_dir*."""
+    payload: dict[str, Any] = {"header": {"version": "0.2"}, "payload": {"files": []}}
+    files: list[dict[str, str]] = []
+    for path in sorted(disk_image_dir.rglob("*")):
+        if not path.is_file():
+            continue
+        rel = path.relative_to(disk_image_dir).as_posix()
+        files.append(
+            {
+                "filename": path.name,
+                "relative_path": rel,
+                "version": version,
+            }
+        )
+    payload["payload"]["files"] = files
+    return payload
+
+
+def require_json_field(data: dict[str, Any], *keys: str) -> Any:
+    """Return ``data[keys...]`` or raise ``ValueError`` with a clear message."""
+    cur: Any = data
+    path = ""
+    for key in keys:
+        path = f"{path}.{key}" if path else key
+        if not isinstance(cur, dict) or key not in cur:
+            raise ValueError(f"Missing {path} value for component")
+        cur = cur[key]
+    if cur in (None, ""):
+        raise ValueError(f"Missing {path} value for component")
+    return cur
+
+
+def require_staged_files_field(entry: dict[str, Any], field: str) -> Any:
+    """Validate one ``staged.files[]`` mapping entry (bash ``jq -er`` equivalent)."""
+    label = f"staged.files[].{field}"
+    if not isinstance(entry, dict) or field not in entry or entry[field] in (None, ""):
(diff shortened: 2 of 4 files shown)


scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,485 @@
...
+) -> None:
+    """Pull one OCI artifact and stage mapped files for Pulp upload."""
+    pull_spec = require_json_field(component, "containerImage")
+    destination_name = require_json_field(component, "staged", "destination")
+    destination = disk_image_dir / str(destination_name) / "FILES"
+    destination.mkdir(parents=True, exist_ok=True)
+
+    with tempfile.TemporaryDirectory() as download_dir:
+        download = Path(download_dir)
+        auth_file = Path(tempfile.mkstemp()[1])
```
swickersh:
```suggestion
        auth_file = file.make_tempfile_path("oras-auth-")
```
Maybe use the one from file module

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,485 @@
...
+            "Needed to publish to customer portal"
+        )
+        print(msg, file=sys.stderr)
+        with stderr_path.open("a", encoding="utf-8") as errf:
+            errf.write(msg + "\n")
+        raise _StepFailure("exit 1")
+
+    with tempfile.TemporaryDirectory() as disk_dir_name:
+        disk_image_dir = Path(disk_dir_name)
+        os.environ["DISK_IMAGE_DIR"] = str(disk_image_dir)
```
swickersh: It doesn't seem like this is ever used anywhere. The disk_image_dir variable is used, but I don't see anything reading the value from the env.

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,485 @@
...
+        auth_file = Path(tempfile.mkstemp()[1])
+        try:
+            auth_out = run_cmd(
+                ["select-oci-auth", str(pull_spec)],
+                check=True,
+            ).stdout
+            auth_file.write_text(auth_out, encoding="utf-8")
+            run_cmd(
+                [
+                    "oras",
```
swickersh: There is an oras helper that this might be a nice addition to

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,485 @@
...
+        components[0].get("staged", {}).get("version")
+        if isinstance(components[0], dict)
+        else None
+    )
+    if not version:
+        msg = (
+            "Error: version not specified in .components[0].staged.version. "
+            "Needed to publish to customer portal"
+        )
+        print(msg, file=sys.stderr)
```
swickersh: This print seems redundant

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,485 @@
...
+
+    for path, content in (
+        (exodus_gw_cert, exodus_cert),
+        (exodus_gw_key, exodus_key),
+        (pulp_cert_file, pulp_cert),
+        (pulp_key_file, pulp_key),
+        (udcache_cert, udc_cert),
+        (udcache_key, udc_key),
+    ):
+        path.write_text(content, encoding="utf-8")
```
swickersh: didn't these all already get written above with the make_tempfile_path?

happybhati: Removed it. Thanks for spotting that.

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Dropped the extra print. Thanks.

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Moved the pull logic into oras_pull() in oras_utils.py so it matches the rest of the OCI helpers. Appreciate the nudge — much cleaner.

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: You're right, nothing reads it in Python — we pass disk_image_dir directly. Removed the export; leftover from the bash port. Thanks.

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Switched to file.make_tempfile_path inside oras_pull(). Thanks for pointing me at the existing helper.

scripts/python/helpers/oras_utils.py
```diff
@@ -23,6 +29,37 @@ def oras_login(registry: str, username: str, password: str) -> None:
     )
 
 
+def oras_pull(
```
swickersh, line 57: Somehow oras_utils.py landed without unit tests prior to your change. (it was me :shamehide)
Could you please add a test file for this function? I can add tests for the other functions later.

happybhati: Yes for sure, I am on it. Thanks for pointing 🙂

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,459 @@
...
+                str(pulp_key_file),
+                "--udcache-url",
+                udc_url,
+            ],
+            env=env,
+            stderr_path=stderr_path,
+            check=True,
+        )
+
+        component_destinations = [
```
swickersh: This looks pre-existing from bash, but worth fixing here while we're rewriting. `component_destinations` is a flat list of all staged files indexed by component position — if any component has more than one staged file, subsequent components will receive the wrong content directory passed to `developer_portal_wrapper`.

Since the component already carries its destination in `staged.destination`, you can derive the directory directly and drop the `component_destinations` list entirely:

```python
for component in components:
    if not isinstance(component, dict):
        continue
    destination_name = require_json_field(component, "staged", "destination")
    content_directory = disk_image_dir / str(destination_name) / "FILES"
    process_component_for_developer_portal(
        component,
        content_directory,
        cgw_hostname,
        env=env,
        stderr_path=stderr_path,
        run_cmd=run_cmd,
    )
```

Also worth adding a unit test for the multi-component edge case (two components, first one with two staged files) to make sure the directories don't get crossed.

Cursor on behalf of Scott

swickersh: Cursor highlighted this one while reviewing with me. If it's truly pre-existing we could create a Jira and tackle it later.

scripts/python/helpers/oras_utils.py
happybhati: Added `test_oras_utils.py` with tests for the auth/oras flow and temp file cleanup.

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: This was pre-existing in the bash task — the Python port carried over the same `component_destinations[$i]` indexing. Fixed here to use `staged.destination` per component, and added a unit test (two components, first with two staged files). Thanks for catching it.

swickersh (dismissed):

scripts/python/helpers/oras_utils.py
```diff
@@ -23,6 +29,37 @@ def oras_login(registry: str, username: str, password: str) -> None:
...
+    pull_spec: str,
+    download_dir: Path,
+    *,
+    run_cmd: RunCmd = subprocess_cmd.run_cmd,
```
johnbieren: Martin made a good point on one of my PRs about style like this - we never really are going to use it with a different function. If you ask cursor, it can adjust and have (in this case) `subprocess_cmd.run_cmd` be hardcoded and still have passing tests

happybhati: Thanks for the pointer.

Removed the `run_cmd` parameter from `oras_pull` and hardcoded `subprocess_cmd.run_cmd`. Updated tests to monkeypatch instead.

swickersh (dismissed):

scripts/python/helpers/oras_utils.py
```diff
@@ -23,6 +26,36 @@ def oras_login(registry: str, username: str, password: str) -> None:
...
+                "pull",
+                "--registry-config",
+                str(auth_file),
+                str(pull_spec),
+            ],
+            cwd=download_dir,
+            stderr_path=stderr_path,
+            check=True,
+        )
+    finally:
```
johnbieren, line 83: so won't this never fail? Is that intended? I would think we would want the function to call if the subprocess fails

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,455 @@
...
+DEFAULT_EXODUS_MOUNT = Path("/mnt/exodusGwSecret")
+DEFAULT_PULP_MOUNT = Path("/mnt/pulpSecret")
+DEFAULT_UDCACHE_MOUNT = Path("/mnt/udcacheSecret")
+DEFAULT_WORKLOADS_MOUNT = Path("/mnt/redhat-workloads-token")
+DEFAULT_CGW_MOUNT = Path("/mnt/cgwSecret")
+
+RunCmd = Callable[..., subprocess.CompletedProcess[str]]
+
+
+class _StepFailure(ValueError):
```
johnbieren: This sounds like it is trying to achieve the same thing that `CheckStepError` in the tekton.py helper already does

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,455 @@
...
+    """Raised when a step fails the same way as bash ``set -e`` + ``exitfunc``."""
+
+    def __init__(self, command: str, *, exit_code: int = 1, line: int = 0) -> None:
+        self.command = command
+        self.exit_code = exit_code
+        self.line = line
+        super().__init__(command)
+
+
+def read_mounted_text(mount: Path, filename: str) -> str:
```
johnbieren: Please import and use https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/authentication.py#L25 instead

scripts/python/tasks/internal/pulp_push_disk_images.py
```diff
@@ -0,0 +1,455 @@
...
+    """Return the stripped contents of *filename* under *mount*."""
+    return (mount / filename).read_text(encoding="utf-8").strip()
+
+
+def check_certificates(
```
johnbieren: Why not use https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/push_artifacts.py#L81 ?

scripts/python/helpers/oras_utils.py
happybhati: Good catch on double-checking this — subprocess failures do propagate (`check=True`). Added a test to make that explicit and a brief comment on the `finally` block. Thanks!

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Agreed — removed `_StepFailure` and switched to `CheckStepError` + `write_failure_result`. Thanks for pointing at the existing tekton helper!

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Done, thanks

scripts/python/tasks/internal/pulp_push_disk_images.py
happybhati: Switched to `push_artifacts._check_cert_expiration` via a small `_validate_certificates` wrapper for the three mounts. Thanks for the pointer!

scripts/python/helpers/oras_utils.py
johnbieren: thanks, the try finally tricked me

johnbieren (dismissed):

twaugh (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-24T14:23:16.673Z |

happybhati:
Rebased onto `main` and resolved conflicts in `oras_utils.py` and `test_oras_utils.py` (merged `oras_resolve` from main with our `oras_pull`). Force-push dismissed approvals — @swickersh @johnbieren could you please re-approve? Thanks!

swickersh (approved):
Is the plan to hold off on the catalog referencing this until we have e2e for this pipeline? 
Or do you have a way to mock up a full disk image release?
These questions don't affect this PR being merged... just the catalog counterpart.

johnbieren:
For next time, I think if you have to rebase in a way that resets approvals, you should get them all again before merging (vs just 1). I know it is a bit of a gray area

happybhati:
> For next time, I think if you have to rebase in a way that resets approvals, you should get them all again before merging (vs just 1). I know it is a bit of a gray area

Ah, my bad, thanks for the feedback. I will definitely keep it in mind next time.

PR #809, opened 2026-06-04
refactor: use git cli instead of GitPython

johnbieren:
This commit drops GitPython in some git functions in favor of using the git cli. In testing, it is showing that GitPython is very memory intensive for some operations, which doesn't work well for tekton tasks.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/vcs/git.py
+++ b/scripts/python/helpers/vcs/git.py
@@ -1,17 +1,15 @@
-"""Host-agnostic Git operations via GitPython."""
+"""Host-agnostic Git operations via the `git` CLI."""
 
 from __future__ import annotations
 
+import os
 import re
 import subprocess
-from collections.abc import Callable, Sequence
+from collections.abc import Sequence
 from pathlib import Path
-from typing import TypeVar
+from typing import IO, Any
 
-import git
-from git.exc import GitCommandError
-
-_T = TypeVar("_T")
+import retry
 
 
 def repository_workdir_name(repository_url: str) -> str:
@@ -22,23 +20,21 @@ def repository_workdir_name(repository_url: str) -> str:
     return base
 
 
-def _append_git_stderr(stderr_path: Path | None, exc: BaseException) -> None:
-    # GitPython does not tee subprocess stderr to a file; append on failure only.
-    if stderr_path is None:
-        return
-    err = getattr(exc, "stderr", None)
-    if err is None:
-        err = str(exc)
-    if not str(err).strip():
-        return
-    safe_text = str(err)
+def _redact_credential_urls(text: str) -> str:
     # Git errors often echo clone URLs with embedded oauth2 / token credentials.
-    safe_text = re.sub(
+    return re.sub(
         r"https://([^/@\s:]+):([^@\s]+)@",
         r"https://\1:[REDACTED]@",
-        safe_text,
+        text,
         flags=re.IGNORECASE,
     )
+
+
+def _append_cmd_stderr(stderr_path: Path | None, message: str) -> None:
+    # CLI git stderr is captured on failure and written here (redacted).
+    if stderr_path is None or not message.strip():
+        return
+    safe_text = _redact_credential_urls(str(message))
     with open(
         stderr_path,
         "a",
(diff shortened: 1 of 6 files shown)


scripts/python/helpers/vcs/git.py
```diff
@@ -140,33 +178,35 @@ def push(
...
             # Stop and raise once `attempt` exceeds *retries*.
             if attempt > retries:
-                code = getattr(push_error, "status", None)
-                if code is None:
-                    code = 1
                 raise subprocess.CalledProcessError(
-                    int(code),
+                    push_error.returncode,
                     f"git push {remote}",
                 ) from push_error
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/helpers/vcs/git.py
```diff
@@ -74,61 +57,116 @@ def clone(
...
-        return repo_dir
-
-    return _with_git_stderr(stderr_path, _clone)
-
-
-def origin_ls_tree_name_only(
+        ],
+        cwd=parent_dir,
+        stderr_path=stderr_path,
+    )
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/helpers/vcs/git.py
```diff
@@ -74,61 +103,116 @@ def clone(
 ) -> Path:
     """
     Shallow clone *clone_url* with sparse checkout of *sparse_dirs* at *revision*.
```
mmalina: Question: When I saw this, my first reaction was: So a function called `clone` is doing shallow clone? What if you want normal clone? Is the assumption that you actually never need history in any of the tasks?

johnbieren: I can't think of anywhere we use the history, but if we come across one that does, we can always update this function

mmalina: sounds good

scripts/python/helpers/vcs/git.py
```diff
@@ -74,61 +103,116 @@ def clone(
...
+    Writes the tree listing to *listing_path* and scans it line-by-line with *pattern*
+    so the full output is never held in the Python process (large monorepos can be
+    tens of MB).
+    """
+    listing_path.parent.mkdir(parents=True, exist_ok=True)
+    _run_git_cmd(
+        [
+            "bash",
+            "-c",
+            "git ls-tree -r --name-only origin/main > " + shlex.quote(str(listing_path)),
```
mmalina: Any reason why not call the command directly and redirect the output?

This what Claude came up with:
```
  Problem: While listing_path is shell-quoted, this approach is unnecessarily complex and risky. The
  code runs bash -c with a command string that includes shell redirection. This is fragile because:
  - It introduces an unnecessary bash dependency
  - The command string construction is error-prone
  - It's better to use subprocess's built-in stdout redirection

  Fix: Write directly to the file using subprocess's stdout parameter:

  result = _run_git_cmd(
      ["git", "ls-tree", "-r", "--name-only", "origin/main"],
      cwd=repo_root,
      stderr_path=stderr_path,
  )
  listing_path.parent.mkdir(parents=True, exist_ok=True)
  listing_path.write_text(result.stdout, encoding="utf-8")
```

scripts/python/helpers/vcs/git.py
```diff
@@ -140,33 +224,38 @@ def push(
...
+            raise RuntimeError("git push rejected") from push_error
+
+    try:
+        retry.retry_with_exponential_backoff(
+            _push_with_rebase,
+            max_attempts=retries + 1,
+            retry_on=RuntimeError,
+            base_sleep_seconds=5,
+        )
+    except RuntimeError as retry_error:
```
mmalina, line 259: From Claude:
```
Incorrect Retry Logic in push() Function (lines 292-341)

  Location: scripts/python/helpers/vcs/git.py:292-341

  The new implementation has a subtle but critical bug in the retry mechanism:

  def _push_with_rebase() -> None:
      try:
          _run_git_cmd(["git", "push", remote, branch], ...)
      except subprocess.CalledProcessError as push_error:
          _run_git_cmd(["git", "pull", "--rebase", remote, branch], ...)
          raise RuntimeError("git push rejected") from push_error

  retry.retry_with_exponential_backoff(
      _push_with_rebase,
      max_attempts=retries + 1,
      retry_on=RuntimeError,
      ...
  )

  Problem: This always does pull --rebase before retrying, even after the last retry. Looking at the
  retry helper, when attempt >= max_attempts, it raises immediately without sleeping. However, the
  _push_with_rebase function unconditionally runs git pull --rebase on every push failure, including
  the final one.

  In the original GitPython code (lines 150-172 in current file), the logic was:
  1. Try push
  2. If rejected, increment counter
  3. If counter > retries, raise (no pull)
  4. Otherwise, pull and retry

  The new code does:
  1. Try push
  2. If rejected, always pull first
  3. Then raise RuntimeError to trigger retry
  4. On final attempt, it will still pull but then raise

  Impact: This means if retries=0 and push fails, it will still do a pointless pull --rebase before
  failing. More importantly, the counter check happens after the pull in the old code but implicitly
  via the retry framework in the new code.

  Fix: The logic needs to check if we're on the last attempt before doing the pull:

  def _push_with_rebase(attempt: int) -> None:
      try:
          _run_git_cmd(["git", "push", remote, branch], ...)
      except subprocess.CalledProcessError as push_error:
          if attempt >= retries + 1:  # Last attempt
              raise
          _run_git_cmd(["git", "pull", "--rebase", remote, branch], ...)
          raise RuntimeError("git push rejected") from push_error

  Or simpler: keep the first push outside the retry loop (like before), and only retry the pull+push
  cycle.
```
To be honest, I am not sure if that's really a big problem to do one extra git pull at the end... So feel free to dismiss.

scripts/python/helpers/vcs/git.py
johnbieren: smart, fixed

scripts/python/helpers/vcs/git.py
johnbieren: yeah not a big deal but easy fix so fixed anyway, thanks

mmalina (approved):

seanconroy2021 (approved):
LGTM - Did it reduce the memory much ?

johnbieren:
> LGTM - Did it reduce the memory much ?

I was getting oomkills at 2.5G before this and now pass with 512Mi

ach912 (approved):

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-08T10:13:00.815Z |

Approved

PR #812, opened 2026-06-05
feat(RELEASE-1989): skopeo helper client

midnightercz:
This commit implements skopeo client supporting copy and inspect operations. Also it provides fake skopeo client which can be injected instead of real skopeo client during tests

### DIFF
--- a/scripts/python/helpers/rsmodels/__init__.py
+++ b/scripts/python/helpers/rsmodels/__init__.py
@@ -1,0 +1,7 @@
+"""rsmodels."""
+
+from .container_image import (  # noqa: F401
+    ContainerImage,
+    ContainerImageConfig,
+    ContainerImageRaw,
+)
--- a/scripts/python/helpers/rsmodels/container_image.py
+++ b/scripts/python/helpers/rsmodels/container_image.py
@@ -1,0 +1,409 @@
+"""Pydantic models representing the metadata of a container image from skopeo inspect."""
+
+from __future__ import annotations
+
+import json
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+from pydantic import BaseModel, Field
+
+
+class LayerData(BaseModel):
+    """Metadata for an individual image layer."""
+
+    media_type: str | None = Field(
+        None, alias="MIMEType", description="MIME type of the layer"
+    )
+    digest: str | None = Field(None, alias="Digest", description="Digest of the layer")
+    size: int | None = Field(None, alias="Size", description="Size of the layer in bytes")
+    annotations: dict[str, str] | None = Field(
+        None, alias="Annotations", description="Layer annotations"
+    )
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+
+class ContainerImage(BaseModel):
+    """Model representing a container image returned by the `skopeo inspect` command."""
+
+    name: str | None = Field(None, alias="Name", description="The image name")
+    tag: str | None = Field(None, alias="Tag", description="The image tag")
+    digest: str | None = Field(None, alias="Digest", description="The image digest")
+    architecture: str | None = Field(
+        None, alias="Architecture", description="The image CPU architecture"
+    )
+    os: str | None = Field(
+        None, alias="Os", description="The target operating system of the image"
+    )
+    created: datetime | None = Field(
+        None, alias="Created", description="The timestamp when the image was created"
+    )
+    docker_version: str | None = Field(
+        None,
+        alias="DockerVersion",
+        description="The version of Docker used to build the image",
+    )
+    labels: dict[str, str] = Field(
+        default_factory=dict, alias="Labels", description="A dictionary of image labels"
+    )
+    layers: list[str] = Field(
+        default_factory=list,
+        alias="Layers",
+        description="A list of layer digests in the image",
+    )
+    layers_data: list[LayerData] = Field(
+        default_factory=list,
+        alias="LayersData",
+        description="Detailed metadata for each layer",
+    )
+    env: list[str] = Field(
+        default_factory=list, alias="Env", description="Environment variables set in the image"
+    )
+    repo_tags: list[str] = Field(
+        default_factory=list,
+        alias="RepoTags",
+        description="Other tags associated with the repository",
+    )
+
+    config: ContainerImageConfig | None = Field(
+        None, alias="Config", description="Container runtime configuration"
+    )
+
+    def __getitem__(self, key: str) -> Any:
+        """Allow dict-like access to model fields by their field name or alias."""
+        if hasattr(self, key) and key in self.model_fields:
+            return getattr(self, key)
+        for field_name, field in self.model_fields.items():
+            if field.alias == key:
+                return getattr(self, field_name)
+        raise KeyError(key)
+
+    def get(self, key: str, default: Any = None) -> Any:
+        """Allow dict-like .get() access."""
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    @classmethod
+    def from_json(cls, json_str: str) -> ContainerImage:
+        """Load ContainerImage from a JSON string.
+
+        Args:
+            json_str: JSON string representation of a ContainerImage
+
+        Returns:
+            ContainerImage instance
+
+        """
+        data = json.loads(json_str)
+        return cls.model_validate(data)
+
+    @classmethod
+    def from_dict(cls, data: dict) -> ContainerImage:
+        """Load ContainerImage from a dictionary.
+
+        Args:
+            data: Dictionary representation of a ContainerImage
+
+        Returns:
+            ContainerImage instance
+
+        """
+        return cls.model_validate(data)
+
+    @classmethod
+    def from_file(cls, file_path: str | Path) -> ContainerImage:
+        """Load ContainerImage from a JSON file.
+
+        Args:
+            file_path: Path to JSON file containing a ContainerImage
+
+        Returns:
+            ContainerImage instance
+
+        """
+        path = Path(file_path)
+        with path.open("r", encoding="utf-8") as f:
+            data = json.load(f)
+        return cls.model_validate(data)
+
+
+class ManifestPlatform(BaseModel):
+    """Platform configuration for a manifest in a manifest list."""
+
+    architecture: str | None = Field(None, description="CPU architecture")
+    os: str | None = Field(None, description="Operating system")
+    variant: str | None = Field(None, description="Architecture variant")
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+
+class ManifestConfig(BaseModel):
+    """Configuration reference in a container manifest."""
+
+    media_type: str | None = Field(
+        None, alias="mediaType", description="MIME type of the config"
+    )
+    size: int | None = Field(None, description="Size of the config in bytes")
+    digest: str | None = Field(None, description="Digest of the config")
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+
+class ManifestLayer(BaseModel):
+    """Layer metadata in a container manifest."""
+
+    media_type: str | None = Field(
+        None, alias="mediaType", description="MIME type of the layer"
+    )
+    size: int | None = Field(None, description="Size of the layer in bytes")
+    digest: str | None = Field(None, description="Digest of the layer")
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+
+class RawManifestEntry(BaseModel):
+    """An individual manifest entry within a manifest list."""
+
+    media_type: str | None = Field(
+        None, alias="mediaType", description="MIME type of the manifest"
+    )
+    size: int | None = Field(None, description="Size of the manifest in bytes")
+    digest: str | None = Field(None, description="Digest of the manifest")
+    platform: ManifestPlatform | None = Field(None, description="Platform details")
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+
+class ContainerImageRaw(BaseModel):
+    """Model representing the raw JSON manifest returned by `skopeo inspect --raw`."""
+
+    schema_version: int | None = Field(
+        None, alias="schemaVersion", description="Schema version of the manifest"
+    )
+    media_type: str | None = Field(
+        None, alias="mediaType", description="MIME type of the manifest or manifest list"
+    )
+    manifests: list[RawManifestEntry] = Field(
+        default_factory=list, description="List of manifests (for manifest lists)"
+    )
+    config: ManifestConfig | None = Field(
+        None, description="Config object (for single manifests)"
+    )
+    layers: list[ManifestLayer] = Field(
+        default_factory=list, description="List of layer objects (for single manifests)"
+    )
+
+    class Config:
+        """Pydantic model configuration."""
+
+        populate_by_name = True
+
+    def __getitem__(self, key: str) -> Any:
+        """Allow dict-like access to model fields by their field name or alias."""
+        if hasattr(self, key) and key in self.model_fields:
+            return getattr(self, key)
+        for field_name, field in self.model_fields.items():
+            if field.alias == key:
+                return getattr(self, field_name)
+        raise KeyError(key)
+
+    def get(self, key: str, default: Any = None) -> Any:
+        """Allow dict-like .get() access."""
+        try:
+            return self[key]
+        except KeyError:
+            return default
+
+    @classmethod
+    def from_json(cls, json_str: str) -> ContainerImageRaw:
+        """Load ContainerImageRaw from a JSON string.
+
+        Args:
+            json_str: JSON string representation of a ContainerImageRaw
+
+        Returns:
+            ContainerImageRaw instance
+
+        """
+        data = json.loads(json_str)
+        return cls.model_validate(data)
+
+    @classmethod
+    def from_dict(cls, data: dict) -> ContainerImageRaw:
+        """Load ContainerImageRaw from a dictionary.
+
+        Args:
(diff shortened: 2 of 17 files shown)


scripts/python/helpers/fake/skopeo.py
```diff
@@ -0,0 +1,371 @@
...
+from __future__ import annotations
+
+import os
+import re
+import yaml
+from pathlib import Path
+from typing import Any, Optional
+from logging import Logger
+
+from rsmodels.secret import Secret
```
theflockers: IMHO this `secret` module could be at the same level as `skopeo`. It seems it can be reused in many other places. One example is the git authentication, which also relies on adding the token to github URL. That could be useful, for instance: `secret.skopeo` or `secret.git`  to prevent leaking these tokens/passowords.

scripts/python/helpers/skopeo.py
```diff
@@ -0,0 +1,354 @@
...
+        self.stderr = stderr
+
+    def __str__(self) -> str:
+        """Return a detailed error message including command and outputs."""
+        return (
+            f"{super().__str__()}\n"
+            f"    Command: {' '.join(self.command)}\n"
+            f"    Return code: {self.returncode}\n"
+            f"    Stderr: {self.stderr}"
+        )
```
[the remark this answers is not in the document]
elenagerman: +1

scripts/python/helpers/fake/skopeo.py
```diff
@@ -0,0 +1,371 @@
...
+                "No mock match found",
+                command=self._build_command("inspect", image=image, format=format),
+                returncode=1,
+                stdout="",
+                stderr=error_msg,
+            )
+
+        # Return the configured value
+        return rule["return"]
+
```
elenagerman: FakeSkopeoClient allows inspect rules without a return field. 
The config will load fine, but inspect() will crash at runtime with KeyError: 'return' when that rule matches.
Probably, we have to require 'return' for every inspect rule during config validation (at client init).

scripts/python/helpers/fake/skopeo.py
midnightercz: What prevents devs to import it from rsmodels? Secret doesn't have any functionality so it should belong to models

scripts/python/helpers/fake/skopeo.py
midnightercz: fixed

scripts/python/helpers/skopeo.py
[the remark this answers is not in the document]
midnightercz: fixed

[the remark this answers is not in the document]
midnightercz: fixed

elenagerman (dismissed):

johnbieren:
Can you describe this PR a bit more? I don't really get why we want a "fake" helper. I haven't seen the need for this in any other python tests - we are able to just mock out the functions we need to

midnightercz:
> Can you describe this PR a bit more? I don't really get why we want a "fake" helper. I haven't seen the need for this in any other python tests - we are able to just mock out the functions we need to

Fake skopeo client is for tests in release service catalog. It's way to setup mocks for tests in relase-service-catalog. See how it's used here: 
https://github.com/konflux-ci/release-service-catalog/pull/2241/changes#diff-9fc0eb6a12b5f9a679be692fddd52055cf694ad0efd483077b3f50c7bd561018L47

johnbieren:
> > Can you describe this PR a bit more? I don't really get why we want a "fake" helper. I haven't seen the need for this in any other python tests - we are able to just mock out the functions we need to
> 
> Fake skopeo client is for tests in release service catalog. It's way to setup mocks for tests in relase-service-catalog. See how it's used here: https://github.com/konflux-ci/release-service-catalog/pull/2241/changes#diff-9fc0eb6a12b5f9a679be692fddd52055cf694ad0efd483077b3f50c7bd561018L47

Here is an example of a mocked skopeo client for mocking in tests in release-service-catalog https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/extract-checksums-from-image/tests/mocks/skopeo
This seems way simpler to me than adding 3k+ lines of "fake" helper code, do you see it differently?

midnightercz:
> Here is an example of a mocked skopeo client for mocking in tests in [release-service-catalog](https://issues.redhat.com/browse/RELEASE-service-catalog) https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/extract-checksums-from-image/tests/mocks/skopeo
This seems way simpler to me than adding 3k+ lines of "fake" helper code, do you see it differently?

Yes, obviously. For several reasons: 1. fake skopeo client provides fine control in what you want to check and return. 2. It's "shipped" together with the real client so when I want to use it in other projects I don't have to be worried about if original client changed or not. 3. It has unit tests so I know it works how it should work. So this is all the functionality we "export" to "users" which is in this case release-service-catalog. From user perspective: 4. I don't have to define mocks several time (per task). 5. I don't have to be worried about consistency in mocks over several tasks - not that it would matter that much, but it's little bummer what your mocks work one way in one task but other way in some other task. 6. In my biased opinion it's not easy to understand what https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/extract-checksums-from-image/tests/mocks/skopeo does at the first glance. Even if it's just 22 lines, you need to spend some time to understand what it really does and why.

elenagerman:
Commit message typo:
Assiste-By: claude -> Assisted-by: Claude

scripts/python/helpers/fake/skopeo.py
```diff
@@ -0,0 +1,383 @@
...
+                cmd.extend([f"--{flag_name}", str(value)])
+
+        return cmd
+
+    def inspect(
+        self,
+        image: str,
+        *,
+        format: Optional[str] = None,
+        retry_times: Optional[str] = None,
```
elenagerman: parameter type mismatch between real and fake client: str vs int

scripts/python/helpers/test_skopeo.py
```diff
@@ -197,3 +202,699 @@ def test_copy_nonzero_exit_code() -> None:
         result = skopeo.copy("docker://img:v1", "dir:/tmp/out")
 
     assert result.returncode == 1
+
```
elenagerman, line 50: All new test functions are missing '-> None' annotations as required by AGENT.md

scripts/python/helpers/skopeo.py
```diff
@@ -31,3 +36,357 @@ def copy(
...
+    """
+
+    def __init__(
+        self,
+        *,
+        debug: bool = False,
+        insecure_policy: bool = False,
+        tmpdir: Optional[str] = None,
+        command_timeout: Optional[str] = None,
+        override_arch: Optional[str] = None,
```
elenagerman: probably you need to use | union syntax instead of Optional[] (Required by AGENTS.md)

```suggestion
        tmpdir: str | None = None,
        command_timeout: str | None = None,
        override_arch: str | None = None,
```

scripts/python/helpers/fake/skopeo.py
midnightercz: fixed

scripts/python/helpers/test_skopeo.py
midnightercz: fixed

elenagerman (dismissed):

theflockers (dismissed):

davidmogar (dismissed):

davidmogar:
LGTM but please wait for @johnbieren to take a look as well.

scripts/python/helpers/skopeo.py
johnbieren, line 1: Why does this create a new skopeo copy and inspect without removing the existing ones?

scripts/python/helpers/rsmodels/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""rsmodels."""
```
johnbieren, line 1: what does rsmodel even stand for?

johnbieren:
The commit has `Signed-off-by: Author <a@example.com>` can you fix that please?

midnightercz:
> The commit has `Signed-off-by: Author <a@example.com>` can you fix that please?

fixed

scripts/python/helpers/skopeo.py
midnightercz: this change is in review process quite some time, partially also because I was on PTO. But I updated the code now

scripts/python/helpers/rsmodels/__init__.py
midnightercz: release service, I can update it to skopeo_models or rsu_models if you don't like the current name

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -117,15 +117,13 @@ def inspect_image_created(image_ref: str) -> str | None:
 
     Return ``None`` on any failure.
     """
-    result = skopeo.inspect(image_ref, config=True)
-    if result.returncode != 0:
-        return None
     try:
-        created = json.loads(result.stdout).get("created")
+        result = SkopeoClient().inspect(image_ref, config=True)
```
FilipNikolovski: `SkopeoClient().inspect` expects image_ref to have a transport prefix like `docker://`, so we should either prefix image_ref with docker:// or update the function to automatically add the prefix.

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -470,12 +468,9 @@ def get_manifest_digests(image_ref: str) -> str:
 
     Raise ``RuntimeError`` if the image is not a manifest list.
     """
-    result = skopeo.inspect(image_ref, raw=True)
-    if result.returncode != 0:
-        raise RuntimeError(f"skopeo inspect --raw failed for {image_ref}: {result.stderr}")
-    raw = json.loads(result.stdout)
+    result = SkopeoClient().inspect(image_ref, raw=True)
```
theflockers, line 471: From Claude:

`ContainerImageRaw.manifests is list[RawManifestEntry] | None. If the inspected image is a single manifest (no manifest list), result.manifests is None and this raises TypeError: 'NoneType' is not iterable. The old code used .get("manifests", []) which defaulted safely. Fix:
  result.manifests or [].`

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -696,7 +691,7 @@ def _poll_and_collect(
 
     try:
         digests = get_manifest_digests(internal_copy)
-    except (RuntimeError, json.JSONDecodeError) as e:
+    except SkopeoClientError as e:
```
theflockers: `get_manifest_digests still raises RuntimeError("Index image produced is not multi-arch with a manifest list") when digests are empty, but _poll_and_collect now only catches SkopeoClientError. Previously this was except (RuntimeError, json.JSONDecodeError). If IIB returns an image
  with unexpected manifest media types (e.g. OCI instead of Docker v2), the RuntimeError propagates uncaught and crashes the task instead of returning RunResult(state="failed") gracefully. Either catch RuntimeError here too, or change get_manifest_digests to raise SkopeoClientError
  instead.`

scripts/python/tasks/internal/update_fbc_catalog.py
midnightercz: fixed, updated tests

scripts/python/helpers/skopeo.py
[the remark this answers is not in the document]
elenagerman: +1

scripts/python/tasks/internal/update_fbc_catalog.py
midnightercz: I changed models to have empty containers by default on init

scripts/python/tasks/internal/update_fbc_catalog.py
midnightercz: right, I forgot there's still internal check. Fixed

scripts/python/helpers/skopeo.py
```diff
@@ -1,33 +1,364 @@
...
+        self,
+        *,
+        debug: bool = False,
+        insecure_policy: bool = False,
+        tmpdir: str | None = None,
+        command_timeout: str | None = None,
+        override_arch: str | None = None,
+        override_os: Optional[str] = None,
+        override_variant: Optional[str] = None,
+        logger: Optional[Logger] = None,
```
elenagerman: inconsistent use of Optional[] vs | union syntax.
you could replace all Optional[T] with T | None and remove the Optional import from line 7.

scripts/python/helpers/rsmodels/secret.py
```diff
@@ -0,0 +1,71 @@
+"""Secure handling of secrets in logs and debugging."""
+
+from __future__ import annotations
+
+from typing import Optional, Union
```
elenagerman, line 5: per repo standards should use | syntax, WDYT?

scripts/python/helpers/skopeo.py
```diff
@@ -1,33 +1,364 @@
...
+
+        """
+        cmd = ["skopeo", *self._build_global_flags(), "copy"]
+
+        if all is not None:
+            if all:
+                cmd.append("--all")
+        if preserve_digests is not None:
+            if preserve_digests:
+                cmd.append("--preserve-digests")
```
elenagerman: redundant boolean checks

elenagerman:
missing test for Secret in error path (a command with secrets fails and str(exception) is called)

elenagerman:
tests should use monkeypatch for env vars

scripts/python/helpers/skopeo.py
midnightercz: fixed

scripts/python/helpers/rsmodels/secret.py
midnightercz: fixed

scripts/python/helpers/skopeo.py
midnightercz: fixed

midnightercz:
> missing test for Secret in error path (a command with secrets fails and str(exception) is called)

I added extra tests for Secret class and also tests to check traceback from skopeo client to not contain any leaks

> tests should use monkeypatch for env vars

I don't think there are any env vars in tests

elenagerman (approved):

scripts/python/helpers/rsmodels/__init__.py
johnbieren: I don't necessarily think the dirname has to change but at least on this line it would be nice to say what it means. I don't think we use `rs` to mean `release-service` very often

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -470,12 +468,9 @@ def get_manifest_digests(image_ref: str) -> str:
...
     Raise ``RuntimeError`` if the image is not a manifest list.
     """
-    result = skopeo.inspect(image_ref, raw=True)
-    if result.returncode != 0:
-        raise RuntimeError(f"skopeo inspect --raw failed for {image_ref}: {result.stderr}")
-    raw = json.loads(result.stdout)
+    result = SkopeoClient().inspect(image_ref, raw=True)
     media = "application/vnd.docker.distribution.manifest.v2+json"
-    digests = [m["digest"] for m in raw.get("manifests", []) if m.get("mediaType") == media]
+    digests = [m.digest for m in result.manifests if m.media_type == media]
```
johnbieren, line 473: why are you switching from `m["digest"]` to `m.digest`? Same for mediaType

scripts/python/helpers/rsmodels/container_image.py
johnbieren, line 1: I don't really get this. Why are we maintaining 409 lines for a pydantic model of the output of skopeo inspect? Are we going to maintain 400+ lines for every binary sub command we use? @ach912 also asked on slack about it and said he doesn't understand why we need so much code and you didn't respond. Seeing
```
        result = SkopeoClient().inspect(
            "docker://" + image_reference,
            authfile=auth_file.name,
            raw=True,
            retry_times=3,
        )
```
instead of what we currently have is great. But 1,000 lines of `rsmodels` stuff does not make it worth it IMO. I don't want to maintain code for the output of commands like `skopeo inspect`. We don't maintain skopeo

midnightercz: Here's my opinion: There's cheap code and expensive code. The cheap code is easy to read, easy to test and easy to maintain (by AI or Human). Pydantic models, that's cheap code. Then there's expensive code, which is code containing specific complex logic, code with big cyclomatic complexity, code which contains lot of other expensive code. There are parts where you just have to use complex logic or lots of branching, loop in sequence. I think it's desired to use cheap code as much as possible in those parts especially but as well in general. There are 1xxx lines of code - which is very unlikely going to change ever unless you are adding more functionality to it. That means once you write tests for it, you don't have to modify that. What are the benefits? As a coder (human or AI) I can easily tell what will be the output of skopeo operations. I know specifically what fields can I expected on the output even if skopeo json output provide some extra fields. Even language server will give you hint and you don't have to check it by yourself. While when using basically dict[str, any]. As a coder (human or AI) I have no idea what's in the dict, I need to look up for container schema to know what can be there. And every time I basically need to check what type of object is in the dict as it can be anything (list or config or manifest).

To the @ach912 question on the channel. He asked when this PR was still with fake skopeo client.  I would say it was related to that. And yes I didn't reply as it wasn't question which needed answer - in my opinion. Everyone just expressed how they feel about the idea of fake skopeo  in that thread, there were no questions to be answered. And we discussed that in a meeting we had later that week.

And now the most important: Can you please then tell me everything you don't like on this PR? Just tell me how do you want to approach the thing. Do you like wrapping binaries to clients, do you like having models for data, do you like pydantic or not? I want to know everything so can I finally finish the PR or close it.
And next time I can use that as guidelines for new PRs. As I don't see anything related to that in CONTRIBUTING or AGENTS and it would really speed up development in the future

PR #817, opened 2026-06-08
feat(RELEASE-2494): convert extract-checksums-from-image to python

FilipNikolovski:
Convert the bash script embedded in the extract-checksums-from-image Tekton task to a standalone Python script with unit tests. Adds skopeo.copy() to the helper module and uses tarfile for extraction.

Assisted-by: Claude Code

### DIFF
--- a/scripts/python/tasks/managed/extract_checksums_from_image.py
+++ b/scripts/python/tasks/managed/extract_checksums_from_image.py
@@ -1,0 +1,196 @@
+#!/usr/bin/env python3
+"""Extract SHA256SUMS checksum files from container images for GitHub releases.
+
+Processes container images listed in a snapshot specification, extracts
+binaries to a temporary directory, retains only checksum files (``*SHA256SUMS``)
+in the output, and writes the relative output path to a Tekton result file.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import shutil
+import subprocess
+import tarfile
+import tempfile
+from collections.abc import Callable
+from pathlib import Path
+from typing import Any
+
+import skopeo
+import tekton
+from logger import logger
+
+PROG = "extract_checksums_from_image.py"
+BINARIES_DIR = "binaries"
+
+
+def load_snapshot(snapshot_path: Path) -> dict[str, Any]:
+    """Load and return the snapshot JSON from *snapshot_path*."""
+    if not snapshot_path.is_file():
+        raise ValueError(f"No valid snapshot file was provided: {snapshot_path}")
+    return json.loads(snapshot_path.read_text(encoding="utf-8"))
+
+
+def load_components(data_path: Path) -> list[str]:
+    """Return the component name whitelist from the data file, or ``[]``."""
+    if not data_path.is_file():
+        logger.info("No data JSON was provided.")
+        return []
+
+    data = json.loads(data_path.read_text(encoding="utf-8"))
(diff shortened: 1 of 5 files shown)


scripts/python/tasks/managed/extract_checksums_from_image.py
```diff
@@ -0,0 +1,191 @@
...
+                    result.args,
+                    output=result.stdout,
+                    stderr=result.stderr,
+                )
+
+            extract_binaries_from_layers(tmp_dir, image_binaries_path)
+
+            extracted_dir = tmp_dir / image_binaries_path
+            if extracted_dir.is_dir():
+                copy_to_binaries(extracted_dir, binaries_path)
```
elenagerman: when IMAGE_PATH isn’t found after layer extraction, the script currently skips that component, but the bash task fails on cp "$IMAGE_PATH"/* with set -e. Worth raising there so we don’t silently succeed with empty/partial checksum output (+test for a new code)

scripts/python/tasks/managed/extract_checksums_from_image.py
```diff
@@ -0,0 +1,191 @@
...
+
+        tmp_dir = Path(tempfile.mkdtemp())
+        try:
+            result = copy_image(f"docker://{image_url}", f"dir:{tmp_dir}")
+            if result.returncode != 0:
+                raise subprocess.CalledProcessError(
+                    result.returncode,
+                    result.args,
+                    output=result.stdout,
+                    stderr=result.stderr,
```
seanconroy2021, line 147: if skopeo fails the logs just say something like a non-zero exit status and the actual error message gets hidden. CalledProcessError doesn't include stderr it just shows the exit and command. Might be worth logging result.stderr before raising makes it easier to debug.

scripts/python/tasks/managed/extract_checksums_from_image.py
FilipNikolovski: Good catch, fixed!

scripts/python/tasks/managed/extract_checksums_from_image.py
FilipNikolovski: Updated to include a log message.

seanconroy2021 (dismissed):
lgtm :)

elenagerman (approved):

seanconroy2021 (approved):

scripts/python/tasks/managed/extract_checksums_from_image.py
```diff
@@ -0,0 +1,196 @@
...
+
+import skopeo
+import tekton
+from logger import logger
+
+PROG = "extract_checksums_from_image.py"
+BINARIES_DIR = "binaries"
+
+
+def load_snapshot(snapshot_path: Path) -> dict[str, Any]:
```
johnbieren, line 29: I am fine with this staying how it is for now, but we of course do this in _many_ tasks, so I hope this is moved to a helper at some point and other tasks import it

johnbieren (approved):

ach912 (approved):

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-16T11:37:47.390Z |

Approved

PR #819, opened 2026-06-09
feat(ISV-7116): create script for direct container signing

Allda:
A new script rh_direct_sign_image.py is created to support direct
container signing method. The script will be used in the new Tekton task
that replaces current signing method.

The script resolves all digests for components, verifies if images are
already signed and filter them out. Whatever remains is signed using
internal request and new direct signing pipeline.

    - PyxisSignature/SigningItem dataclasses for structured data
    - Multi-arch digest collection via skopeo inspect --raw
    - Source container resolution via select-oci-auth + oras resolve
    - ConfigMap reading for SIG_KEY_NAME(S)
    - Greedy base64 batching with --batch-max-size and --output args
    - Full unit test suite (34 tests)
    
Assisted-by: Claude Sonnet 4.6
Signed-off-by: Ales Raszka <araszka@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -172,6 +172,7 @@ ENV PATH="$PATH:/home/pubtools-pulp-wrapper"
 ENV PATH="$PATH:/home/pubtools-marketplacesvm-wrapper"
 ENV PATH="$PATH:/home/developer-portal-wrapper"
 ENV PATH="$PATH:/home/publish-to-cgw-wrapper"
+ENV PATH="$PATH:/home/scripts/python/tasks/managed"
 # Flat imports: helpers and task scripts must be importable.
 # Tests use the same layout via pyproject [tool.pytest.ini_options] pythonpath.
 # Keep /home for other modules (e.g. pyxis, sbom) that expect it.
--- a/scripts/python/helpers/oras_utils.py
+++ b/scripts/python/helpers/oras_utils.py
@@ -4,7 +4,36 @@ from __future__ import annotations
 
 import re
 import subprocess
+import tempfile
 from pathlib import Path
+
+from subprocess_cmd import run_cmd
+
+
+def oras_resolve(reference: str) -> str:
+    """Resolve the digest of an OCI image reference using oras.
+
+    Obtains registry credentials via ``select-oci-auth``, writes them to a
+    temporary auth file, then runs ``oras resolve`` and returns the digest.
+
+    Raises ``RuntimeError`` if oras resolve exits non-zero.
+    """
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json") as auth_file:
+        select_auth = run_cmd(["select-oci-auth", reference])
+        auth_file.write(select_auth.stdout)
+        auth_file.flush()
+
+        result = run_cmd(
+            ["oras", "resolve", "--registry-config", auth_file.name, reference],
+            check=False,
+        )
+
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"oras resolve failed for {reference!r} (exit {result.returncode}):"
+            f" {result.stderr.strip()}"
+        )
+    return result.stdout.strip()
 
 
 def oras_login(registry: str, username: str, password: str) -> None:
--- a/scripts/python/helpers/subprocess_cmd.py
+++ b/scripts/python/helpers/subprocess_cmd.py
@@ -26,7 +26,7 @@ def run_cmd(
     """Run *cmd*; capture stdout as text; optionally append stderr to *stderr_path*."""
     # Child must inherit pod env (PATH, KUBECONFIG, etc.); only overlay *env*.
     merged: dict[str, str] = {**os.environ, **dict(env or {})}
-    err_f: Any = subprocess.DEVNULL
+    err_f: Any = subprocess.PIPE
     fh: Any = None
     try:
         if stderr_path is not None:
--- a/scripts/python/tasks/managed/rh_direct_sign_image.py
+++ b/scripts/python/tasks/managed/rh_direct_sign_image.py
@@ -1,0 +1,811 @@
+#!/usr/bin/env python3
+"""Prepare container signing batches for Konflux release snapshot components."""
+
+from __future__ import annotations
+
+import argparse
+import base64
+import json
+import logging
+import re
+import tempfile
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+import pyxis
+from kubectl import get_configmap
+from logger import logger as LOGGER
+from oras_utils import oras_resolve
+from subprocess_cmd import run_cmd
+
+PYXIS_INSTANCE_MAP = {
+    "production": "https://graphql-pyxis.api.redhat.com/graphql/",
+    "production-internal": "https://graphql.pyxis.engineering.redhat.com/graphql/",
+    "stage": "https://graphql-pyxis.preprod.api.redhat.com/graphql/",
+    "stage-internal": "https://graphql.pyxis.stage.engineering.redhat.com/graphql/",
+}
+
+SINGLE_MANIFEST_MEDIA_TYPES = {
+    "application/vnd.oci.image.manifest.v1+json",
+    "application/vnd.docker.distribution.manifest.v2+json",
+}
+
+
+@dataclass(frozen=True)
+class PyxisSignature:
+    """An existing signature record returned from Pyxis."""
+
+    reference: str
+    sig_key_id: str
+
+
+@dataclass
+class SigningItem:
+    """A single image reference, digest, and signing key that needs to be signed."""
+
+    reference: str  # e.g. "registry.redhat.io/myproduct/myrepo:v1.0"
+    digest: str  # e.g. "sha256:abc123"
+    repository: str  # stripped repo used for Pyxis lookup, e.g. "myproduct/myrepo"
+    key: str  # signing key ID, e.g. "redhate2etesting"
+
+
+_TASK_LABEL = "internal-services.appstudio.openshift.io/group-id"
+_PIPELINERUN_LABEL = "internal-services.appstudio.openshift.io/pipelinerun-uid"
+_INTENTION_LABEL = "internal-services.appstudio.openshift.io/intention"
+
+
+@dataclass(frozen=True)
+class SubmitConfig:
+    """Configuration for submitting signing batches via the internal-request CLI."""
+
+    pipeline: str
+    pipeline_image: str
+    requester: str
+    pyxis_ssl_cert_secret_name: str
+    pyxis_graphql_url: str
+    kerberos_keytab_secret: str
+    kerberos_keytab: str
+    kerberos_principal: str
+    signing_repo: str
+    signing_revision: str
+    service_account: str
+    request_timeout: str
+    pipeline_timeout: str
+    task_timeout: str
+    task_id: str
+    pipelinerun_uid: str
+    concurrent_limit: int
+    intention: str
+
+
+def validate_file(arg: str) -> Path:
+    """Return a Path for *arg* if the file exists, otherwise raise FileNotFoundError.
+
+    Intended for use as an argparse ``type`` validator.
+
+    Args:
+        arg: File path string provided on the command line.
+
+    Returns:
+        Resolved Path object pointing to the existing file.
+
+    Raises:
+        FileNotFoundError: If the path does not point to an existing file.
+
+    """
+    if (file := Path(arg)).is_file():
+        return file
+    raise FileNotFoundError(arg)
+
+
+def setup_argparser() -> argparse.ArgumentParser:
+    """Build and return the CLI argument parser.
+
+    Returns:
+        Configured argument parser with all signing preparation arguments.
+
+    """
+    parser = argparse.ArgumentParser(description="Prepare container signing configuration.")
+    parser.add_argument(
+        "--pyxis-server",
+        required=True,
+        choices=PYXIS_INSTANCE_MAP.keys(),
+        help="Pyxis server instance to use",
+    )
+    parser.add_argument(
+        "--snapshot", required=True, type=validate_file, help="Konflux release snapshot path"
+    )
+    parser.add_argument(
+        "--data-file", required=True, type=validate_file, help="Konflux data file path"
+    )
+    parser.add_argument(
+        "--sign-registry-access-file",
+        required=True,
+        type=validate_file,
+        help="File listing repositories that require registry-access signing (one per line)",
+    )
+    parser.add_argument(
+        "--output",
+        type=Path,
+        default=None,
+        help="Directory where batch files are written (one per batch); omit to skip",
+    )
+    parser.add_argument(
+        "--batch-max-size",
+        type=int,
+        default=14 * 1024,
+        help="Maximum size in bytes of each base64-encoded batch (default: %(default)s)",
+    )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Enable debug logging",
+    )
+
+    submit = parser.add_argument_group(
+        "request submission (active when --submit-requests is set)"
+    )
+    submit.add_argument(
+        "--submit-requests",
+        action="store_true",
+        help="Submit signing batches via internal-request after collecting them",
+    )
+    submit.add_argument(
+        "--pipeline",
+        default="container-signing",
+        help="Internal pipeline name for signing (default: %(default)s)",
+    )
+    submit.add_argument(
+        "--pipeline-image",
+        required=True,
+        help="Container image override for the signing pipeline",
+    )
+    submit.add_argument(
+        "--requester",
+        required=True,
+        help="Name of the user requesting signing, for auditing",
+    )
+    submit.add_argument(
+        "--request-timeout",
+        default="1800",
+        help="InternalRequest timeout in seconds (default: %(default)s)",
+    )
+    submit.add_argument(
+        "--pipeline-timeout",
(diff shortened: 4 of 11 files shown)


scripts/python/tasks/managed/rh_direct_sign_image.py
```diff
@@ -0,0 +1,610 @@
+#!/usr/bin/env python3
```
FilipNikolovski, line 1: I assume this script replaces the `sign-image` step from `rh-sign-image` managed task? If so, we should move this to the scripts/python/tasks/managed dir, and rename it to sign_image.py.

scripts/python/tasks/internal/rh_direct_sign_image.py
```diff
@@ -0,0 +1,610 @@
...
+    )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Verbose output",
+    )
+    return parser
+
+
+def setup_logger(level: int = logging.INFO):
```
FilipNikolovski: Use the logger library from `helpers`.

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+    if not pyxis_url:
+        raise ValueError(
+            "Invalid pyxisServer parameter. Only 'production', 'production-internal', "
+            "'stage-internal' and 'stage' allowed."
+        )
+    LOGGER.debug("Resolved pyxis server '%s' to URL: %s", pyxis_server, pyxis_url)
+    return pyxis_url
+
+
+def run_subprocess(cmd: list[str]) -> subprocess.CompletedProcess:
```
FilipNikolovski: Usually these types of functions, we move to `helpers`. Check if there's already similar functionality to this one and use that. If not please move it and add additional tests for it.

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+    try:
+        return subprocess.run(cmd, capture_output=True, text=True, check=True)
+    except subprocess.CalledProcessError as e:
+        LOGGER.error(
+            "Command %s failed (exit %d): %s", " ".join(cmd), e.returncode, e.stderr.strip()
+        )
+        raise
+
+
+def get_configmap(name: str) -> dict[str, Any]:
```
FilipNikolovski: Same here, check the helpers dir.

scripts/python/tasks/internal/rh_direct_sign_image.py
```diff
@@ -0,0 +1,610 @@
...
+
+    Returns:
+        List of digest strings starting with the top-level digest.
+
+    """
+    top_level_digest = image_reference.split("@", 1)[1]
+
+    result = run_subprocess(
+        [
+            "skopeo",
```
FilipNikolovski: We have a helper for skopeo, but I think we need to extend it to support the other flags like --no-tags etc.

scripts/python/tasks/internal/rh_direct_sign_image.py
```diff
@@ -0,0 +1,610 @@
...
+    sha = reference_container_image.split("@sha256:", 1)[1]
+    source_reference = f"{source_repo}:sha256-{sha}.src"
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json") as auth_file:
+        select_auth = run_subprocess(["select-oci-auth", source_reference])
+        auth_file.write(select_auth.stdout)
+        auth_file.flush()
+
+        result = run_subprocess(
+            ["oras", "resolve", "--registry-config", auth_file.name, source_reference]
```
FilipNikolovski: I believe we have helpers for oras as well.

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+    LOGGER.info(
+        "Found %d signing candidates for component %s",
+        len(items),
+        component.get("name"),
+    )
+
+    return items
+
+
+def main():
```
FilipNikolovski: For main() use try/catch to catch all Exceptions, the script must not fail unexpectedly (check the other scripts e.g. https://github.com/konflux-ci/release-service-utils/blob/ca7e5b1317d266e5e10a668ff28948d3bd2d82c5/scripts/python/tasks/internal/create_advisory.py#L613 on how they write the results etc.)

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+    log_level = logging.DEBUG if args.verbose else logging.INFO
+    setup_logger(level=log_level)
+
+    pyxis_url = PYXIS_INSTANCE_MAP[args.pyxis_server]
+    LOGGER.info("Using Pyxis instance URL: %s", pyxis_url)
+
+    sign_registry_access_repos = set(args.sign_registry_access_file.read_text().splitlines())
+
+    data_file = json.loads(args.data_file.read_text())
+    snapshot = json.loads(args.snapshot.read_text())
```
FilipNikolovski: Check if the files exist before proceeding here.

pyxis/prepare_container_signing.py
FilipNikolovski:
```suggestion
def main() -> int:
```

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+        level: Logging level (e.g. logging.INFO, logging.DEBUG).
+
+    """
+    log_format = "%(asctime)s [%(name)s] %(levelname)s %(message)s"
+    stream_handler = logging.StreamHandler(sys.stdout)
+    stream_handler.setLevel(level)
+    logging.basicConfig(level=level, format=log_format, handlers=[stream_handler])
+
+
+def get_pyxis_instance(pyxis_server: str) -> str:
```
FilipNikolovski: This function is only used in the tests, I suggest we either move it there, or remove it completely.

FilipNikolovski (changes requested):

pyxis/prepare_container_signing.py
```diff
@@ -0,0 +1,610 @@
...
+        [
+            "skopeo",
+            "inspect",
+            "--retry-times",
+            "3",
+            "--no-tags",
+            "--raw",
+            f"docker://{image_reference}",
+        ]
+    )
```
elenagerman: skopeo inspect was called without credentials. Konflux snapshot containerImage pullspecs usually point at private Quay repos, and the release pod’s docker config often has repo-specific tokens, not just registry-wide ones. Skopeo doesn’t read those correctly on its own; you need select-oci-auth to build an auth file and pass it with --authfile.
We already do exactly that in [extract_artifacts.py ](https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/extract_artifacts.py#L138-L168) for the same containerImage field. Without the fix, signing prep could fail on private images even though artifact extraction works.

scripts/python/tasks/managed/rh_direct_sign_image.py
Allda: This script doesn't replace the signing step; this serves as a partial replacement of existing bash code that mainly talks to Pyxis to find out existing signatures. Signing itself is handled in a separate pipeline and repository.

pyxis/prepare_container_signing.py
Allda: The snapshot and data files are checked in the Tekton tasks before it is passed to the script. Do you want to check it twice?

pyxis/prepare_container_signing.py
Allda: I built it based on the current bash code, which uses skopeo without specifying any repository-specific auth file - https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/rh-sign-image/rh-sign-image.yaml#L274. In the task, there is a note that what you're describing is valid for oras -https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/rh-sign-image/rh-sign-image.yaml#L302. This script mirrors the same features. Do you want me to change that logic?

scripts/python/tasks/managed/rh_direct_sign_image.py
FilipNikolovski: There is already initiative now to replace the tekton tasks bash scripts with python, and there's already a ticket for replacing the rh-sign-image task: https://redhat.atlassian.net/browse/RELEASE-2521. If we partially replace the logic now, we would have to go back and potentially redo this whole work?

So I'm not sure what's the path forward here, @davidmogar @johnbieren can you share some input on this please?

Allda: I may not have been clear here. The task `rh-sign-image` is going to be deprecated and replaced by the new task `rh-direct-sign-image` which uses a different mechanism for talking with signing pipeline. The [RELEASE-2521](https://redhat.atlassian.net/browse/RELEASE-2521) can be dropped in favour of tasks from https://redhat.atlassian.net/browse/ISV-6620. I'll cover both Python code and tekton task as part of this jira.

I'll move this script into the corresponding loccation in scripts/python/tasks/managed.

davidmogar: Thanks for letting us know @Allda. I'm closing RELEASE-2521 then.

Allda: I pushed a new version that moves the code into the requested location and re-writes all bash code into python.

pyxis/prepare_container_signing.py
Allda: I added the additional authfile to the skopeo with right output based on select-oci-auth

elenagerman (dismissed):

scripts/python/helpers/subprocess_cmd.py
```diff
@@ -26,7 +26,7 @@ def run_cmd(
     """Run *cmd*; capture stdout as text; optionally append stderr to *stderr_path*."""
     # Child must inherit pod env (PATH, KUBECONFIG, etc.); only overlay *env*.
     merged: dict[str, str] = {**os.environ, **dict(env or {})}
-    err_f: Any = subprocess.DEVNULL
+    err_f: Any = subprocess.PIPE
```
FilipNikolovski, line 29: This changes the behavior in all scripts, we should verify that this won't introduce any breaking changes.

scripts/python/tasks/managed/rh_direct_sign_image.py
```diff
@@ -0,0 +1,813 @@
...
+    top_level_digest = image_reference.split("@", 1)[1]
+
+    with tempfile.NamedTemporaryFile(mode="w", suffix=".json") as auth_file:
+        select_auth = run_cmd(["select-oci-auth", image_reference])
+        auth_file.write(select_auth.stdout)
+        auth_file.flush()
+
+        result = run_cmd(
+            [
+                "skopeo",
```
FilipNikolovski, line 265: We have a skopeo helper functions, can we use it here? I assume we'll need to tweak the code a little bit to allow for more flags to be passed on.

scripts/python/tasks/internal/rh_direct_sign_image.py
FilipNikolovski: This was marked resolved, but there were no changes to this line, can you add an explanation why we're keeping this instead of using a helper function?

FilipNikolovski (commented):
Is this script intended to be used as a new catalog task like rh-direct-sign-image, or is it meant to replace the existing rh-sign-image task in-place?

Also, please update the commit message and the PR description to reflect the new changes.

Allda:
> Is this script intended to be used as a new catalog task like rh-direct-sign-image, or is it meant to replace the existing rh-sign-image task in-place?
> 
> Also, please update the commit message and the PR description to reflect the new changes.

This is going to be used in the new task that is already defined here: https://github.com/konflux-ci/release-service-catalog/pull/2301/changes

And I updated the commit message + PR title, and summary.

scripts/python/tasks/managed/rh_direct_sign_image.py
Allda: I decided to keep it here and reuse the existing helper function `run_cmd` that handles the subprocess commands instead of calling a custom subprocess. When I checked existing oras and skopeo helper functions, all of them are quite inconsistent, and each runs differently and raises different errors. This way, it is unified and uses the same mechanism. If you want, I can move it to a helper function, but it would be good to first unify the interface for running external commands.

scripts/python/helpers/subprocess_cmd.py
Allda: I checked it before, and this function was previously used in a single place, `create_advisory.py` where it uses a dedicated stderr file, so this change doesn't affect any existing code.

scripts/python/tasks/internal/rh_direct_sign_image.py
Allda: I moved it to the oras helper module.

scripts/python/tasks/managed/rh_direct_sign_image.py
FilipNikolovski: Ok let's leave it as-is for now, it's not a blocker.

FilipNikolovski (dismissed):
lgtm

scripts/python/tasks/managed/rh_direct_sign_image.py
```diff
@@ -0,0 +1,803 @@
...
+        "-p",
+        f"taskGitRevision={config.signing_revision}",
+        "-l",
+        f"{_TASK_LABEL}={config.task_id}",
+        "-l",
+        f"{_PIPELINERUN_LABEL}={config.pipelinerun_uid}",
+        "-l",
+        "internal-services.appstudio.openshift.io/rate-limited=true",
+        "-l",
+        "internal-services.appstudio.openshift.io/rate-limiting-group=signing-server",
```
mmalina, line 703: Can you also add the "internal-services.appstudio.openshift.io/intention" label like I am adding here? https://github.com/konflux-ci/release-service-catalog/pull/2294/changes#diff-1845fd4c6eea2321a0d785d50d8a079a8afc187730a4f6be21c5c2c9f5082d7fR196

Allda: Added

scripts/python/tasks/managed/rh_direct_sign_image.py
```diff
@@ -0,0 +1,811 @@
...
+        variables = {
+            "repository": repository,
+            "manifest_digest": manifest_digest,
+            "page": page,
+            "page_size": page_size,
+        }
+        data = pyxis.graphql_query(
+            graphql_api, {"query": SIGNATURES_GRAPHQL_QUERY, "variables": variables}
+        )
+        page_data = data["find_signatures"]["data"]
```
lslebodn, line 365: Is it intentional that error is not checked from the response?

Allda: The error is already checked by the Pyxis module and raises an exception.

lslebodn: fair enough

FilipNikolovski (dismissed):

scripts/python/helpers/oras_utils.py
```diff
@@ -4,8 +4,37 @@
 
 import re
 import subprocess
+import tempfile
 from pathlib import Path
 
+from subprocess_cmd import run_cmd
+
+
+def oras_resolve(reference: str) -> str:
```
johnbieren, line 13: Can you add a test for this?

johnbieren (commented):
I think some stuff from `rh_direct_sign_image.py` should be in helpers, but we can deal with that later

scripts/python/helpers/oras_utils.py
Allda: I forgot to push the test file. Now it is available. Please check again.

scripts/python/tasks/managed/rh_direct_sign_image.py
mmalina, line 1: This script is gonna be used in the managed task, not the internal task, right? So it should be under scripts/python/tasks/managed

Dockerfile
```diff
@@ -172,6 +172,7 @@ ENV PATH="$PATH:/home/pubtools-pulp-wrapper"
 ENV PATH="$PATH:/home/pubtools-marketplacesvm-wrapper"
 ENV PATH="$PATH:/home/developer-portal-wrapper"
 ENV PATH="$PATH:/home/publish-to-cgw-wrapper"
+ENV PATH="$PATH:/home/scripts/python/tasks/internal"
```
mmalina: I think this can be removed. For one thing, I think the script belongs to the managed dir. But also, it seems that the convention for these task scripts is to have the full path referenced in the task definition. We're not yet aligned on the exact way of calling it though - some call it via `python3 path/to/python/file` and others call it directly via `/path/to/python/file`. We will address that later.

scripts/python/tasks/managed/rh_direct_sign_image.py
Allda: Good catch, I moved it.

Dockerfile
Allda: Once you agree on the convention, can you address this in a separate PR?

mmalina: Yes, but regardless of that, no other task script is in the PATH currently - and there are plenty already. So once we agree on something, it will be only a catalog change. Having it in PATH would be a third option. So I think it makes sense to puck one of the existing ways.

johnbieren (approved):

twaugh (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-23T15:51:55.121Z |

PR #825, opened 2026-06-15
feat(RELEASE-2472): convert check-labels task to python

querti:
Replace the bash implementation with a single python script. The previous two steps (enforce-name-label, enforce-cpe-label) each looped over all components separately. The new script combines both checks into a single pass over the components. This is a slight change in behavior, but should not matter in practice.

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/check_labels.py
+++ b/scripts/python/tasks/managed/check_labels.py
@@ -1,0 +1,222 @@
+#!/usr/bin/env python3
+"""Validate container image labels against expected values from snapshot and data files.
+
+Check that each image component's ``name`` label matches either its
+``canonicalName`` or the repository URL, and that its ``cpe`` label matches
+the ``releaseNotes.cpe`` value from the data file.
+
+With ``--enforce true``, mismatches cause a non-zero exit. Without it, mismatches
+are logged as warnings and the script exits successfully.
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+from pathlib import Path
+from typing import Any
+
+from file import load_json_dict
+from logger import logger
+
+PROG = "check_labels.py"
+
+IMAGE_MEDIA_TYPES = frozenset(
+    {
+        "application/vnd.oci.image.config.v1+json",
+        "application/vnd.docker.container.image.v1+json",
+    }
+)
+
+
+class LabelValidationError(Exception):
+    """Raise when a label validation fails in enforce mode."""
+
+
+def derive_name_from_url(url: str) -> str:
+    """Derive the namespace/repo path from a container image URL.
+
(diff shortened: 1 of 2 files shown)


scripts/python/tasks/managed/check_labels.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python3
+"""Validate container image labels against expected values from snapshot and data files.
+
+Check that each image component's ``name`` label matches either its
+``canonicalName`` or the repository URL, and that its ``cpe`` label matches
+the ``releaseNotes.cpe`` value from the data file.
+
+With ``--enforce``, mismatches cause a non-zero exit. Without it, mismatches
```
swickersh: "--enforce true" since it's not a flag

scripts/python/tasks/managed/check_labels.py
```diff
@@ -0,0 +1,240 @@
...
+    if not snapshot_path.is_file():
+        logger.error("No valid snapshot file was provided: %s", snapshot_path)
+        return 1
+
+    if not data_path.is_file():
+        logger.error("No valid data file was provided: %s", data_path)
+        return 1
+
+    snapshot = json.loads(snapshot_path.read_text(encoding="utf-8"))
+    data = json.loads(data_path.read_text(encoding="utf-8"))
```
swickersh: Would it be worthwhile to use the json helper https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/file.py#L13 ?

swickersh (commented):
couple small nits. lgtm

scripts/python/tasks/managed/check_labels.py
querti: Fixed!

scripts/python/tasks/managed/check_labels.py
querti: Fixed!

swickersh (dismissed):

johnbieren (dismissed):
nice improvement from the bash script

seanconroy2021 (dismissed):
LGTM

twaugh (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-18T14:42:20.333Z |

querti:
I'd like to request a re-approval, as I've changed how exception handling is done. (more details https://github.com/konflux-ci/release-service-utils/pull/831#discussion_r3438299713)

seanconroy2021 (approved):

johnbieren (approved):

PR #831, opened 2026-06-16
feat(RELEASE-2473): add cleanup-internal-requests Python script

querti:
Remove the pre-delete CRD existence check (kubectl api-resources) that
the original bash script used. The check is redundant because kubectl
delete already fails gracefully when the resource type does not exist
on the cluster. Attempting the delete directly simplifies the logic and
removes a potential source of false negatives where the check itself
could fail for unrelated reasons, silently skipping cleanup.

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/cleanup_internal_requests.py
+++ b/scripts/python/tasks/managed/cleanup_internal_requests.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+"""Delete InternalRequest CRs associated with a specific PipelineRun."""
+
+from __future__ import annotations
+
+import argparse
+import subprocess
+
+from logger import logger
+
+PROG = "cleanup_internal_requests.py"
+
+LABEL_KEY = "internal-services.appstudio.openshift.io/pipelinerun-uid"
+
+
+def cleanup(pipeline_run_uid: str) -> None:
+    """Delete InternalRequest resources matching the PipelineRun UID."""
+    label_selector = f"{LABEL_KEY}={pipeline_run_uid}"
+    result = subprocess.run(
+        [
+            "kubectl",
+            "delete",
+            "internalrequest",
+            "-l",
+            label_selector,
+        ],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        raise RuntimeError(
+            f"kubectl delete failed (exit {result.returncode}): {result.stderr.strip()}"
+        )
+    logger.info("kubectl delete output: %s", result.stdout.strip())
+
+
+def run(pipeline_run_uid: str) -> None:
+    """Orchestrate cleanup: skip when UID is empty, otherwise delete."""
+    if not pipeline_run_uid:
+        logger.info("No pipelineRunUid provided, skipping cleanup")
+        return
+
+    cleanup(pipeline_run_uid)
+
+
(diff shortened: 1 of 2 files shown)


scripts/python/tasks/managed/cleanup_internal_requests.py
```diff
@@ -0,0 +1,93 @@
...
+    """Check whether the InternalRequest CRD is registered on the cluster."""
+    result = subprocess.run(
+        ["kubectl", "api-resources"],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+    if result.returncode != 0:
+        logger.warning(
+            "kubectl api-resources failed (exit %d)",
```
mmalina: I guess maybe this is acceptable if the original script did the same. But just because if fails doesn't necessarily mean it does not exist, right? So I wonder if the check could be improved.

querti: I have changed the logic so that if this check fails, it is treated as a non-critical error and the delete is attempted anyway. It should be safe since the delete of non-existent resource shouldn't produce any side effects. What do you think?

mmalina: This fails:
```
kubectl delete bla bla
error: the server doesn't have a resource type "bla"
```

This passes:
```
kubectl delete pod -l bla
No resources found
```

So as long as the CRD is on the cluster, it's fine. This makes me think: Maybe we could remove that check completely? Then in the unlikely event that the CRD is not there, the delete command fails and we raise an exception. But even here, I think we should log the output, otherwise you will never know why it failed.

querti: I removed the crd check and simplified the script. WDYT?

mmalina: yeah, that looks good, thanks

mmalina (dismissed):

elenagerman (dismissed):

scripts/python/tasks/managed/cleanup_internal_requests.py
```diff
@@ -0,0 +1,71 @@
...
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Parse arguments and run cleanup."""
+    args = _parse_args(argv)
+    try:
+        run(args.pipeline_run_uid.strip())
+    except RuntimeError as exc:
```
johnbieren: The internal tasks do things this way, but the managed ones didn't (this is what I mentioned in standup today that Martin caught). Can you remove this try except and just let the exception log a traceback if one occurs?

querti: Good point, fixed.

mmalina (approved):

johnbieren (approved):

twaugh (commented):
The commit message talks about removing a check but nothing is removed here.

twaugh (approved):
Looks like this is the prerequisite for https://github.com/konflux-ci/release-service-catalog/pull/2300 -- the comment makes sense in that context.

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-23T13:20:47.446Z |

PR #833, opened 2026-06-16
feat(RELEASE-2460): add disk-image support to artifact helpers

swickersh:
- coincides with https://github.com/konflux-ci/release-service-catalog/pull/2334
    - compress_artifacts: copy disk images as-is; fix multi-file
      per-arch scenario (e.g. ISO + QCOW2 sharing one os+arch)
    - push_unsigned: pass disk images through without unpacking,
      handling both files[] and staged.files[] sources
    - push_artifacts: inject staged.files[] into files[] for CGW
      when disk-image component has no files[] entries
    - extract_artifacts: fail early when disk-image component
      declares a non-linux OS entry, preventing signing failures
    - all helpers: detect disk-image via contentType at both
      component level and contentGateway.contentType

Assisted-by: Cursor AI

### DIFF
--- a/scripts/python/helpers/compress_artifacts.py
+++ b/scripts/python/helpers/compress_artifacts.py
@@ -5,7 +5,8 @@ For each component:
 * Pulls signed macOS and Windows OCI artifacts from Quay into a ``signed/`` directory.
 * Restores supplementary files (readme, license, changelog) that were held during signing.
 * Compresses each file entry into the final deliverable format:
-  - macOS / Linux → ``.tar.gz`` (from ``os/arch/`` directory)
+  - macOS / Linux (non-disk-image) → ``.tar.gz`` (from ``os/arch/`` directory)
+  - Linux disk images (``.qcow2``, ``.iso``) → copied as-is to ``ready_for_distribution/``
   - Windows → ``.zip`` (from ``os/arch/`` directory, extension corrected from
     ``.tar.gz``/``.tar``)
 * Updates ``SNAPSHOT_JSON`` to reflect corrected Windows filenames in ``files[]``.
@@ -35,9 +36,11 @@ import tarfile
 import zipfile
 from pathlib import Path
 
+import disk_image_utils
 import oras_utils
 
 PROG = "compress_artifacts.py"
+
 
 QUAY_SECRET_MOUNT = Path(os.environ.get("QUAY_SECRET_MOUNT", "/mnt/quaySecret"))
 CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
@@ -103,7 +106,12 @@ def _windows_filename(source_filename: str) -> str:
 
 
 def _compress_file_entry(
-    entry: dict, array_name: str, component_dir: Path, ready_dir: Path
+    entry: dict,
+    array_name: str,
+    component_dir: Path,
+    ready_dir: Path,
+    *,
+    is_disk_image_component: bool = False,
 ) -> str:
     """Compress one file entry into ready_dir and return the (possibly normalized) source path.
 
@@ -111,11 +119,12 @@ def _compress_file_entry(
     the archive is created as a ``.zip`` instead of ``.tar.gz``/``.tar``, and the returned
     source path reflects the corrected filename so the snapshot can be updated accordingly.
 
-    Raises RuntimeError on failure (missing source, unknown OS, or empty arch directory).
+    Files are copied directly to ``ready_dir`` (without archiving) when either:
+    - *is_disk_image_component* is True (set when contentType: disk-image), or
+    - the filename has an unambiguous disk-image suffix (.qcow2, .iso, .iso.gz,
+      .raw.gz, .vhd.gz).
 
-    Note: all files are currently compressed regardless of type. ISOs should be
-    passed through as-is rather than wrapped in a tarball — this will need to be
-    addressed before ISO delivery is supported.
+    Raises RuntimeError on failure (missing source, unknown OS, or empty arch directory).
     """
     source = entry.get("source")
     if not source:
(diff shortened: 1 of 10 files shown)


swickersh:
Not actually ready for review, but my test images aren't building in draft mode.

scripts/python/helpers/push_artifacts.py
```diff
@@ -365,6 +365,19 @@ def run(exodus_gw_env: str, cgw_hostname: str, cert_expiration_warn_days: int) -
...
+            # Merge staged.files into files[] so publish_to_cgw_wrapper registers them.
+            # NOTE: this intentionally mutates the component dict in-place. It is safe
+            # because the Pulp push and CDN exclusion logic for this component have already
+            # completed above, and the only remaining consumer is publish_to_cgw_wrapper
+            # called below via json.dumps(snapshot).
+            is_disk_image = (
+                component.get("contentType") == "disk-image"
+                or cg.get("contentType") == "disk-image"
+            )
+            if is_disk_image and not component.get("files"):
```
seanconroy2021, line 379: Should this be files not in component instead? Not component.get(files) would also match files: [] which would mean it could be empty.

scripts/python/helpers/extract_artifacts.py
```diff
@@ -227,6 +278,24 @@ def _matches(entry: dict, os_name: str) -> bool:
...
 
+        is_disk_image_component = (
+            component.get("contentType") == "disk-image"
+            or (component.get("contentGateway") or {}).get("contentType") == "disk-image"
+        )
+
+        # Disk images must always be os: linux. Fail immediately if the RPA declares a
+        # disk-image component with macOS or Windows entries — those would route into
+        # the signing workflow and produce failures deep in the pipeline.
+        if is_disk_image_component:
```
seanconroy2021: Could we move this validation to the top of run() before the ThreadPoolExecutor? It only reads the snapshot, but right now it runs after all images are pulled and extracted, so it would be an invalid RPA and it would pull all the images and then fail. 

I think this could also be done in the schema in check-data-keys, so both checks would fail much earlier.

scripts/python/helpers/compress_artifacts.py
```diff
@@ -39,6 +40,23 @@
 
 PROG = "compress_artifacts.py"
 
+# Unambiguous disk-image file suffixes (simple and compound). Files matching
+# these are copied as-is to ready_for_distribution without being wrapped in a
+# tar archive, even when the component does not carry contentType: disk-image.
+# NOTE: .tar.gz is intentionally excluded — it is ambiguous between binary
+# archives and disk images (e.g. GCP images packaged as tarballs). Use
+# contentType: disk-image on the component to handle those cases.
+_DISK_IMAGE_SUFFIXES: frozenset[str] = frozenset(
```
seanconroy2021: Is this not the same as push_unsigned.py? Would it make sense to be its own helper so we only need to update one place.

scripts/python/helpers/push_artifacts.py
swickersh: the comment I had before was wrong/misleading. Updated. And we actually want this behavior. If shipping to contentGateway, then we need to copy staged.files[] into files[] for downstream events.

scripts/python/helpers/extract_artifacts.py
swickersh: good idea. Done, thanks

scripts/python/helpers/compress_artifacts.py
swickersh: Good idea. Extracted to common helper. Thanks for suggestion

seanconroy2021 (dismissed):
LGTM

davidmogar (dismissed):

scripts/python/helpers/test_compress_artifacts.py
```diff
@@ -148,6 +149,162 @@ def test_compress_file_entry_windows(tmp_path: Path, monkeypatch: pytest.MonkeyP
...
+            "files",
+            comp_dir,
+            ready_dir,
+        )
+
+    assert (ready_dir / "rhel-10.0-x86_64-kvm.qcow2").read_bytes() == b"kvm-content"
+    assert (ready_dir / "rhel-10.0-x86_64-boot.iso.gz").read_bytes() == b"iso-content"
+
+
+def test_is_disk_image_file_unambiguous_extensions() -> None:
```
johnbieren: are this and line 250 not just testing `is_disk_image_file`? Isn't that already tested in `test_disk_image_utils.py`?

scripts/python/helpers/test_push_unsigned.py
```diff
@@ -283,6 +284,97 @@ def test_unpack_file_entries_skips_missing_fields(tmp_path: Path) -> None:
...
+        comp_dir,
+        comp_dir / "unsigned",
+    )
+    dest = comp_dir / "linux" / "x86_64" / "rhel-10.0-x86_64-boot.iso"
+    assert dest.exists()
+    assert dest.read_bytes() == b"iso raw content"
+    assert not disk_image.exists()
+
+
+def test_is_disk_image_file_unambiguous_extensions() -> None:
```
johnbieren: are this and line 333 not just testing is_disk_image_file? Isn't that already tested in test_disk_image_utils.py?

scripts/python/helpers/test_compress_artifacts.py
swickersh: Yes and yes. Good catch. Deleted, thanks!

scripts/python/helpers/test_push_unsigned.py
swickersh: deleted. Thanks

johnbieren (dismissed):

johnbieren (approved):

swickersh:
had to rebase

ach912 (approved):
lgtm

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-13T14:19:03.346Z |

Approved

PR #837, opened 2026-06-17
feat(RELEASE-2471): add check-data-keys.py script

johnbieren:
This commit adds a python script to replicate the functionality of the inline bash script in the check-data-keys task in the catalog repo. The task will be updated to use this python module instead.

As part of this work, the schema file was copied into this repository. It will eventually be removed from the catalog repo. The linters for the schema file that currently exist in the catalog repo are added here.

Assisted-By: Cursor

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -58,3 +58,31 @@ jobs:
       - uses: konflux-ci/renovate-config-validator-action@main
         with:
           config_file: .github/renovate.json
+  check-jsonschema:
+    name: Validate json schema file
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - name: Setup python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
+        with:
+          python-version: '3.12'
+      - run: pip install check-jsonschema
+      - name: Run json meta schema check
+        run: check-jsonschema --check-metaschema schemas/dataKeys.json
+  lint-jsonschema:
+    name: Lint json schema file
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - name: Run json schema lint
+        run: |
+          FILE="schemas/dataKeys.json"
+          if ! diff -u "$FILE" <(jq . "$FILE"); then
+            echo "Linting issue in $FILE"
+            echo "To fix: jq . $FILE | sponge $FILE"
+            exit 1
+          fi
+          echo "$FILE is formatted correctly"
--- a/schemas/dataKeys.json
+++ b/schemas/dataKeys.json
@@ -1,0 +1,2185 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "type": "object",
+  "properties": {
+    "enforceContainerFirstSecurityLabels": {
+      "type": "boolean",
+      "default": false,
+      "description": "If true, name and cpe label values will be enforced"
+    },
+    "systems": {
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "systemName": {
+            "type": "string",
+            "description": "The name of the system",
+            "enum": [
+              "releaseNotes",
+              "cdn",
+              "fbc",
+              "sign",
+              "mrrc",
+              "nrrc",
+              "cgw",
+              "jira",
+              "infra",
+              "github",
+              "contentGateway",
+              "pyxis",
+              "mapping",
+              "ootsign",
+              "conforma",
+              "intention",
+              "atlas",
+              "pulp",
+              "cloudMarketplaces"
+            ]
+          },
+          "dynamic": {
+            "type": "boolean",
+            "default": false,
+            "description": "If true, system is optional unless one of its fields are present in the data"
+          }
+        },
+        "required": [
+          "systemName"
+        ]
+      }
+    },
+    "requestType": {
+      "type": "string",
+      "description": "The request type to be used when requesting internal workloads",
+      "enum": [
+        "internal-request",
+        "internal-pipelinerun"
+      ]
+    },
+    "fbc": {
+      "type": "object",
+      "properties": {
+        "request": {
+          "type": "string",
+          "description": "The internal pipeline name to handle requests e.g. iib"
+        },
+        "publishingCredentials": {
+          "type": "string",
+          "description": "The credentials used to publish the image e.g. example-fbc-publishing-credentials"
+        },
+        "buildTags": {
+          "type": "array",
+          "description": "The tags to be added to the build e.g. [ 'example-tag-1', 'example-tag-2' ]",
+          "items": {
+            "type": "string"
+          }
+        },
+        "addArches": {
+          "type": "array",
+          "description": "The arches to be added to the build e.g. [ 'x86_64', 'amd64' ]",
+          "items": {
+            "type": "string"
+          }
+        },
+        "hotfix": {
+          "type": "boolean",
+          "description": "Indicates if the build is a hotfix"
+        },
+        "preGA": {
+          "type": "boolean",
+          "description": "Indicates if the build is a preGA"
+        },
+        "stagedIndex": {
+          "type": "boolean",
+          "description": "Indicates if the build is a staged index"
+        },
+        "productName": {
+          "type": "string",
+          "description": "The product name e.g. exampleproduct"
+        },
+        "productVersion": {
+          "type": "string",
+          "description": "The product version e.g v1.0.0"
+        },
+        "buildTimeoutSeconds": {
+          "type": "integer",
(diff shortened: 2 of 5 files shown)


scripts/python/tasks/managed/test_check_data_keys.py
```diff
@@ -0,0 +1,414 @@
+"""Tests for `check_data_keys` task logic."""
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/tasks/managed/test_check_data_keys.py
```diff
@@ -0,0 +1,414 @@
...
+        return cached
+
+    try:
+        body = _download_schema(DEFAULT_SCHEMA_URL)
+        json.loads(body)
+    except (json.JSONDecodeError, requests.RequestException) as exc:
+        pytest.fail(f"could not load catalog schema from {DEFAULT_SCHEMA_URL}: {exc}")
+
+    cached.write_text(body, encoding="utf-8")
+    return cached
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/tasks/managed/check_data_keys.py
```diff
@@ -0,0 +1,151 @@
...
+        msg = "schema URL must be set"
+        raise ValueError(msg)
+    if normalized.startswith(("http://", "https://")):
+        logger.info("Downloading schema from %s", normalized)
+        try:
+            body = http_client.get_text(normalized, timeout=120)
+        except requests.RequestException as exc:
+            msg = f"Failed to download schema file: {normalized}"
+            raise ValueError(msg) from exc
+        dest.write_text(body, encoding="utf-8")
```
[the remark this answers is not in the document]
johnbieren: fixed

schemas/dataKeys.json
```diff
@@ -0,0 +1,2185 @@
...
+              "const": "intention"
+            }
+          }
+        }
+      },
+      "then": {
+        "required": [
+          "intention"
+        ]
+      }
```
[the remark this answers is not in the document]
johnbieren: not changing anything in dataKeys.json in this PR

schemas/dataKeys.json
```diff
@@ -0,0 +1,2185 @@
...
+              "signing"
+            ],
+            "properties": {
+              "sign": {
+                "required": [
+                  "signKey",
+                  "signCASecret"
+                ]
+              }
+            }
```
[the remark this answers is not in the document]
johnbieren: not changing anything in dataKeys.json in this PR

scripts/python/tasks/managed/check_data_keys.py
```diff
@@ -0,0 +1,115 @@
...
+def main() -> int:
+    """Run the check-data-keys workflow; exit non-zero on failure."""
+    try:
+        run_check_data_keys(
+            data_dir=Path(tekton.require_env("PARAM_DATA_DIR")),
+            data_path=Path(tekton.require_env("PARAM_DATA_PATH")),
+            schema_path=file.path_from_env_variable("SCHEMA_FILE", DEFAULT_SCHEMA_PATH),
+            systems_json=os.environ.get("PARAM_SYSTEMS", ""),
+        )
+    except (FileNotFoundError, ValueError, OSError, Exception) as exc:
```
mmalina: All of these inherit Exception, see the hierarchy here: https://docs.python.org/3/library/exceptions.html#exception-hierarchy

So probably just catch Exception? The rest is redundant.

But I also wonder about this pattern that I've seen in many scripts now - why bother catching the exception? I thought that because of the chaining, the log would still contain the full traceback anyway. Turns out I was wrong, because SystemExit behaves differently. In fact, it will hide the traceback.

So it will just print something like:
```
my_tool: Something went wrong inside the function!
```
And that's it. Is that really a good idea? That hides the full traceback and so it makes it harder to see where exactly it failed. Maybe I should bring this up in the team channel for broader discussion?

.github/workflows/lint.yaml
```diff
@@ -58,3 +58,33 @@ jobs:
       - uses: konflux-ci/renovate-config-validator-action@main
         with:
           config_file: .github/renovate.json
+  check-jsonschema:
+    name: Validate json schema file
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6
+      - name: Install check-jsonschema
```
mmalina: Maybe use the setup-python standard action? Apparently it uses caching for the dependencies. e.g.
```
  - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
    with:
      python-version: '3.12'
  - run: pip install check-jsonschema
```
We have that here as well: https://github.com/konflux-ci/release-service-catalog/blob/development/.github/workflows/promote_branch.yaml#L216C36-L216C82

But if you prefer to just copy and paste what was there for now, that's fine with me.

scripts/python/tasks/managed/check_data_keys.py
```diff
@@ -0,0 +1,115 @@
...
+    return data
+
+
+def validate_data_against_schema(schema_path: Path, data_path: Path) -> None:
+    """Validate *data_path* against *schema_path* using check-jsonschema."""
+    regex_impl = RegexImplementation(RegexVariantName.default)
+    checker = SchemaChecker(
+        SchemaLoader(str(schema_path)),
+        InstanceLoader([CustomLazyFile(str(data_path))]),
+        TextReporter(verbosity=1, stream=sys.stderr),
```
mmalina, line 69: Just to be sure: This should print all the specific errors to stderr, right? So that users can see exactly what was wrong.

johnbieren: Here's an example (from the draft catalog PR's unit test)
```
[run-task : check-data-keys] INFO: Loading data from /var/workdir/release/d58c7dc8-c5d1-4429-86ed-9e3dfc9f9aa8/data.json
[run-task : check-data-keys] INFO: Merging 1 required system(s) into data
[run-task : check-data-keys] INFO: Validating /var/workdir/release/d58c7dc8-c5d1-4429-86ed-9e3dfc9f9aa8/data.json against schema /home/schemas/dataKeys.json
[run-task : check-data-keys] Schema validation errors were encountered.
[run-task : check-data-keys]   /var/workdir/release/d58c7dc8-c5d1-4429-86ed-9e3dfc9f9aa8/data.json::$: 'releaseNotes' is a required property
[run-task : check-data-keys] check_data_keys.py: schema validation failed for /var/workdir/release/d58c7dc8-c5d1-4429-86ed-9e3dfc9f9aa8/data.json
```

.github/workflows/lint.yaml
johnbieren: So you're saying remove
```
      - name: Install check-jsonschema
        run: |
          python3 -m venv venv
          source venv/bin/activate
          python3 -m pip install check-jsonschema
      - name: Run json meta schema check
        run: |
          source venv/bin/activate
          check-jsonschema --check-metaschema schemas/dataKeys.json
```
in favor of
```
  - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405  # v6
    with:
      python-version: '3.12'
  - run: pip install check-jsonschema
  - run: check-jsonschema --check-metaschema schemas/dataKeys.json
```
?

scripts/python/tasks/managed/check_data_keys.py
johnbieren: For managed tasks, I think we should have the full traceback. The internal ones need the special treatment to have it return as a result. I think it is worth raising on the channel, yes

johnbieren: I updated this one to not catch the exception

scripts/python/tasks/managed/check_data_keys.py
```diff
@@ -0,0 +1,115 @@
...
+
+def merge_systems_into_data(
+    data: dict[str, Any],
+    systems: list[dict[str, Any]],
+) -> dict[str, Any]:
+    """Append *systems* to the `systems` array in *data*."""
+    existing = data.get("systems")
+    merged = list(existing) if isinstance(existing, list) else []
+    merged.extend(systems)
+    data["systems"] = merged
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/tasks/managed/check_data_keys.py
mmalina: looks good, thanks

.github/workflows/lint.yaml
mmalina: yes, exactly

scripts/python/tasks/managed/check_data_keys.py
mmalina: great, thanks

.github/workflows/lint.yaml
johnbieren: done

mmalina (approved):

twaugh (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-06-24T14:06:41.552Z |

PR #846, opened 2026-06-23
chore: remove deprecated bash task script

johnbieren:
The check-embargoed-cves task was converted to use python some time ago (even the production catalog task uses python now). So, it is safe now to remove the `scripts/bash` dir and references to it, as that was the only bash task script in it.


mmalina (approved):

johnbieren:
Not user facing because no task uses this script. Going ahead and merging

PR #851, opened 2026-06-24
feat(RELEASE-2479): add collect_index_images.py script

johnbieren:
This commit adds a python script to replicate the functionality of the inline bash script in the collect-index-images managed task in the catalog repo. The task will be updated to use this python module instead.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/image_ref.py
+++ b/scripts/python/helpers/image_ref.py
@@ -7,9 +7,63 @@ import re
 
 import http_client
 import requests
+from logger import logger
 
 _QUAY_SHA_TAG = re.compile(r"^[0-9a-f]{40}$")
 _MAX_QUAY_TAG_PAGES = 50
+
+
+def translate_delivery_repo(repo: str) -> list[dict[str, str]]:
+    """Translate a Quay delivery-repo reference to public registry URLs.
+
+    Return two dicts with `repo` and `url` keys: one for `redhat.io` and
+    one for `access.redhat.com`.
+    """
+    if not repo.strip():
+        msg = "Please pass a repo to translate like 'quay.io/redhat-prod/product----repo'"
+        raise ValueError(msg)
+
+    normalized = repo.replace("----", "/")
+    io_url: str
+    access_url: str
+
+    if normalized.startswith("quay.io/redhat-prod/"):
+        io_url = "registry.redhat.io" + normalized[len("quay.io/redhat-prod") :]
+        access_url = "registry.access.redhat.com" + normalized[len("quay.io/redhat-prod") :]
+    elif normalized.startswith("quay.io/redhat-pending/"):
+        io_url = "registry.stage.redhat.io" + normalized[len("quay.io/redhat-pending") :]
+        access_url = (
+            "registry.access.stage.redhat.com" + normalized[len("quay.io/redhat-pending") :]
+        )
+    elif normalized.startswith("quay.io/rh-flatpaks-prod/"):
+        io_url = "flatpaks.registry.redhat.io" + normalized[len("quay.io/rh-flatpaks-prod") :]
+        access_url = (
+            "registry.access.redhat.com" + normalized[len("quay.io/rh-flatpaks-prod") :]
+        )
+    elif normalized.startswith("quay.io/rh-flatpaks-stage/"):
+        io_url = (
+            "flatpaks.registry.stage.redhat.io"
+            + normalized[len("quay.io/rh-flatpaks-stage") :]
+        )
+        access_url = (
+            "registry.access.stage.redhat.com" + normalized[len("quay.io/rh-flatpaks-stage") :]
+        )
+    elif normalized.startswith("quay.io/redhat/"):
+        io_url = "registry.redhat.io" + normalized[len("quay.io/redhat") :]
+        access_url = "registry.access.redhat.com" + normalized[len("quay.io/redhat") :]
+    else:
+        logger.warning(
+            "Repo to translate is not in expected format. If this is not "
+            "an index image, the expected format is: "
+            "quay.io/redhat-[prod,pending]/product----repo",
+        )
+        io_url = normalized
+        access_url = ""
+
+    return [
+        {"repo": "redhat.io", "url": io_url},
+        {"repo": "access.redhat.com", "url": access_url},
+    ]
 
 
 def resolve_quay_digest_to_git_sha(digest: str, container_image: str) -> str | None:
--- a/scripts/python/tasks/managed/collect_index_images.py
+++ b/scripts/python/tasks/managed/collect_index_images.py
@@ -1,0 +1,187 @@
+#!/usr/bin/env python3
+"""Build a Pyxis index-image snapshot JSON from internal-request results."""
+
+from __future__ import annotations
+
+import json
+import re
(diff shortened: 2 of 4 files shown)


scripts/python/tasks/managed/collect_index_images.py
```diff
@@ -0,0 +1,187 @@
...
+        data_dir=data_dir,
+        internal_request_results_file=Path(
+            tekton.require_env("PARAM_INTERNAL_REQUEST_RESULTS_FILE"),
+        ),
+        build_timestamp=tekton.require_env("PARAM_BUILD_TIMESTAMP"),
+        snapshot_path=data_dir / SNAPSHOT_FILENAME,
+        index_image_snapshot_result_path=tekton.result_paths_from_env(
+            "RESULT_INDEX_IMAGE_SNAPSHOT_PATH",
+        )[0],
+    )
```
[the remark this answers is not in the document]
johnbieren: we want hardcoded functions

scripts/python/tasks/managed/collect_index_images.py
```diff
@@ -0,0 +1,187 @@
...
+
+    snapshot_components: list[dict[str, Any]] = []
+    for index, row in enumerate(components_in):
+        if not isinstance(row, dict):
+            msg = f"components[{index}] must be a JSON object"
+            raise ValueError(msg)
+
+        target_index = row.get("target_index")
+        source_index = row.get("index_image_resolved")
+        if not isinstance(target_index, str) or not target_index.strip():
```
querti, line 103: Instead of all these checks, maybe we could use jsonschema to validate the data format at once?

scripts/python/helpers/image_ref.py
```diff
@@ -12,6 +13,60 @@
...
+            + normalized[len("quay.io/rh-flatpaks-stage") :]
+        )
+        access_url = (
+            "registry.access.stage.redhat.com" + normalized[len("quay.io/rh-flatpaks-stage") :]
+        )
+    elif normalized.startswith("quay.io/redhat/"):
+        io_url = "registry.redhat.io" + normalized[len("quay.io/redhat") :]
+        access_url = "registry.access.redhat.com" + normalized[len("quay.io/redhat") :]
+    else:
+        print(
```
querti: Should we use logger instead?

scripts/python/helpers/image_ref.py
```diff
@@ -12,6 +13,60 @@
...
+    """
+    if not repo.strip():
+        msg = "Please pass a repo to translate like 'quay.io/redhat-prod/product----repo'"
+        raise ValueError(msg)
+
+    normalized = repo.replace("----", "/")
+    io_url: str
+    access_url: str
+
+    if normalized.startswith("quay.io/redhat-prod/"):
```
querti, line 30: This approach seems a bit hard to maintain. Claude suggested this as an improvement:

_REPO_TRANSLATIONS = [
    ("quay.io/redhat-prod", "registry.redhat.io", "registry.access.redhat.com"),
    ("quay.io/redhat-pending", "registry.stage.redhat.io", "registry.access.stage.redhat.com"),
    ("quay.io/rh-flatpaks-prod", "flatpaks.registry.redhat.io", "registry.access.redhat.com"),
    ("quay.io/rh-flatpaks-stage", "flatpaks.registry.stage.redhat.io", "registry.access.stage.redhat.com"),
    ("quay.io/redhat", "registry.redhat.io", "registry.access.redhat.com"),
]
for source, io_host, access_host in _REPO_TRANSLATIONS:
    if normalized.startswith(source + "/"):
        suffix = normalized[len(source):]
        io_url = io_host + suffix
        access_url = access_host + suffix
        break

(not tested)

WDYT?

scripts/python/tasks/managed/collect_index_images.py
```diff
@@ -0,0 +1,187 @@
...
+            msg = f"components[{index}].image_digests must be a JSON array"
+            raise ValueError(msg)
+
+        logger.info(
+            "Processing index image %s (%d/%d)",
+            target_index,
+            index + 1,
+            len(components_in),
+        )
+        translated = image_ref.translate_delivery_repo(target_index)
```
querti: Should this be called with repository instead of target_index?

scripts/python/helpers/image_ref.py
johnbieren: I think it is the same difficulty to maintain in either instance

scripts/python/tasks/managed/collect_index_images.py
johnbieren: Not according to https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/collect-index-images/collect-index-images.yaml#L149

scripts/python/helpers/image_ref.py
johnbieren: yes, great catch. fixed

scripts/python/tasks/managed/collect_index_images.py
johnbieren: We'd have to maintain a schema file to do that. I think that is out of scope of this conversion. It may come eventually though - I know Sean has been investigating some stuff on smaller schema files instead of the one large one we currently have

scripts/python/tasks/managed/collect_index_images.py
querti: The function has a message "Please pass a repo to translate like 'quay.io/redhat-prod/product----repo'". I don't think tag is meant to be a part of the input parameter.

FilipNikolovski (commented):
lgtm once @querti comments are resolved.

scripts/python/tasks/managed/collect_index_images.py
johnbieren: I'm not sure what the tag is in the message. Can you say that in a different way (or just let me know what a better error message is and I'll update to that)?

querti: The input parameter of the translate_delivery_repo function is meant to be registry+repo, which makes sense. The message in the function even explcitly mentions this, per my comment above. But this function is actually called with registry+repo+tag, which I think is by mistake and "repository" variable should be used instead of "target_index". The current code still works because how the function truncates it, but that's by accident.

johnbieren: @FilipNikolovski as the author of the line I pasted (that is under question), can you chime in here?

FilipNikolovski: I think it's more correct to use the "repository" variable, but the `translate-delivery-repo` doesn't care about tags, so functionally it would be the same?

johnbieren: Updated. Thanks Lubo for catching that and Fitz for confirming

querti (approved):

davidmogar (approved):

ach912 (approved):
lgtm

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-01T07:24:32.895Z |

Approved

twaugh (approved):

PR #853, opened 2026-06-25
fix: use full shallow clone in process_file_updates

elenagerman:
## Describe your changes
The python port from #775 used sparse shallow clone, but the bash task does a full shallow clone. 
e2e file updates failed because the target file wasn’t present after sparse checkout (yq → exit 1). 

This change makes prepare_repository() use a full shallow clone to match bash. git.clone() still supports sparse shallow clone when sparse_dirs is provided for other callers.

## Relevant Jira

### DIFF
--- a/scripts/python/tasks/internal/process_file_updates.py
+++ b/scripts/python/tasks/internal/process_file_updates.py
@@ -300,21 +300,6 @@ def write_paths_manifest(paths_json: str, temp_dir: Path) -> tuple[Path, list[di
     return update_paths_file, json.loads(paths_json)
 
 
-def sparse_dirs_from_paths(paths_data: list[dict[str, Any]]) -> list[str]:
-    """Return sparse-checkout paths covering every ``path`` entry in *paths_data*."""
-    sparse_dirs: set[str] = set()
-    for entry in paths_data:
-        entry_path = entry.get("path")
-        if not entry_path:
-            continue
-        path = Path(str(entry_path))
-        if path.parent == Path("."):
-            sparse_dirs.add(path.as_posix())
-        else:
-            sparse_dirs.add(path.parent.as_posix())
-    return sorted(sparse_dirs)
-
-
 def prepare_repository(
     repo: str,
     revision: str,
@@ -324,8 +309,7 @@ def prepare_repository(
 ) -> Path:
     """Clone *repo* at *revision*, rebase on *upstream_repo*, and return the repo cwd."""
     logger.info("=== UPDATING %s ON BRANCH %s ===", repo, revision)
-    sparse_dirs = sparse_dirs_from_paths(paths_data)
-    if not sparse_dirs:
+    if not any(entry.get("path") for entry in paths_data):
         raise tekton.CheckStepError(
             "cloning repository",
             ValueError("paths JSON must include at least one path entry"),
@@ -334,7 +318,6 @@ def prepare_repository(
         temp_dir,
         repo,
         revision=revision,
-        sparse_dirs=sparse_dirs,
         shallow=True,
     )
     vcs_git.rebase_onto_remote(
--- a/scripts/python/helpers/vcs/git.py
+++ b/scripts/python/helpers/vcs/git.py
@@ -96,8 +96,9 @@ def clone(
 ) -> Path:
     """Clone *clone_url* into *parent_dir*.
 
-    When *shallow* is true, perform a blob-filtered shallow clone and optional
-    sparse checkout of *sparse_dirs* at *revision*.
+    When *shallow* is true, perform a shallow clone at *revision*. With
+    *sparse_dirs*, use blob-filtered sparse checkout; otherwise clone the full
+    tree at depth 1 (legacy bash ``git_clone_and_checkout`` without ``--sparse-dir``).
 
     Creates *parent_dir* when it does not exist yet.
     """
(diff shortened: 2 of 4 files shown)


theflockers (dismissed):

scripts/python/tasks/internal/process_file_updates.py
mmalina, line 1: Based on my findings, the original version worked exactly the same - it would include literal "\n" in the file, not new line. Are you sure there is any change? Can you share more details? What test failed and how?

elenagerman: You're right. I re-tested the bash flow, and it writes literal \n to the file, same as Python, not real newlines. 
So I dropped that change. The real issue is slightly different: the Python port used a sparse shallow clone, while bash does a full shallow clone. 
I fixed it by matching bash (the fix is needed anyway).
I changed the title and description for this PR to clarify the change.

mmalina: I think it's still worth checking with e2e if this works the same way with the newlines. It seems the only e2e that has a seed defined is this one: https://github.com/konflux-ci/release-service-catalog/blob/development/integration-tests/push-to-addons-registry/resources/managed/rpa.yaml
What's really weird is that the target repo had its last MR created 8 months ago: https://gitlab.cee.redhat.com/rhtap-release/managed-tenants-stage/-/merge_requests/?sort=created_date&state=opened&first_page_size=20

Any idea why? I created a test PR to run this e2e test and see what is going on: https://github.com/konflux-ci/release-service-catalog/pull/2326

mmalina: So it actually worked: https://gitlab.cee.redhat.com/rhtap-release/managed-tenants-stage/-/merge_requests/567/diffs
What's worth noting is that the RPA has:
```
seed: 'indexImage: \nname: my-addon\nrelatedImages: []'
```
And the internal task param is:
```
    - name: paths
      value: '[{"path":"addons/test-product/addonimagesets/stage/my-addon-new.yaml","replacements":[{"key":".indexImage","replacement":"|indexImage:.*|indexImage: |"}],"seed":"indexImage: \\nname: my-addon\\nrelatedImages: []"}]'
```
And the resulting file:
```
indexImage: 
name: my-addon
relatedImages: []
```
So it's totally expected that the internal task param has `\\n`, but this will still result in an actual newline. I definitely recommend you confirm this in the e2e run in the catalog PR.

mmalina: I created a follow up jira to clean up the existing MRs in the target repo: https://redhat.atlassian.net/browse/RELEASE-2686

mmalina: From Claude (and I think he's right):
```
  So the right next steps are:

  1. Ask the author to verify with the push-to-addons-registry e2e test — run it with the Python
  script both with and without _expand_seed_content. That will definitively show whether the fix is
  needed and whether it produces correct output.
  2. If the fix is needed, consider whether _expand_seed_content is the right layer or whether it
  would be cleaner to handle the escaping at the JSON input level — so the seed value arrives as a
  proper string with real newlines, rather than relying on post-hoc expansion of \n sequences. That
  would eliminate the "accidental escaping" problem entirely rather than replacing one accident with
  a deliberate workaround.
```

elenagerman: @mmalina Thanks for checking with the push-to-addons e2e and [#567](https://gitlab.cee.redhat.com/rhtap-release/managed-tenants-stage/-/merge_requests/567); that's very useful!

Anyway, the reason for the current PR is actually a failures in a real catalog PR related to this change: [#2321](https://github.com/konflux-ci/release-service-catalog/pull/2321)
On the seed/\\n point: agreed. The \\n in the logged paths param is just escaping; the file ends up with real newlines, which matches what we'd expect. 
But that doesn't change the fact that we still have a problem with sparse-clone on the no-seed paths (shebert.yml, etc.), not seed handling. 

I'll still keep an eye on push-to-addons in the catalog e2e after the utils image bump, but your run suggests we're fine on the seed side.

On why managed-tenants-stage had no MRs for ~8 months: I definitely not a right person to ask it, but, IMO, most likely the test just doesn't run that often, and when our catalog e2e failed at run-file-updates, nothing new got opened there anyway.

elenagerman: regarding Claude's suggestion: 

- I don't think step 1 is the right next step anymore.

I already dropped _expand_seed_content from the utils side. The push-to-addons e2e run ([#567](https://gitlab.cee.redhat.com/rhtap-release/managed-tenants-stage/-/merge_requests/567)) showed the seeded file has real newlines, and the \\n in the logged paths param is just JSON escaping in the display and not just literal \n in the file. After json.loads, the seed already arrives as a normal string with real newlines, same as bash after jq + echo -e. So an A/B test with/without _expand_seed_content would not tell us much.
The failures seen on PR were on paths with no seed (shebert.yml, etc.). 

On cluster, the same payload succeeded with bash (full shallow clone) and failed with Python (sparse clone). That is the fix I am pursuing, not seed expansion as originally intended (not a correct failure reason assumption).


- On step 2: I agree in principle that post-hoc \n expansion is the wrong layer. But we should not add handling at JSON input either right now, because the input path already works when JSON uses standard \n escapes. Adding expansion would be replacing one accidental behavior with another workaround, and could break cases where \n is meant literally.

elenagerman: +1 on investigating [RELEASE-2686](https://redhat.atlassian.net/browse/RELEASE-2686). 
Thanks a lot for opening it!!!
My read on the 8-month gap: not missing cleanup, but MR reuse
process-file-updates lists open MRs matching "Konflux release", fetches each head, and compares diffs (find_existing_mr_with_same_diff). If one of the existing open MRs already has the same change, it returns that MR URL and does not open a new one. 
e2e does not clean up GitLab MRs: Integration test cleanup removes K8s resources and GitHub test repos; it does not close MRs in managed-tenants-stage. That explains MR accumulation over time, not a disappearing-MR theory.

mmalina: > I'll still keep an eye on push-to-addons in the catalog e2e after the utils image bump, but your run suggests we're fine on the seed side.

My test only proved that the current bash-based task works correctly. So please make sure to double check in your catalog PR that it's still correct.

> If one of the existing open MRs already has the same change, it returns that MR URL and does not open a new one.

Yes, I don't think the test didn't run for 8 months. It's just been reusing the same open MR. I did some cleanup now, but kept the last two MRs. But the jira I created will be used to discuss if the e2e needs changing.

mmalina (approved):

johnbieren (approved):

davidmogar (approved):

ach912 (approved):
lgtm

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-01T07:23:52.664Z |

Approved

PR #855, opened 2026-06-25
fix(RELEASE-1993): fix missing result in update-fbc-catalog

FilipNikolovski:
Always write the iibLog result file even when the log URL is empty, write all five result files on early failure paths, and fetch full build details for reused builds so the logs field is populated.

Assisted-by: Claude Code

### DIFF
--- a/scripts/python/tasks/internal/update_fbc_catalog.py
+++ b/scripts/python/tasks/internal/update_fbc_catalog.py
@@ -17,16 +17,15 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
-import requests
-from requests_kerberos import HTTPKerberosAuth, OPTIONAL
-
 import authentication
 import file
 import http_client
 import iib
+import requests
 import skopeo
 import tekton
 from logger import logger
+from requests_kerberos import OPTIONAL, HTTPKerberosAuth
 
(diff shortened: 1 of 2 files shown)


seanconroy2021 (dismissed):
LGTM

scripts/python/tasks/internal/update_fbc_catalog.py
```diff
@@ -815,13 +850,22 @@ def _write_failure(
     build_state_path: Path,
     exit_code_path: Path,
     reason: str,
+    json_build_info_path: Path | None = None,
+    index_image_digests_path: Path | None = None,
+    iib_log_path: Path | None = None,
 ) -> None:
     """Write failure state to Tekton result files."""
```
johnbieren: Can you update this to say why we write empty content into 3 of the files if provided?

FilipNikolovski: Updated.

johnbieren (approved):

ach912 (approved):

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-02T10:58:42.911Z |

Approved

PR #860, opened 2026-06-29
feat(RELEASE-2480): add collect_registry_token_secret script

happybhati:
This commit adds a python script to replicate the inline bash in the
collect-registry-token-secret managed task in the catalog repo. The task
will be updated to use this python module instead.
- collect_registry_token_secret.py: read data JSON, detect public repos,
  return registry secret name (or empty) via Tekton result
- test_collect_registry_token_secret.py: 11 pytest cases

Assisted-by: Cursor

### DIFF
--- a/scripts/python/helpers/snapshot.py
+++ b/scripts/python/helpers/snapshot.py
@@ -8,6 +8,11 @@ from typing import Any
 import file
 
 
+def _is_truthy(value: Any) -> bool:
+    """Return True for boolean ``True`` or the case-insensitive string ``"true"``."""
+    return value is True or str(value).lower() == "true"
+
+
 def first_component(snapshot_path: Path) -> dict[str, str]:
     """Return fields from the first snapshot `components` entry.
 
@@ -65,3 +70,19 @@ def component_push_source_container(
     if "pushSourceContainer" not in component and default_push_source_container:
         return True
     return False
+
+
+def component_public(data: dict[str, Any], component: dict[str, Any]) -> bool:
+    """Return whether a component should be made public.
+
+    A component is public when its own `public` field is truthy, or when
+    `public` is absent and the mapping-level default (`mapping.defaults.public`)
+    is true.
+    """
+    mapping = data.get("mapping")
+    default = False
+    if isinstance(mapping, dict):
+        defaults = mapping.get("defaults")
+        if isinstance(defaults, dict):
+            default = _is_truthy(defaults.get("public", False))
+    return _is_truthy(component.get("public", default))
--- a/scripts/python/tasks/managed/collect_registry_token_secret.py
+++ b/scripts/python/tasks/managed/collect_registry_token_secret.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+"""Collect the registry token secret name from the release data file."""
+
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import file
+import snapshot
+import tekton
+from logger import logger
+
+
+def is_secret_required(data: dict[str, Any]) -> bool:
+    """Return True when defaults or any component require making repos public.
+
+    An empty component (no ``public`` override) resolves to the mapping-level
+    default, so this also covers the case of ``mapping.defaults.public`` being
+    true with no components at all.
+    """
+    if snapshot.component_public(data, {}):
+        return True
+
+    mapping = data.get("mapping") or {}
+    for component in mapping.get("components") or []:
+        if isinstance(component, dict) and snapshot.component_public(data, component):
+            return True
+
+    return False
+
+
+def collect_registry_token_secret(data: dict[str, Any]) -> str:
+    """Return the registry secret name, or an empty string when not required.
+
+    Raises:
+        KeyError: When a secret is required but ``mapping.registrySecret``
+            is absent.
+
+    """
+    if not is_secret_required(data):
+        logger.info("No repos to make public, so no secret is required. Exiting...")
+        return ""
+
+    mapping = data.get("mapping") or {}
+    return str(mapping["registrySecret"]).strip()
+
+
+def run(data_file: Path) -> str:
+    """Load *data_file* and return the registry secret name (or empty)."""
+    data = file.load_json_dict(data_file)
+    return collect_registry_token_secret(data)
+
+
+def main() -> int:
+    """Read Tekton env vars, resolve the secret name, and write the result."""
+    data_dir = Path(tekton.require_env("PARAM_DATA_DIR"))
+    data_path = Path(tekton.require_env("PARAM_DATA_PATH"))
+    (result_path,) = tekton.result_paths_from_env("RESULT_REGISTRY_SECRET")
+
+    secret = run(data_dir / data_path)
+    result_path.write_text(secret, encoding="utf-8")
+    return 0
+
+
+if __name__ == "__main__":  # pragma: no cover
+    raise SystemExit(main())
--- a/scripts/python/helpers/test_snapshot.py
+++ b/scripts/python/helpers/test_snapshot.py
@@ -9,6 +9,25 @@ import pytest
 import snapshot
 
 
+@pytest.mark.parametrize(
+    ("value", "expected"),
+    [
+        (True, True),
+        ("true", True),
+        ("True", True),
+        ("TRUE", True),
+        (False, False),
+        ("false", False),
+        (None, False),
+        (1, False),
+        ("yes", False),
+    ],
+)
+def test_is_truthy(value: object, expected: bool) -> None:
+    """Only boolean `True` or a case-insensitive `"true"` string is truthy."""
+    assert snapshot._is_truthy(value) is expected
+
+
 def test_first_component_missing_components(tmp_path: Path) -> None:
     """Reject snapshots with no `components` list."""
     path = tmp_path / "snapshot.json"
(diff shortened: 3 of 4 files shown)


scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,85 @@
...
+        ValueError: When a secret is required but ``mapping.registrySecret``
+            is absent from the data file.
+
+    """
+    if not is_secret_required(data):
+        logger.info("No repos to make public, so no secret is required. Exiting...")
+        return ""
+
+    mapping = data.get("mapping") or {}
+    if "registrySecret" not in mapping:
```
seanconroy2021: More of nit but this doesn't catch empty or whitespace registrySecret values if someone sets "" in the data. it'll pass through here.  You could collapse the absent,None,empty or whitespace checks into one and strip the result same way tekton.require_env does it

Something like this: 
```
mapping = data.get("mapping") or {}
secret = mapping.get("registrySecret")
if not secret or not str(secret).strip():
    raise ValueError("Registry secret missing in data JSON file")
return str(secret).strip()
```

happybhati: Good catch — collapsed the checks and added tests for empty/whitespace secrets. Should be good now, thanks!

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,82 @@
...
+from __future__ import annotations
+
+from pathlib import Path
+from typing import Any
+
+import file
+import tekton
+from logger import logger
+
+PROG = "collect_registry_token_secret.py"
```
elenagerman: I didn't find this constant usage

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def _is_public(value: Any) -> bool:
+    """Return True when a mapping ``public`` flag is enabled."""
+    if value is True:
+        return True
+    return str(value).lower() == "true"
+
```
elenagerman: This function handles both boolean and string, but tests only cover the boolean case

elenagerman:
Please consider testing non-dict component entries and empty/nul mapping

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Thanks for catching that! Removed it, wasn't being used anywhere.

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Good point! Added a test for the string `"true"` case — `test_is_secret_required_when_component_public_string `covers that path now.

happybhati:
> Please consider testing non-dict component entries and empty/nul mapping

Thanks for the suggestion! Added tests for all three cases — non-dict/None entries in the components list, empty mapping, and missing mapping key. All covered now.

elenagerman (dismissed):

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,80 @@
...
+from pathlib import Path
+from typing import Any
+
+import file
+import tekton
+from logger import logger
+
+
+def _is_public(value: Any) -> bool:
+    """Return True when a mapping ``public`` flag is enabled."""
```
johnbieren: huh? I don't see `public` being in the function at all

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,80 @@
...
+    if value is True:
+        return True
+    return str(value).lower() == "true"
+
+
+def is_secret_required(data: dict[str, Any]) -> bool:
+    """Return True when defaults or any component require making repos public.
+
+    Mirrors the bash task: ``mapping.defaults.public`` or any
```
johnbieren: the bash task is going away. this is going to be meaningless soon

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,80 @@
...
+    ``mapping.components[*].public`` set to true.
+    """
+    mapping = data.get("mapping") or {}
+    defaults = mapping.get("defaults") or {}
+    if _is_public(defaults.get("public", False)):
```
johnbieren: surely other tasks care about the public setting in a component, no? It yes, I think this should be in the snapshot.py helper

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,80 @@
...
+        ValueError: When a secret is required but ``mapping.registrySecret``
+            is absent, empty, or whitespace-only.
+
+    """
+    if not is_secret_required(data):
+        logger.info("No repos to make public, so no secret is required. Exiting...")
+        return ""
+
+    mapping = data.get("mapping") or {}
+    secret = mapping.get("registrySecret")
```
johnbieren: can't you just do `mapping["registrySecret"]` and let python raise a keyerror if it doesn't exist instead of raising your own ValueError?

scripts/python/tasks/managed/collect_registry_token_secret.py
```diff
@@ -0,0 +1,80 @@
...
+
+    return str(secret).strip()
+
+
+def run(data_file: Path) -> str:
+    """Load *data_file* and return the registry secret name (or empty)."""
+    if not data_file.is_file():
+        raise FileNotFoundError("No valid data file was provided.")
+
+    data = file.load_json_dict(data_file)
```
johnbieren, line 51: please use the helper in file.py. It already raises an exception if the file doesn't exist

johnbieren (changes requested):

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Oops, fair catch — moved it to snapshot.is_public() so the name actually matches what it does now.

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: My bad, slipped through — removed the bash mention.

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Good call, moved it there so other tasks can reuse it too.

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Ha, you're right, way simpler — fixed, letting KeyError raise naturally now.

scripts/python/tasks/managed/collect_registry_token_secret.py
happybhati: Oops, redundant check on my part — removed it, file.load_json_dict() already handles that.

scripts/python/tasks/managed/collect_registry_token_secret.py
johnbieren: The move is good but I still don't see the connection to `public` - all it does is check if an arbitrary parameter passed to it is boolean true or string true

scripts/python/tasks/managed/collect_registry_token_secret.py
johnbieren: This is the part I'd expect to be in the snapshot.py helper. What is now `is_secret_required` because it finds out if public is true or false for the component

scripts/python/helpers/snapshot.py
```diff
@@ -65,3 +70,23 @@ def component_push_source_container(
...
+    mapping = data.get("mapping")
+    if not isinstance(mapping, dict):
+        return False
+    defaults = mapping.get("defaults")
+    if not isinstance(defaults, dict):
+        return False
+    return _is_truthy(defaults.get("public", False))
+
+
+def component_public(component: dict[str, Any], default_public: bool) -> bool:
```
johnbieren: do we care what the default is ever? Isn't it only relevant when finding out if the component has public set to true or false? I feel like line 86 and 75 should combine into one function get_component_public or something

happybhati: Thanks, makes sense — combined into component_public(data, component), resolving the default internally now.

scripts/python/helpers/snapshot.py
```diff
@@ -8,6 +8,11 @@
 import file
 
 
+def _is_truthy(value: Any) -> bool:
```
johnbieren, line 11: missing tests for this

johnbieren (dismissed):

ach912 (dismissed):
lgtm

seanconroy2021 (dismissed):

twaugh (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-20T12:00:20.355Z |

Approved

happybhati:
Sorry for the noise @elenagerman @johnbieren @ach912 @seanconroy2021 — the PR got pulled from the merge queue because gitlint's title-length check failed: GitHub appends (#860) to the squash-merge title, and that pushed it over the 72-char limit. I only shortened the commit title to fix it, no code changes, but it looks like that dismissed the approvals too. Not sure why a title-only tweak did that — apologies again! Could you please re-approve 🙇

elenagerman (approved):

seanconroy2021 (approved):

johnbieren (approved):

ach912 (approved):

PR #866, opened 2026-07-03
feat(RELEASE-2478): convert collect-gh-params to python

FilipNikolovski:
Convert the collect-gh-params managed task to a python script.

Assisted-by: Claude Code

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ pythonpath = [
 
 [dependency-groups]
 dev = [
+    "basedpyright>=1.39.9",
     "black>=25.11.0",
     "flake8>=7.3.0",
     "pytest>=8.4.2",
--- a/scripts/python/tasks/managed/collect_gh_params.py
+++ b/scripts/python/tasks/managed/collect_gh_params.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+"""Collect three parameters for the create-github-release task.
+
+The githubSecret from the Data file, the repository from the
+snapshot file, and the release_version from the binaries of the
+extract-checksums-from-image task.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import tekton
+from file import load_json_dict
+
+
+def validate_input_files(
+    data_file: Path,
+    snapshot_file: Path,
+) -> None:
+    """Raise RuntimeError when any required input file is missing."""
+    for label, path in [("data", data_file), ("snapshot", snapshot_file)]:
+        if not path.is_file():
+            raise RuntimeError(f"No valid {label} file was provided.")
+
+
+def collect_params(
+    *,
+    data_file: Path,
+    snapshot_file: Path,
+    binaries_path: Path,
+    result_repository: Path,
+    result_release_version: Path,
+    result_github_secret: Path,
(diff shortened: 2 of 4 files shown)


happybhati (approved):

ach912 (approved):
lgtm

scripts/python/tasks/managed/collect_gh_params.py
```diff
@@ -0,0 +1,93 @@
...
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import tekton
+from file import load_json_dict
+
+
+def validate_input_files(
```
johnbieren, line 17: You can resolve this without any changes but I think this function should be in the file.py helper. I have seen it in multiple python task scripts

johnbieren (approved):

davidmogar (approved):

pyproject.toml
```diff
@@ -45,6 +45,7 @@ pythonpath = [
 
 [dependency-groups]
 dev = [
+    "basedpyright>=1.39.9",
```
twaugh, line 48: Is this (and the resulting uv.lock change) necessary here?

twaugh (approved):

pyproject.toml
FilipNikolovski: It's just a dev dependency I added to get my LSP working, it's pretty harmless but I can create a separate PR for this if necessary.

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-08T08:22:04.841Z |

Approved

scripts/python/tasks/managed/collect_gh_params.py
FilipNikolovski: Sure, should I do a refactor as part of this PR across the scripts as well?

johnbieren: https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/file.py#L18 actually seems to already do it. But if you want to do it in a followup that is fine I suppose

FilipNikolovski: You mean to just directly use load_json_dict instead?

johnbieren: well, that function appears to load a json file and fail if it doesn't exist, which is the same that is being done here, right? I didn't know it existed. But either way my original point was that we have "fail if snapshot or data file doesn't exist" in a *ton* of tasks so should have it as a helper

FilipNikolovski: I think it makes sense to do a quick validation of the files, to keep it consistent with the bash scripts, I can make the refactor in a separate PR.

johnbieren: Yes it definitely should validate the files. I am saying I don't want us to maintain 60 versions of a quick validation of files. Fine with it in a separate PR

pyproject.toml
FilipNikolovski: @twaugh can we resolve this?

PR #868, opened 2026-07-06
feat(RELEASE-2481): add collect-signing-params standalone script

elenagerman:
## Describe your changes
Add standalone Python script for the collect-signing-params managed task. The script reads keyless signing configuration from a Kubernetes ConfigMap and writes parameters to Tekton result files.

## Relevant Jira
[RELEASE-2481](https://redhat.atlassian.net/browse/RELEASE-2481)

Assisted-by: Claude

[RELEASE-2481]: https://redhat.atlassian.net/browse/RELEASE-2481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/python/tasks/managed/collect_signing_params.py
+++ b/scripts/python/tasks/managed/collect_signing_params.py
@@ -1,0 +1,199 @@
+#!/usr/bin/env python3
+"""Collect Konflux signing configuration parameters from a Kubernetes ConfigMap.
+
+Retrieve keyless signing parameters from the cluster-config ConfigMap and write
+them to Tekton result files. If the ConfigMap is not found, output empty strings
+for all parameters except enableKeylessSigning which defaults to "false".
+"""
+
+from __future__ import annotations
+
+import argparse
+from pathlib import Path
+from typing import Any
+
+import kubectl
+import tekton
+from logger import logger
+
+RESULT_KEYS = (
+    "enableKeylessSigning",
+    "defaultOIDCIssuer",
+    "rekorExternalUrl",
+    "rekorUrl",
+    "fulcioExternalUrl",
+    "fulcioUrl",
+    "tufExternalUrl",
+    "tufUrl",
+    "buildIdentityRegexp",
+    "tektonChainsIdentity",
+)
+
+USAGE = """\
+usage: collect_signing_params.py --results-dir DIR [--config-map-name NAME] \
+[--config-map-namespace NAMESPACE]
+
+Collect signing parameters from a Kubernetes ConfigMap and write them to
+Tekton result files.
+
+required arguments:
+  --results-dir DIR           Directory to write Tekton result files
+
+optional arguments:
+  --config-map-name NAME      ConfigMap name (default: cluster-config)
+  --config-map-namespace NS   ConfigMap namespace (default: konflux-info)
+"""
+
+
+def get_empty_signing_params() -> dict[str, str]:
+    """Return signing parameters with empty values, enableKeylessSigning as 'false'."""
+    params = {key: "" for key in RESULT_KEYS}
+    params["enableKeylessSigning"] = "false"
+    return params
+
+
+def _prefer_internal_or_external(
+    data: dict[str, Any], internal_key: str, external_key: str
+) -> str:
+    """Return internal URL if set, otherwise fall back to external URL.
+
+    Args:
+        data: The ConfigMap data dictionary.
+        internal_key: The key for the internal URL.
+        external_key: The key for the external URL.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


midnightercz (dismissed):

theflockers (dismissed):

scripts/python/tasks/managed/collect_signing_params.py
```diff
@@ -0,0 +1,188 @@
...
+        result_path.write_text(value, encoding="utf-8")
+        logger.info("Wrote result %s = %s", key, value if value else "(empty)")
+
+
+def collect_signing_params(
+    config_map_name: str,
+    config_map_namespace: str,
+    results_dir: Path,
+    *,
+    get_configmap: Any = None,
```
johnbieren: nit - I don't think this needs to be a variable, we should be able to just use kubectl.get_configmap, but it is fine

scripts/python/tasks/managed/test_collect_signing_params.py
johnbieren, line 1: Maybe it exists and I am missing it, but I think it would be nice to have a test ensuring that (for example) rekorExternalUrl is used for rekorUrl when it exists but rekorInternalUrl does not

scripts/python/tasks/managed/collect_signing_params.py
elenagerman: This pattern is explicitly recommended in the repo's AGENTS.md:

"Inject dependencies as callable kwargs with real defaults, replaced only in tests."

scripts/python/tasks/managed/test_collect_signing_params.py
elenagerman: absolutely! good idea! thank you!
added

scripts/python/tasks/managed/collect_signing_params.py
johnbieren: @mmalina @FilipNikolovski can you chime in here? IIRC Martin has asked me not to do this (and it made sense to me) and Fitz is the author of the AGENTS.md file

scripts/python/tasks/managed/test_collect_signing_params.py
```diff
@@ -0,0 +1,444 @@
...
+        }
+    )
+    params = collect_signing_params.extract_signing_params_from_configmap(configmap)
+
+    assert params["rekorUrl"] == "https://rekor-internal.example.com"
+    assert params["fulcioUrl"] == "https://fulcio-internal.example.com"
+    assert params["tufUrl"] == "https://tuf-internal.example.com"
+
+
+def test_extract_signing_params_external_only_no_fallback() -> None:
```
johnbieren: I think this shows the script does not work as the bash does. Looking at https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/collect-signing-params/collect-signing-params.yaml#L143 if external is provided and internal is not, external should be used as results.rekorUrl (same for the fulcio and tuf) - the results should not be empty

scripts/python/tasks/managed/collect_signing_params.py
FilipNikolovski: Since we can monkey patch the functions from the modules in the tests, we can directly use `kubectl.get_configmap` I don't think there's a need to add it as a variable. I can update the AGENTS.md if everyone agrees.

mmalina: yeah, I don't think we need to use callables either. I didn't realize it was in AGENTS.md. I would vote for removing it from there.

scripts/python/tasks/managed/test_collect_signing_params.py
elenagerman: Good catch! You're right, my bad, the fallback logic was missing. 
fixed
thank you @johnbieren !!

scripts/python/tasks/managed/collect_signing_params.py
elenagerman: Makes sense; monkey patching is simpler. 
modified
Thank you all!

johnbieren (dismissed):

johnbieren (dismissed):

ach912 (dismissed):
lgtm

twaugh (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-20T12:00:31.897Z |

Approved

ach912 (approved):

PR #869, opened 2026-07-06
feat(RELEASE-2456): add standalone add-fbc-contribution script

elenagerman:
## Describe your changes
Add a standalone Python script for the add-fbc-contribution managed task logic so it can run from the utils image instead of an inline catalog script.

## Relevant Jira
[RELEASE-2456](https://redhat.atlassian.net/browse/RELEASE-2456)

Assisted-by: Claude

[RELEASE-2456]: https://redhat.atlassian.net/browse/RELEASE-2456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,961 @@
...
+    if result.returncode != 0 or not result.stdout.strip():
+        return {}
+
+    try:
+        return json.loads(result.stdout)
+    except json.JSONDecodeError:
+        return {}
+
+
+def decompress_json_build_info(compressed: str) -> dict[str, Any]:
```
seanconroy2021: decompress_json_build_info duplicates iib.decompress_build_info https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/iib.py#L152

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,961 @@
...
+    )
+    parser.add_argument(
+        "--verbose",
+        action="store_true",
+        help="Enable debug logging",
+    )
+    return parser
+
+
+def load_json_file(path: Path) -> dict[str, Any]:
```
seanconroy2021: load_json_file duplicates file.load_json_dict https://github.com/konflux-ci/release-service-utils/blob/1cf299607fd3de44a55adce98e5f4595fe6e26d9/scripts/python/helpers/file.py#L18

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,961 @@
...
+    return pipeline_timeout, task_timeout
+
+
+def execute_batch(
+    batch_num: int,
+    from_index: str,
+    group: OCPGroup,
+    config: AddFBCContributionConfig,
+    data: dict[str, Any],
+    timestamp_format: str,
```
seanconroy2021: timestamp_format is passed to execute_batch but never used inside it?

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,961 @@
...
+        "Creating %d batch(es) for %d components in OCP group %s",
+        num_batches,
+        num_components,
+        group.ocp_version,
+    )
+
+    successful_batches: list[int] = []
+    failed_batches: list[int] = []
+    latest_iib_index_image = ""
+    current_from_index = group.from_index
```
seanconroy2021: current_from_index = group.from_index is never read it's overwritten by get_current_from_index() right ?

happybhati:
Will this affect the refactoring of the task 
blocked for a long time https://github.com/konflux-ci/release-service-catalog/pull/2020

elenagerman:
> Will this affect the factoring of the task blocked for a long time [konflux-ci/release-service-catalog#2020](https://github.com/konflux-ci/release-service-catalog/pull/2020)



> Will this affect the factoring of the task blocked for a long time [konflux-ci/release-service-catalog#2020](https://github.com/konflux-ci/release-service-catalog/pull/2020)

Good question! This PR converts the add-fbc-contribution logic to Python, a different approach from #2020's bash step-splitting. 
But:
It depends on what the plan was. To finish the bash refactoring, and based on this, convert it into 3 python scripts, or just 1 python script.
 PR #2132 (multi-OCP support) still needs to be merged first, regardless, since both approaches require those changes.
IMO, the current PR should be frozen until a decision is made

@davidmogar @johnbieren @theflockers WDYT?

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: good catch!! thank you!!
removed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: good catch! thank you!
fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: sure; thank you! fixed

elenagerman:
let's hold off on merging this until https://github.com/konflux-ci/release-service-catalog/pull/2132 is merged.

johnbieren:
> > Will this affect the factoring of the task blocked for a long time [konflux-ci/release-service-catalog#2020](https://github.com/konflux-ci/release-service-catalog/pull/2020)
> 
> > Will this affect the factoring of the task blocked for a long time [konflux-ci/release-service-catalog#2020](https://github.com/konflux-ci/release-service-catalog/pull/2020)
> 
> Good question! This PR converts the add-fbc-contribution logic to Python, a different approach from #2020's bash step-splitting. But: It depends on what the plan was. To finish the bash refactoring, and based on this, convert it into 3 python scripts, or just 1 python script.  PR #2132 (multi-OCP support) still needs to be merged first, regardless, since both approaches require those changes. IMO, the current PR should be frozen until a decision is made
> 
> @davidmogar @johnbieren @theflockers WDYT?

I agree that 2132 should be merged first, especially since it comes from someone outside the team, and then its changes should be rebased into this and 2020. As far as 3 python scripts or one, if possible I think we should have the task call just one python script. I think the organization of it with python imports will be way cleaner than is achievable with one bash step. It does not matter much to me whether 2020 merges first and then this one rebases with any logic changes it brings or the other way around, as long as we end up with python when done

happybhati:
@elenagerman You can now continue with this as the blocker PR is merged now. Thanks for your patience!
https://github.com/konflux-ci/release-service-catalog/pull/2020

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,836 @@
...
+                data,
+            )
+        except BatchFailedError as e:
+            LOGGER.error(
+                "Batch %d failed after all retries for OCP %s: %s",
+                batch_num + 1,
+                group.ocp_version,
+                e,
+            )
+            return False
```
happybhati: Wondering if the retry behavior here might diverge from the current bash task?

In bash we try all batches first, collect failures, then retry the failed ones later (so other batches can still proceed, and IIB flakiness / in-progress reuse has a chance). Here it looks like we retry each batch immediately with exponential backoff, and if it still fails we stop the OCP group.

Might that change how multi-batch OCP groups behave under partial failures? Also the module docstring still says failed batches are retried at the end.

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,836 @@
...
+        completion_time = str(int(dt.timestamp()))
+    except (ValueError, OSError):
+        LOGGER.error("Failed to parse completion_time: %s", completion_time_raw)
+        return
+
+    if len(completion_time) != 10 or not completion_time.isdigit():
+        LOGGER.error(
+            "Invalid completion_time format (expected 10 digits): %s", completion_time
+        )
+        return
```
happybhati: Hmm, wondering what happens if `completion_time` is missing/invalid here?

In bash, `execute_batch` returns failure in that case so the batch is retried. Here the batch was already marked successful, and we just `return` without appending components. Could that leave us with a "successful" run but incomplete `internal-requests-results.json`?

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,836 @@
...
+
+        first_component = expanded_components[0]
+        from_index = first_component.get("updatedFromIndex", "")
+        target_index = first_component.get("targetIndex", "")
+
+        build_tags = list(global_build_tags)
+        if target_index:
+            target_tag = target_index.rsplit(":", 1)[-1]
+            if target_tag != target_index:
+                build_tags.append(target_tag)
```
happybhati: Wondering if we still want to fail when `targetIndex` has no tag?

In the current bash task we `exit 1` if there's no `:tag`. Here it looks like we just skip appending a build tag and continue. Might that let bad mapping config slip through?

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,836 @@
...
+
+    For hotfix or pre-ga releases, target_index already has the timestamp
+    (pattern: -<10_digit_epoch>). For staged releases, target_index is empty
+    so result is also empty. Otherwise, append the completion_time.
+    """
+    if not target_index:
+        return ""
+    if len(target_index) >= 11 and target_index[-11] == "-" and target_index[-10:].isdigit():
+        return target_index
+    return f"{target_index}-{completion_time}"
```
happybhati, line 493: Wondering if this timestamp check might treat some hotfix/pre-GA tags differently than bash?

Bash used something like "ends with 10 digits" (`.*[0-9]{10}$`), while this requires an exact `-<10 digits>` suffix. So e.g. a 14-digit suffix like `...-20240306163911` would stay as-is in bash, but here it would get another `-<epoch>` appended. Are hotfix/pre-GA tags always 10-digit epoch in practice?

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,836 @@
...
+
+    for fragment in fragments:
+        component_result = {
+            "fbc_fragment": fragment,
+            "target_index": group.target_index,
+            "target_index_with_timestamp": target_index_with_timestamp,
+            "ocp_version": group.ocp_version,
+            "image_digests": index_image_digests,
+            "index_image": build_info.get("index_image", ""),
+            "index_image_resolved": build_info.get("index_image_resolved", ""),
```
happybhati: Just double-checking this key change — wondering if it was intentional?

Current bash reads `internal_index_image_copy_resolved`, but here we use `index_image_resolved`. Looking at IIB / `update_fbc_catalog`, `index_image_resolved` seems right (and probably a good fix), but wanted to confirm we won't empty out `index_image_resolved` for downstream `collect-index-images` / publish if the old key was somehow what production relied on.

elenagerman: The Python code using build_info.get("index_image_resolved", "") is correct (IMO) and is actually a bug fix. The bash script was reading from a nonexistent field, which would result in null values propagating to downstream tasks.

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: good catch!!
fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: sure, fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: good catch, @happybhati !
thank you!
fixed

scripts/python/tasks/managed/add_fbc_contribution.py
happybhati: Thanks — appreciate the fix.

scripts/python/tasks/managed/add_fbc_contribution.py
happybhati: Thanks Elena — looks good with the deferred retry restored.

scripts/python/tasks/managed/add_fbc_contribution.py
happybhati: Thanks — bash parity looks right now.

scripts/python/tasks/managed/add_fbc_contribution.py
happybhati: Thanks!

scripts/python/tasks/managed/add_fbc_contribution.py
happybhati: Thanks for clarifying — agree, using `index_image_resolved` is the right call.

happybhati (dismissed):

elenagerman:
@theflockers @yashvardhannanavati  PTAL

seanconroy2021 (dismissed):

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,906 @@
...
+    index_image: str = ""
+    results: dict[str, Any] = field(default_factory=dict)
+    error_message: str = ""
+
+
+def setup_argparser() -> argparse.ArgumentParser:
+    """Build and return the CLI argument parser."""
+    parser = argparse.ArgumentParser(
+        description="Add FBC contributions to index images via InternalRequests.",
+        prog=PROG,
```
theflockers: `PROG` is used only here. Maybe you can just write the value here instead of a variable. If the concern is the file name changing for any reason, the alternative would be using `os.path.basename(__file__)`, so it would be always up to date.

Not crucial change - feel free to reject it.

scripts/python/tasks/managed/add_fbc_contribution.py
theflockers: It should be `internal_index_image_copy_resolved`. The reason is, `index_image_resolved` holds the image that is not kept by IIB and changes when a new build is started in parallel. 
On the other hand, `internal_index_image_copy_resolved` holds the build image for this specific build. This is functional and working in production. Changing back to `index_image_resolved` will break `create-pyxis-image`.

theflockers: All IIB builds holds the following fields:
```
index_image
index_image_resolved
internal_index_image_copy
internal_index_image_copy_resolved
```

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,906 @@
...
+    if not batch_result.success or not batch_result.results:
+        return False
+
+    json_build_info_compressed = batch_result.results.get("jsonBuildInfo", "")
+    if not json_build_info_compressed:
+        LOGGER.error("Missing jsonBuildInfo in batch results")
+        return False
+
+    try:
+        build_info = iib.decompress_build_info(json_build_info_compressed)
```
theflockers: Is it necessary to decompress again as it was done in `execute_batch` just before `process_batch_results`?

theflockers: https://github.com/konflux-ci/release-service-utils/pull/869/changes#diff-4c1fe7888812e220edfe0bdeca0233b153bf83ace30df9c9c2749bb00777e141R593

elenagerman: fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: fixed

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,906 @@
...
+            success=False,
+            error_message=str(e),
+        )
+    except Exception as e:
+        LOGGER.error("Batch %d failed to create request: %s", batch_num + 1, e)
+        return BatchResult(
+            batch_num=batch_num,
+            success=False,
+            error_message=str(e),
+        )
```
theflockers, line 416: I am missing the error handling for the internal-request itself. Internal Requests exit success by default, even if something goes wrong, because Openshift can't save results on failed IRs, so to propagate the error, it needs to end successfully at the IR side and set the exit code and result message. Using that result message we can decide if the manage task should exit 0 or fail with the propagated message from the IR.

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: fixed

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: fixed

theflockers:
Once FBC tests passes I can approve - I am watching it.

elenagerman:
> Once FBC tests passes I can approve - I am watching it.

thanks a lot!

happybhati (approved):

scripts/python/tasks/managed/add_fbc_contribution.py
```diff
@@ -0,0 +1,921 @@
...
+
+    if args.request_results_file_result:
+        args.request_results_file_result.write_text(
+            str(
+                request_results_file.relative_to(config.data_dir.parent)
+                if request_results_file.is_relative_to(config.data_dir.parent)
+                else request_results_file
+            ),
+            encoding="utf-8",
+        )
```
yashvardhannanavati, line 899: Assisted-by: Claude

request_results_file is resolved absolute under config.data_dir (confirmed via file.resolve_path_under_base, which always returns an absolute, resolved path). Then the code tries to make it relative to data_dir.parent — one level above the data directory — instead of writing the absolute path bash always wrote. With the default dataDir=/var/workdir/release, this produces release/<plr-uid>/ir-<tr-uid>-result.json instead of /var/workdir/release/<plr-uid>/ir-<tr-uid>-result.json.

This changes the actual content of a Tekton result that downstream consumers read. The bash task even has a comment flagging this exact result as fragile: "Note: adding a new result as modifying the current one used breaks e2e for single component. to be handled in RELEASE-1640." — this value's format has known downstream sensitivity, so a silent format change here is a real risk. The existing test (test_main_writes_result_files) only asserts the file .exists(), never checks its content, so this wasn't caught.

yashvardhannanavati:
one comment flagged by Claude, otherwise, looks good to me. thank you!

scripts/python/tasks/managed/add_fbc_contribution.py
elenagerman: good catch! thanks a lot @yashvardhannanavati !
fixed; now writes str(request_results_file) directly (the absolute path); test was improved accordingly

theflockers (approved):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-31T08:05:35.495Z |

davidmogar (approved):

PR #875, opened 2026-07-08
fix(KONFLUX-12379): rework collect_index_images

theflockers:
Replace build_timestamp logic with direct reading of
target_index_with_timestamp from results. Create
separate components for floating and timestamped
target indexes instead of one component with multiple
tags. Skip duplicates when both values are equal
(hotfix/pre-ga releases). Filter empty values for
staged releases. Add type validation for target_index
fields to prevent silent data loss, while allowing
partial fields (target_index without timestamp).

Assisted-by: Claude

### DIFF
--- a/scripts/python/tasks/managed/collect_index_images.py
+++ b/scripts/python/tasks/managed/collect_index_images.py
@@ -4,7 +4,6 @@
 from __future__ import annotations
 
 import json
-import re
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


happybhati (dismissed):

midnightercz (approved):

happybhati (approved):
LGTM, Nice work!

theflockers:
Reminder for the reviewer that fbc e2e won't pass as the fix also needs this update to move forward.

johnbieren (approved):

ach912 (approved):
lgtm

twaugh (approved):

PR #879, opened 2026-07-09
feat(RELEASE-2488): convert embargo-check task logic to python

querti:
Add embargo_check.py merging the check-issues and check-cves
tekton steps into a single Python script. Extract shared
Jira helpers (server normalization, URL construction,
credentials, REST operations) into a new jira.py helper
module and refactor close_advisory_issues.py to use it.

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/embargo_check.py
+++ b/scripts/python/tasks/managed/embargo_check.py
@@ -1,0 +1,377 @@
+#!/usr/bin/env python3
+"""Check if issues or CVEs in releaseNotes are embargoed.
+
+Validates Jira/Bugzilla issues from ``releaseNotes.issues.fixed`` by querying
+their REST APIs, injects a ``public`` boolean into each issue entry, validates
+that Vulnerability-type Jira issues have their CVE listed in the content section,
+and delegates CVE embargo checking to an InternalRequest running
+``check-embargoed-cves``.
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import random
+import time
+from pathlib import Path
+from typing import Any
+
+import file
+import http_client
+import requests
+import tekton
+from internal_request import SPAWN_OVERHEAD_SECONDS, InternalRequestWaitError, create
+from internal_request.internal_request import PIPELINERUN_UID_LABEL
+from jira import (
+    SUPPORTED_JIRA_SERVER,
+    jira_issue_url,
+    normalize_issue_server,
+    read_jira_credentials,
+)
+from logger import logger
+from requests.auth import HTTPBasicAuth
+from subprocess_cmd import run_cmd
+
+PROG = "embargo_check.py"
+
+CVE_FIELD = "customfield_10667"
+
+MAX_JIRA_404_RETRIES = 3
+
+
+def _get_with_jira_404_retry(
+    session: requests.Session,
+    url: str,
+    auth: HTTPBasicAuth,
+) -> dict[str, Any]:
+    """GET a Jira URL, retrying on transient 404s (RELEASE-2386).
+
+    Jira Cloud occasionally returns spurious 404 responses for issues that
+    exist.  This wrapper retries those with exponential backoff and jitter,
+    then returns the parsed JSON object (matching the ``jira_get_json``
+    contract).  Rate-limit (429) and server-error retries are handled by the
+    session's urllib3 Retry adapter.
+    """
+    retries = 0
+    while True:
+        resp = session.get(url, auth=auth, timeout=60.0)
+        if resp.status_code == 404:
+            retries += 1
+            if retries < MAX_JIRA_404_RETRIES:
+                delay = 1.0 * (2 ** (retries - 1)) + random.randint(0, 2)
+                logger.warning(
+                    "Received 404 for %s (attempt %d/%d), retrying in %.1fs",
+                    url,
+                    retries,
+                    MAX_JIRA_404_RETRIES,
+                    delay,
+                )
+                time.sleep(delay)
+                continue
+        resp.raise_for_status()
+        data = resp.json()
+        if not isinstance(data, dict):
+            msg = f"expected JSON object from {url}"
+            raise ValueError(msg)
+        return data
+
+
+def _get_content_items(data: dict[str, Any]) -> list[dict[str, Any]] | None:
+    """Return the content items list from releaseNotes, or None if absent.
+
+    Checks ``images`` first, then ``artifacts``.  Returns the list even when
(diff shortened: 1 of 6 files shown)


scripts/python/tasks/managed/embargo_check.py
```diff
@@ -0,0 +1,412 @@
...
+    """GET a Jira URL, retrying on transient 404s (RELEASE-2386).
+
+    Jira Cloud occasionally returns spurious 404 responses for issues that
+    exist.  This wrapper retries those with exponential backoff and jitter,
+    then returns the parsed JSON object (matching the ``jira_get_json``
+    contract).  Rate-limit (429) and server-error retries are handled by the
+    session's urllib3 Retry adapter.
+    """
+    retries = 0
+    while True:
```
midnightercz, line 57: there's module retry.py with retry_with_exponential_backoff. Why not use it?

scripts/python/tasks/managed/embargo_check.py
```diff
@@ -0,0 +1,412 @@
...
+        str(request_timeout),
+        "--pipeline-timeout",
+        pipeline_timeout,
+        "--task-timeout",
+        task_timeout,
+        "-s",
+        "true",
+    ]
+
+    result = subprocess.run(
```
midnightercz: There's also helper for this

midnightercz (changes requested):

scripts/python/tasks/managed/embargo_check.py
querti: Changed to use a helper.

scripts/python/tasks/managed/embargo_check.py
querti: The retry in the original bash script uses jitter, so a custom retry has to be used to preserve the behavior.

midnightercz (dismissed):

scripts/python/tasks/managed/test_close_advisory_issues.py
johnbieren, line 1: I think this has more tests to remove. For example (I am sure there are others) the tests for `jira_get_json` are still here but that function was moved to `jira.py`

scripts/python/tasks/managed/embargo_check.py
```diff
@@ -0,0 +1,403 @@
...
+    for item in items:
+        fixed = item.get("cves", {}).get("fixed", {})
+        cves.update(fixed.keys())
+    return sorted(cves)
+
+
+def _check_issue_visibility(session: requests.Session, url: str) -> bool:
+    """Return True if the issue is accessible without authentication."""
+    try:
+        resp = session.get(url, auth=None, timeout=60.0)
```
johnbieren, line 111: should this be using _get_with_jira_404_retry?

scripts/python/tasks/managed/embargo_check.py
```diff
@@ -0,0 +1,403 @@
...
+        return []
+
+    cves_str = " ".join(cves)
+    logger.info("Checking the following CVEs: %s", cves_str)
+
+    pipeline_timeout = _format_timeout(request_timeout + 300)
+    task_timeout = _format_timeout(request_timeout)
+
+    cmd = [
+        "internal-request",
```
johnbieren: https://github.com/konflux-ci/release-service-utils/commit/db7ca5773c329ebe3cb0678be7024f815caacce7 just merged, maybe you can switch to using the python module?

scripts/python/tasks/managed/embargo_check.py
```diff
@@ -0,0 +1,403 @@
...
+    read_jira_credentials,
+)
+from logger import logger
+from requests.auth import HTTPBasicAuth
+
+PROG = "embargo_check.py"
+
+CVE_FIELD = "customfield_10667"
+
+PIPELINERUN_LABEL = "internal-services.appstudio.openshift.io/pipelinerun-uid"
```
johnbieren: There is a constant now for this https://github.com/konflux-ci/release-service-utils/commit/db7ca5773c329ebe3cb0678be7024f815caacce7

scripts/python/tasks/managed/test_close_advisory_issues.py
querti: Good point, removed redundant tests

scripts/python/tasks/managed/embargo_check.py
querti: Refactored to use the new module

scripts/python/tasks/managed/embargo_check.py
querti: Fixed.

scripts/python/tasks/managed/embargo_check.py
querti: The current behavior matches the original script, so I think it would be safer to keep it as is. WDYT?

johnbieren: That's totally valid. Maybe you can leave this as is but once it is merged (and catalog update merged) you can open a new PR (no jira tied to it) to change this to the one with retries? I can't think of any reason we wouldn't want retries on it. Does that work for you?

johnbieren (approved):

scripts/python/tasks/managed/embargo_check.py
querti: Yeah, that works.

ach912 (approved):
lgtm

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-23T07:19:58.491Z |

Approved

scripts/python/tasks/managed/embargo_check.py
querti: UPDATE: I investigated in more detail if it's appropriate to use _get_with_jira_404_retry for retry here. The 404 retry was implemented for a particular bug (https://redhat.atlassian.net/browse/RELEASE-2386) , and I'm not sure if it's appropriate to also add it for unauthenticated call. Outside of that, the call has normal retries handled by the requests session, they just don't cover 404. I think we should only implement 404 retry if we observe 404 being returned incorrectly here. Otherwise, the code would just churn retries on every non-public Jira.

WDYT?

twaugh (approved):

scripts/python/tasks/managed/embargo_check.py
johnbieren: normal retries is good enough for me, thanks for the investigation

PR #880, opened 2026-07-09
feat(RELDEV-312): sign-and-push-to-internal-oci scripts

semccart-rh:
These scripts will be called by the internal pipeline of the same name.

Inspired by the push-artifacts-to-cdn Python re-write.

This script is a part of the "2 Build 2 Release" design by pkhander@redhat.com, originally made for Red Hat Build of Podman Desktop.

Content will be pushed to a location yet to be finalized. My proposal:
quay.io/<QUAY_ORG>/[non]prod/<ORIGIN_TENANT>/<COMPONENT>:tag
where tag = <PLR_UID>-<COMMIT_SHA>-(windows|mac)

Co-Authored-By: Claude Code (Opus 4.6 1M Context)


semccart-rh:
@querti do you mind re-adding the ok-to-test? Thank you!

elenagerman:
Hey @semccart-rh !
All looks good overall. 
One observation: extract_oci_artifacts.py and extract_artifacts.py share ~50% of their code (4 nearly identical functions).

Not a blocker, but worth to consider to extract common pieces into a shared module to reduce maintenance burden.

semccart-rh:
Thank you for taking a look @elenagerman!
I have removed functions that were identical in those two files in extract_oci_artifacts, and imported them from extract_artifacts. Please let me know if there is anything else I need to do.

elenagerman:
Nice job! 
The only thing: please pay attention that you probably will fail the Linter check during the pre-merge test run verification, since it will be 7 signs more added (' (#880)' to the message automatically, and the limit is 72, so your commit messages have to be shorter (for now, max 66 signs)
something like this:
`
1: T1 Title exceeds max length (75>72): "......
`

and by fixing this, you are losing approvals

semccart-rh:
Good point, I shortened the title @elenagerman. Thanks!

elenagerman (dismissed):

querti:
@semccart-rh the Python is still failing

semccart-rh:
@querti my apologies. I fixed the docstring failures.

querti:
We try to have 100 percent line coverage for the new code. Could you please ensure 100% coverage?

semccart-rh:
@querti I see the e2e tests are failing due to api limits. Is the recommended approach to just re-run them until we do not hit that, or is there another way to address this?

johnbieren:
> @querti I see the e2e tests are failing due to api limits. Is the recommended approach to just re-run them until we do not hit that, or is there another way to address this?

We have jira(s) open about this, it has been biting us a lot lately. We mainly just retry until it works (for now) but I am not sure how long it takes for the api limit to recover

src/helpers/sign_mac/sign_mac.py
```diff
@@ -134,6 +137,141 @@ def _ssh_opts(key_path: str, known_hosts: str) -> list[str]:
...
+
+    env_vars = {
+        "QUAY_USER": quay_user,
+        "QUAY_PASS": quay_pass,
+        "QUAY_DEST_USER": dest_quay_user,
+        "QUAY_DEST_PASS": dest_quay_pass,
+        "CSC_KEY_PASSWORD": keychain_password,
+        "CSC_NAME": signing_identity,
+        "UNSIGNED_REF": f"{quay_url}/unsigned/{name}@{unsigned_digest}",
+        "SIGNED_REF": f"{dest_quay_url}/{origin}/{name}:{tag}",
```
semccart-rh, line 223: I'm thinking of two options here (please let me know if there are better options):

1. A new quay org is created for this pipeline, and these signed artifacts are pushed here for the product team to later pull from in their second build (The code is set up in this way now)
2. We reuse the product team's redhat-user-workloads repo `f"{dest_quay_url}/{origin}/{component}:{tag}` (dest_quay_url is passed by the [tekton task](https://github.com/konflux-ci/release-service-catalog/pull/2364/changes#diff-171edb40f963c3b201614aee681a9bb7d0e0929dd7bdd2bcae76b39ecfa78063R189)) to avoid maintaining another repo. The product team will already have the pull credentials for their second build

semccart-rh:
> User semccart-rh is not allowed to trigger CI via pull_request in this repo.

@johnbieren would you mind re-adding ok-to-test?

elenagerman (dismissed):

semccart-rh:
@johnbieren @elenagerman do you mind re-adding the test?
It looks like the e2e tests have failed, plus I had to rebase due to #907
Also please let me know if there is an easier way of getting through the e2e tests in lieu of these failures

johnbieren:
/ok-to-test
I think because it is your first PR in this repo it has to be approved every time. Also, before merging this (fine during review phase though) you'll have to squash the 15 commits down to 1

midnightercz:
@semccart-rh 
here's migration check list, it should be just enough to give this to an agent

- Migrate `scripts/python/tasks/managed|internal/<mod>.py` → `src/tasks/managed|internal/<mod>/<mod>.py`
- Migrate `scripts/python/tasks/managed|internal/test_<mod>.py` → `src/tasks/managed|internal/<mod>/tests/test_<mod>.py`
- Migrate `scripts/python/helpers/` → `src/helpers/`
- Create `__init__.py` in new directories with docstring from main module
- In `<mod>/__init__.py` import all public methods from `<mod>/<mod>.py`
- Merge if destination file exists
- Remove migrated source files
- Run `chmod +x` on `src/tasks/internal/<mod>/<mod>.py` and `src/tasks/managed/<mod>/<mod>.py`
- Verify all `src/` modules are correctly linked in Dockerfile
- Fix imports in `./src/helpers/*/*`: change same-directory imports to `release_service_utils.helpers.<module>`
- Fix imports in `./src/tasks/{internal,managed}/*/*` and `utils/*`: change helper imports to `release_service_utils.helpers.<module>`
- Run `uv run pytest`
- Run `uv run flake8 src/`
- For long `mock.patch` lines: create `TASK = "release_service_utils.tasks.{managed,internal}.<task>"` and use `f"{TASK}.*"`

semccart-rh:
Thank you @midnightercz. I had to fix the executable permissions & Dockerfile links. Would you mind re-adding ok-to-test?

src/tasks/internal/sign_and_push_to_internal_oci/sign_and_push_to_internal_oci.py
```diff
@@ -0,0 +1,106 @@
...
+            origin=args.origin,
+        )
+        sign_windows.run_custom_signing(
+            args.quay_url,
+            args.pipeline_run_uid,
+            signing_script=args.windows_signing_script,
+            dest_quay_url=args.dest_quay_url,
+            origin=args.origin,
+        )
+    except Exception as exc:
```
querti, line 140: This seems different than what push_artifacts_to_cdn does here

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
```diff
@@ -0,0 +1,212 @@
...
+def _get_wanted_filenames(component: dict) -> set[str]:
+    """Return the set of filenames to extract from layer titles."""
+    wanted: set[str] = set()
+    for entry in list(component.get("files") or []) + list(
+        (component.get("staged") or {}).get("files") or []
+    ):
+        source = entry.get("source", "")
+        if source:
+            wanted.add(Path(source).name)
+        filename = entry.get("filename", "")
```
querti: Possible issue found by an LLM, I don't have enough context to verify: 

> source is the file’s name inside the OCI artifact (the layer title). filename is the name we want later when publishing — they are often different, e.g. app-bundle.tar.gz vs product-1.0-darwin.tar.gz.
Adding filename to wanted means extract will require a layer that does not exist and fail (see the missing = wanted - found check below). Please match layers by source basename (and entitlementsArtifact if needed). Use filename only if you want to rename the file on disk after copying.
The tests currently always use source == filename, so this isn’t covered. test_get_wanted_filenames_from_filename_field even asserts both names are required.

src/helpers/sign_mac/sign_mac.py
```diff
@@ -134,6 +137,141 @@ def _ssh_opts(key_path: str, known_hosts: str) -> list[str]:
...
+
+    Returns an empty string if the component has no entitlements artifact configured
+    or the artifact file is not found locally.
+    """
+    artifact_name = _get_entitlements_artifact_name(component)
+    if not artifact_name:
+        return ""
+
+    artifact_path = component_dir / artifact_name
+    if not artifact_path.exists():
```
querti, line 174: Possible issue found by an LLM, I don't have enough context to verify:

> If the RPA sets entitlementsArtifact, that file is required for a correct Mac sign ([RELDEV-387](https://redhat.atlassian.net/browse/RELDEV-387) — Electron apps crash without those entitlements). Skipping with a warning means we still sign, just without ENTITLEMENTS_REF, and the pipeline looks successful.
Please raise if the artifact is configured but missing. Returning "" is only appropriate when no entitlements were requested.

src/helpers/sign_mac/sign_mac.py
```diff
@@ -134,6 +137,141 @@ def _ssh_opts(key_path: str, known_hosts: str) -> list[str]:
...
+        "CSC_KEY_PASSWORD": keychain_password,
+        "CSC_NAME": signing_identity,
+        "UNSIGNED_REF": f"{quay_url}/unsigned/{name}@{unsigned_digest}",
+        "SIGNED_REF": f"{dest_quay_url}/{origin}/{name}:{tag}",
+        "OUTPUT_DIGEST": digest_file,
+    }
+    if entitlements_ref:
+        env_vars["ENTITLEMENTS_REF"] = entitlements_ref
+    args_str = " ".join(shlex.quote(a) for a in signing_args)
+    stdin_script = "#!/bin/bash\nset -eu\n"
```
querti, line 229: Issue found by an LLM:

> This exports QUAY_PASS, QUAY_DEST_PASS, and CSC_KEY_PASSWORD, then execs the host script. The built-in Mac signing script uses set +x around those values so they never hit Tekton logs. The host wrapper (sign_mac_wrapper.sh) uses set -eux, so these exports will be traced.
Please wrap the export block in set +x / set -x (and unset the secrets after handing them off if the wrapper does not need them as env). Same concern for the Windows custom path, which sets $env:QUAY_PASS with no equivalent guard.

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
```diff
@@ -0,0 +1,212 @@
...
+        shutil.copy2(str(blob_path), str(dest_file))
+        logger.info("  Extracted layer '%s' (%s bytes)", title, blob_path.stat().st_size)
+        found.add(title)
+
+    missing = wanted - found
+    if missing:
+        raise RuntimeError(f"OCI artifact for '{name}' is missing layers: {sorted(missing)}")
+
+
+def process_component(component: dict) -> None:
```
querti, line 113: Suggested refactor by an LLM:

> This is almost the same as extract_artifacts.process_component / _extract_from_oras (skopeo copy via select-oci-auth, then match org.opencontainers.image.title and copy blobs). You already reused parse_args, setup_docker_config, and create_os_flag_files from that module; the pull + extract path is still a second copy.
The only new behavior I see is --override-os/--override-arch for darwin/arm64 manifests, plus pulling entitlementsArtifact. Could those go into extract_artifacts instead of a parallel module? Otherwise we have two extractors to keep in sync whenever auth, retries, or ORAS matching changes.

src/tasks/internal/sign_and_push_to_internal_oci/sign_and_push_to_internal_oci.py
semccart-rh: Fixed to be consistent with push_artifacts_to_cdn's style

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
semccart-rh: This makes sense, the code is updated

src/helpers/sign_mac/sign_mac.py
semccart-rh: Yes that is valid. I made the appropriate updates

src/helpers/sign_mac/sign_mac.py
semccart-rh: Good catch, I applied the suggestions

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
semccart-rh: Moved logic into dedicated function in extract_artifacts & called that from extract_oci_artifacts

semccart-rh:
Thank you for the review @querti, I have addressed the comments. Please let me know if there is anything else I can improve upon.
@johnbieren do you mind adding ok-to-test in the meantime?

querti (approved):

semccart-rh:
@johnbieren would you mind adding another ok-to-test?

johnbieren:
/ok-to-test
I wish I could just tell prow to keep trusting you :upside_down_face:

semccart-rh:
> /ok-to-test I wish I could just tell prow to keep trusting you 🙃

I think we would both benefit from that lol. Let me know if you prefer pings here or on Slack. Maybe pinging ic-konflux-release would be preferred.

It looks like the e2e tests are failing on gh api errors. Could I have another ok-to-test @johnbieren?

semccart-rh:
@johnbieren I think this is ready to be reviewed by you or @davidmogar  (based on CODEOWNERS). As a part of that process, could [this question](https://github.com/konflux-ci/release-service-utils/pull/880#discussion_r3872246290) about the artifact's destination be addressed?

semccart-rh:
@johnbieren I had to resolve merge conflicts, could I get another ok-to-test?

semccart-rh:
@davidmogar thank you! Do you mind retesting? The tests failed on component creation

semccart-rh:
@johnbieren could you add ok-to-test?

semccart-rh:
> @johnbieren I think this is ready to be reviewed by you or @davidmogar (based on CODEOWNERS). As a part of that process, could [this question](https://github.com/konflux-ci/release-service-utils/pull/880#discussion_r3872246290) about the artifact's destination be addressed?

Just bringing this back up @johnbieren @davidmogar. Thank you in advance

src/helpers/extract_artifacts/extract_artifacts.py
```diff
@@ -46,9 +46,44 @@
 logger = logging.getLogger(__name__)
 
 
-def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
+def skopeo_copy_to_dir(
```
johnbieren: Missing tests for this one. It is also a bit weird to have a skopeo copy defined here when we already have https://github.com/konflux-ci/release-service-utils/blob/main/src/helpers/skopeo/skopeo.py#L60, maybe that one can be expanded to fit the use case here

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
```diff
@@ -0,0 +1,185 @@
...
+    num_files = len(component.get("files") or [])
+    num_staged = len((component.get("staged") or {}).get("files") or [])
+    if num_files == 0 and num_staged == 0:
+        logger.info(
+            "Skipping component '%s' - no files defined (not a binary artifact component)",
+            name,
+        )
+        return
+
+    pullspec = component.get("containerImage")
```
johnbieren: If you just do `component["containerImage"]` you get python's exception process and don't need the next two lines

src/helpers/push_unsigned/push_unsigned.py
```diff
@@ -122,25 +122,10 @@ def _unpack_file_entries(
             shutil.move(str(archive_path), str(target_dir / archive_name))
         else:
             with tarfile.open(str(archive_path)) as tf:
-                _safe_extract_archive(tf, target_dir, archive_name)
+                oras_utils.safe_extract_archive(tf, target_dir, archive_name)
             archive_path.unlink()
 
 
-def _safe_extract_archive(tf: tarfile.TarFile, target_dir: Path, archive_name: str) -> None:
```
johnbieren, line 152: If this is being removed, it should have tests to remove from test_push_unsigned.py too

johnbieren (commented):
Only got about halfway through, will review the rest tomorrow

querti (approved):

elenagerman (dismissed):

Dockerfile
```diff
@@ -213,6 +213,7 @@ RUN ln -s /home/src/tasks/managed/cleanup_workspace/cleanup_workspace.py /home/s
     ln -s /home/src/tasks/internal/process_file_updates/process_file_updates.py /home/scripts/python/tasks/internal/process_file_updates.py && \
     ln -s /home/src/tasks/internal/pulp_push_disk_images/pulp_push_disk_images.py /home/scripts/python/tasks/internal/pulp_push_disk_images.py && \
     ln -s /home/src/tasks/internal/push_artifacts_to_cdn/push_artifacts_to_cdn.py /home/scripts/python/tasks/internal/push_artifacts_to_cdn.py && \
+    ln -s /home/src/tasks/internal/sign_and_push_to_internal_oci/sign_and_push_to_internal_oci.py /home/scripts/python/tasks/internal/sign_and_push_to_internal_oci.py && \
```
johnbieren: This shouldn't be needed. You can just call the script as a python module in the catalog task. These were created for backwards compatibility with already converted tasks

johnbieren:
> Only got about halfway through, will review the rest tomorrow

done now

Dockerfile
semccart-rh: Removed

src/helpers/extract_oci_artifacts/extract_oci_artifacts.py
semccart-rh: Changed in accordance with your suggestion

src/helpers/push_unsigned/push_unsigned.py
semccart-rh: It seems that this function is not being tested in test_push_unsigned

src/helpers/extract_artifacts/extract_artifacts.py
semccart-rh: Removed from extract_artifacts & extended skopeo.copy

semccart-rh:
@johnbieren thank you for the feedback. I have applied your suggestions. Could you please add ok-to-test?

semccart-rh:
@johnbieren I had to resolve a merge conflict, could I have another ok-to-test?

johnbieren (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-09-11T14:53:41.625Z |

src/helpers/sign_mac/sign_mac.py
semccart-rh: Decided to go with redhat-user-workloads

PR #881, opened 2026-07-09
feat(RELEASE-1988): expand process_file_updates yaml replacement coverage

elenagerman:
## Describe your changes
Add unit tests that exercise the real yq/awk/sed replacement path

## Relevant Jira

Assisted-by: Claude

### DIFF
--- a/scripts/python/tasks/internal/process_file_updates.py
+++ b/scripts/python/tasks/internal/process_file_updates.py
@@ -9,7 +9,7 @@ configuration repositories, creating an MR for human review.
   ``FILE_UPDATES_SECRET_MOUNT``): ``gitlab_host``, ``gitlab_access_token``,
   ``git_author_name``, ``git_author_email``.
 * Clones ``--repo`` at ``--ref``, rebases on ``--upstream-repo``, applies path
-  seeds and YAML replacements (``yq``/``sed``), then commits or reuses an MR.
+  seeds and YAML replacements (``yq`` for key lookup), then commits or reuses an MR.
 * Writes ``RESULT_FILE_UPDATES_*`` and internal-request name results.
 * After a valid run with result env vars, almost always exits ``0``; failures use
   ``RESULT_FILE_UPDATES_STATE``. Invalid YAML on a replacement target exits ``1``.
@@ -23,6 +23,7 @@ Catalog Tekton tests mock ``git`` via ``tests/mocks/git`` on ``PATH``.
 from __future__ import annotations
 
 import argparse
+import difflib
 import json
 import os
 import re
@@ -174,19 +175,14 @@ def list_konflux_open_mrs(upstream_repo: str, gitlab_client: gitlab.Gitlab) -> l
 
 def blank_lines_before_yaml(target_file: Path) -> int:
     """Lines before first non-comment YAML line (matches legacy bash/sed numbering)."""
-    proc = subprocess.run(
-        [
-            "awk",
-            r"/[[:alpha:]]+/{ if(! match($0, \"^#\")) { print NR-1; exit } }",
-            str(target_file),
-        ],
-        text=True,
-        capture_output=True,
-        check=False,
-    )
-    if proc.returncode != 0 or not proc.stdout.strip():
+    try:
+        with target_file.open(encoding="utf-8") as f:
+            for i, line in enumerate(f):
+                if re.search(r"[a-zA-Z]", line) and not line.lstrip().startswith("#"):
+                    return i
+    except OSError:
         return 0
-    return int(proc.stdout.strip().splitlines()[0])
+    return 0
 
 
 def parse_replacement_expression(replacement: str) -> tuple[str, str] | None:
@@ -204,7 +200,7 @@ def apply_replacement_block(
     replacement: str,
     temp_dir: Path,
 ) -> tuple[str | None, Path | None]:
-    """Apply one ``|search|replace|`` sed to a YAML block; write diff to *temp_dir*.
+    """Apply one ``|search|replace|`` to a YAML block; write diff to *temp_dir*.
 
     Returns ``(None, diff_path)`` on success or ``(error, diff_path)`` on failure.
     """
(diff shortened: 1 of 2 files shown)


theflockers (dismissed):
LGTM

scripts/python/tasks/internal/process_file_updates.py
midnightercz, line 176: I know it's not part of this MR, but what's the point of pythonizing bash script if you run shell commands which can be replaced by few python lines? This definitely doesn't help with reducing memory consumed by python scripts

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -125,6 +132,174 @@ def secret_mount(tmp_path: Path) -> Path:
...
+def _require_gnu_sed() -> None:
+    """Skip YAML tooling tests that rely on GNU ``sed -i`` range syntax."""
+    if shutil.which("sed") is None:
+        pytest.skip("sed is not installed")
+    proc = subprocess.run(["sed", "--version"], text=True, capture_output=True, check=False)
+    if proc.returncode != 0 or "GNU sed" not in (proc.stdout + proc.stderr):
+        pytest.skip("GNU sed is required for replacement tooling tests")
+
+
+def _require_yaml_tooling() -> None:
```
midnightercz: require_tooling maybe if it checks other deps as well?

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -575,6 +748,22 @@ def test_blank_lines_before_yaml_returns_zero_on_failure(tmp_path: Path) -> None
         assert process_file_updates.blank_lines_before_yaml(target) == 0
 
 
+@pytest.mark.parametrize(
+    ("content", "expected"),
```
midnightercz: expected_result or expected_removed_lines maybe?

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -1419,3 +1608,436 @@ def test_main_parse_args_non_int_exit_code() -> None:
...
     ):
         assert process_file_updates.main(["process_file_updates.py"]) == 1
+
+
+def test_apply_replacements_integration_leading_comments_and_document_start(
+    tmp_path: Path,
+) -> None:
+    """Replacements honor leading comments and ``---`` via real ``yq``/``sed``."""
+    _require_yaml_tooling()
+    repo = tmp_path / "repo"
```
midnightercz: these 4-5 lines more or less repeats in several tests. It could be great if you can move them to setup function or fixture. Something like write_repo_content(path, content)

midnightercz (changes requested):

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: great idea! thank you!!
done

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: sure; renamed

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: sure; renamed

scripts/python/tasks/internal/process_file_updates.py
elenagerman: all replaced with the pure Python

happybhati (dismissed):

scripts/python/tasks/internal/process_file_updates.py
johnbieren: +1 - and thanks for fixing it Elena. I have been trying to say the same wherever I notice it when reviewing the python scripts. I think we should call subprocess.run as little as possible. I am even converting the internal-request bash script as part of my `create-advisory` conversion

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -6,7 +6,7 @@
 - ``replacements``: ``test_run_file_updates_creates_mr``,
   ``test_apply_replacements_for_entry_applies_replacement``
 - ``seed``: ``test_seed_target_file_writes_and_stages``
-- ``combo``: Tekton only (integration smoke)
+- ``combo``: Tekton only (catalog E2E smoke)
```
johnbieren: I don't really know what this line means

scripts/python/tasks/internal/test_process_file_updates.py
johnbieren: I don't think this or line 4 is very valuable. It is a nice note when the test file is first added, but now that the catalog task was converted, it pretty much has no meaning IMO

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -1419,3 +1574,420 @@ def test_main_parse_args_non_int_exit_code() -> None:
...
+    assert "releaseImage: primary-rel:9.9" in updated
+    assert "indexImage: metrics:9.9" in updated
+    assert updated.count("indexImage:") == 3
+    assert len(updated.splitlines()) == len(large_yaml.splitlines())
+
+
+def test_apply_replacement_block_yaml_tooling_multiline_size_mismatch(
+    tmp_path: Path,
+) -> None:
+    """Real ``sed`` block validation rejects multiline scalar size drift."""
```
johnbieren: Wasn't `sed` usage removed?

scripts/python/tasks/internal/test_process_file_updates.py
```diff
@@ -1419,3 +1574,420 @@ def test_main_parse_args_non_int_exit_code() -> None:
...
+        value_size,
+        "|line one|line ONE|",
+        tmp_path,
+    )
+    assert err == "Text block size differs from the original"
+    assert diff_path is not None
+
+
+def test_apply_replacement_block_yaml_tooling_greedy_replace(tmp_path: Path) -> None:
+    """Real ``sed`` detects greedy replacements that touch multiple lines."""
```
johnbieren: Wasn't `sed` usage removed?

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: "combo" = a path entry with both seed and replacements; creates a new file, then edits it. Not unit-tested in utils because the interesting part is the git flow (clone → seed → replace → commit → push → MR), which requires real git operations. Ideally, that should be covered by the catalog integration test (by file replacement or a new test). But, finally,  I didn't get your approval to implement it. So it is not relevant for now

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: fixed

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: good catch!! thank you @johnbieren !
I forgot to change it
fixed

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: all removed

ach912 (approved):

davidmogar (approved):

scripts/python/tasks/internal/test_process_file_updates.py
johnbieren: I don't really understand why it requires real git operations. Can't you monkey patch things so like clone will create a directory?

johnbieren (approved):

happybhati (approved):

scripts/python/tasks/internal/test_process_file_updates.py
elenagerman: already done :)

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-19T09:05:53.562Z |

Approved

midnightercz (approved):

PR #886, opened 2026-07-13
feat(RELEASE-2703): add AI skills for task creation and helper usage

happybhati:
## Summary
Adds repo-specific AI skills for release-service-utils per [RELEASE-2703](https://redhat.atlassian.net/browse/RELEASE-2703).
- `writing-new-task-scripts` — how to create Python task scripts (managed vs internal, Tekton wiring, testing, dual-PR workflow)
- `using-helpers` — catalog of shared helpers in `scripts/python/helpers/`
Skills follow the agentskills.io spec, are symlinked to `.claude/skills/` and `.cursor/skills/`, and `AGENTS.md` documents them.

[RELEASE-2703]: https://redhat.atlassian.net/browse/RELEASE-2703?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -24,7 +24,9 @@ Container image (UBI9) with Python scripts, wrappers, and templates used by Tekt
 - Type hints on all function signatures.
 - Docstrings on every module and public function (triple-double-quote, imperative mood).
 - Black formatter, line length 95. Flake8 linter, line length 95, E203 ignored.
-- Named loggers: `LOGGER = logging.getLogger("module_name")`.
+- Named loggers: `from logger import logger` (shared helper) in task scripts;
+  `logging.getLogger("module_name")` only in standalone helper submodules
+  (e.g. the artifact-signing pipeline).
 - argparse for CLI arguments; env vars for Tekton-injected config.
 - Entry point: `if __name__ == "__main__": raise SystemExit(main())`.
 - Exception chaining: always use `raise ... from e`.
@@ -62,3 +64,13 @@ Container image (UBI9) with Python scripts, wrappers, and templates used by Tekt
 - Wrapper scripts call external tools (pubtools-pulp, pubtools-marketplacesvm, etc.) via subprocess or library APIs.
 - Internal task scripts (`scripts/python/tasks/internal/`) must catch all exceptions in `main()` and save errors to Tekton result files (the script itself must succeed).
 - Managed task scripts (`scripts/python/tasks/managed/`) must not catch exceptions in `main()` — let the script fail with traceback.
+
+## Skills
+
+AI skills are in `skills/`. Each skill has a `SKILL.md` following the [agentskills.io](https://agentskills.io) spec.
+Symlinked to `.claude/skills/` and `.cursor/skills/` for agent discovery.
+
+Available skills:
+
+- `writing-new-task-scripts` — how to create a new Python task script following repo conventions
+- `using-helpers` — catalog of shared helpers to avoid reinventing existing code
--- a/skills/writing-new-task-scripts/SKILL.md
+++ b/skills/writing-new-task-scripts/SKILL.md
@@ -1,0 +1,301 @@
+---
+name: writing-new-task-scripts
+description: >-
+  Use when creating a new Python task script in release-service-utils, converting
+  a Tekton task from release-service-catalog, adding scripts/python/tasks/ code,
+  wiring Tekton params/results, or writing co-located pytest tests for a task.
+---
+
+# Writing New Task Scripts
+
+Create Python task scripts under `scripts/python/tasks/` that Tekton tasks in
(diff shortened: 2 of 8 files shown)


skills/writing-new-task-scripts/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+## Structure: separate `run()` from `main()`
+
+Extract testable logic into `run()` or named functions so `main()` stays a thin
+Tekton-env-to-function wrapper.
+
+**Do not add callable kwargs to a function's signature just to make it
+mockable** (e.g. `kinit: Any = authentication.kinit_with_retry`). This was the
+old convention — still visible in `check_embargoed_cves.py`, and `AGENTS.md`
+still documents it today — but the team has moved away from it in review after
+review (PRs #868, #872, and others): reviewers now consistently ask authors to
```
elenagerman: typo?

```suggestion
still documents it today — but the team has moved away from it in review (PRs #868, #872, and others): reviewers now consistently ask authors to
```

AGENTS.md
elenagerman: still describes the older per-module style
probably need to be consistent with https://github.com/konflux-ci/release-service-utils/pull/886/changes#diff-0d0f7eed4374c3876f29f79549e0bd8a6bfbd850dfd78f479d2389f212c05063R75

skills/writing-new-task-scripts/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+- Fixture data: minimal JSON under `tmp_path`; copy patterns from `test_check_data_keys.py`
+- Mock HTTP/subprocess/auth calls with `@patch("<task_module>.<helper_module>.<function>")`
+  at the task module's namespace — do **not** add callable kwargs to production
+  code to make mocking easier (see "Structure" above)
+- Cover every type/branch a validation function accepts, not just one (e.g. if a
+  flag can be JSON `true` or the string `"true"`, test both — a real review
+  comment caught a bool-only test suite for a function that also handled strings)
+
+Do **not** add unit tests in the conversion Jira story unless the ticket explicitly
+asks — but when you do add them, follow the patterns above.
```
elenagerman: IMO, that’s easy for agents to over-apply and skip coverage. 
maybe we need to drop or rephrase it to: prefer co-located tests when adding non-trivial logic? WDYT?

skills/writing-new-task-scripts/SKILL.md
```diff
@@ -0,0 +1,299 @@
...
+Tekton-env-to-function wrapper.
+
+**Do not add callable kwargs to a function's signature just to make it
+mockable** (e.g. `kinit: Any = authentication.kinit_with_retry`). This was the
+old convention — still visible in `check_embargoed_cves.py`, and `AGENTS.md`
+still documents it today — but the team has moved away from it in review after
+review (PRs #868, #872, and others): reviewers now consistently ask authors to
+hardcode the real call and drop the injected parameter. Treat `AGENTS.md`'s
+line on this as stale; this skill reflects the newer, repeatedly-enforced
+convention.
```
elenagerman: maybe it is worth to update AGENTS.md so it mathes the skill instead?
if I not mistaken, agents often read AGENTS.md first. 
And leaving both creates guaranteed conflicting guidance.

happybhati: Agreed on the conflict, but I didn't want to duplicate work — Filip's already got #898 open removing that exact AGENTS.md line (plus cleaning up the is_file() pre-check pattern Johnny called out). Rewrote our skill to just state the current rule plainly without pointing at a specific PR, so it holds up regardless of when #898 lands. Thanks for pushing on this!

skills/writing-new-task-scripts/SKILL.md
happybhati: Fair point, that phrasing could've been used as an excuse to skip coverage. Reworded it to default to writing co-located tests, and only skip for genuinely mechanical ports or when a ticket explicitly defers testing. Thanks!

AGENTS.md
happybhati: You're right, that line was stale — I checked the codebase and 20 files use the shared from logger import logger helper, so I updated AGENTS.md to describe that instead of the old per-module logging.getLogger() pattern. Thanks for flagging it!

skills/writing-new-task-scripts/SKILL.md
happybhati: Good catch, thanks! Fixed — that sentence was clunky. Pushed in the latest commit.

elenagerman (dismissed):

elenagerman (approved):

PR #892, opened 2026-07-15
feat(RELEASE-2484): add collect-task-params python script

elenagerman:
## Describe your changes
Add standalone Python script for collect-task-params managed task.

## Relevant Jira
[RELEASE-2484](https://redhat.atlassian.net/browse/RELEASE-2484)

Assisted-by: Claude

[RELEASE-2484]: https://redhat.atlassian.net/browse/RELEASE-2484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/python/tasks/managed/collect_task_params.py
+++ b/scripts/python/tasks/managed/collect_task_params.py
@@ -1,0 +1,227 @@
+#!/usr/bin/env python3
+"""Extract values from a data JSON file based on specified keys and result indices.
+
+This script takes an array of resultIndex/key pairs and extracts the corresponding
+values from a JSON data file. Each extracted value is placed at the specified
+resultIndex in the output array for use by downstream Tekton tasks. Optional default
+values can be provided for keys that may not exist in the data file.
+"""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+import tekton
+from file import load_json_dict
+from logger import logger
+
+
+def parse_keys_to_extract(keys_json: str) -> list[dict[str, Any]]:
+    """Parse and validate the keysToExtract JSON string.
+
+    The input must be a non-empty JSON array of objects. Each object should have
+    'resultIndex' and 'key' fields, with an optional 'default' field.
+    """
+    parsed = json.loads(keys_json)
+
+    if not isinstance(parsed, list):
+        raise ValueError(
+            f"keysToExtract must be a valid JSON array. Received value: {keys_json}"
+        )
+
+    if len(parsed) == 0:
+        raise ValueError("keysToExtract array is empty")
+
+    return parsed
+
+
+def validate_key_spec(spec: dict[str, Any], index: int, key_count: int) -> tuple[int, str]:
+    """Validate a single key extraction specification.
+
+    Return the validated (result_index, key) tuple, or raise ValueError.
+    """
+    result_index = spec.get("resultIndex")
+    key = spec.get("key")
+
+    if result_index is None or key is None:
+        raise ValueError(
+            f"Invalid key extraction specification at index {index}: "
+            "missing resultIndex or key"
+        )
+
+    if not isinstance(result_index, int) or result_index < 0:
+        raise ValueError(
+            f"resultIndex at position {index} must be a non-negative integer, "
+            f"got: {result_index}"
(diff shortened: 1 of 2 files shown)


scripts/python/tasks/managed/collect_task_params.py
```diff
@@ -0,0 +1,232 @@
...
+
+    return result_index, str(key)
+
+
+def transform_key_for_jq(key: str) -> list[str]:
+    """Transform a jq-style key path into a list of path segments.
+
+    Handles keys like '.foo.bar', '.foo["bar"]', '.foo.bar-baz', etc.
+    The transformation mimics the bash script's sed transformation that quotes
+    path segments for jq, but here we parse into segments for Python traversal.
```
FilipNikolovski: The current docstring explains where the logic came from (bash sed) but not what it does or what callers should expect. Perhaps something along these lines is better?

```
Keys in collect-task-params use jq dot notation: a leading '.', dot-separated
object keys, bracket notation for array indices (``.arr[0]``), and quoted
bracket keys for special characters (``.foo["bar-baz"]``).

Returns a list of segments used to walk a parsed JSON dict/list.

Examples:
        ".foo.bar"            -> ["foo", "bar"]
        ".release-notes.summary" -> ["release-notes", "summary"]
        '.foo["bar"]'         -> ["foo", "bar"]
        ".arr[0]"             -> ["arr", "0"]
        "plainKey"            -> ["plainKey"] 
        "."                   -> [] 
```
wdyt?

I'd also suggest renaming the function to something like `parse_jq_key_path` or `split_key_path` to be clearer.

FilipNikolovski (dismissed):
Nice work, I only have minor feedback, otherwise LGTM.

scripts/python/tasks/managed/collect_task_params.py
elenagerman: Great catch, thanks! You're right, the original docstring was more about where the code came from rather than what it actually does.

I've updated it with your suggested text and renamed the function to parse_jq_key_path

FilipNikolovski (dismissed):

scripts/python/tasks/managed/collect_task_params.py
```diff
@@ -0,0 +1,235 @@
...
+import json
+from pathlib import Path
+from typing import Any
+
+import tekton
+from file import load_json_dict
+from logger import logger
+
+
+class KeyExtractionError(Exception):
```
johnbieren: Python already has a KeyError, why are we making our own version of it? What do we gain by maintaining this extra code?

scripts/python/tasks/managed/collect_task_params.py
```diff
@@ -0,0 +1,235 @@
...
+
+def parse_keys_to_extract(keys_json: str) -> list[dict[str, Any]]:
+    """Parse and validate the keysToExtract JSON string.
+
+    The input must be a non-empty JSON array of objects. Each object should have
+    'resultIndex' and 'key' fields, with an optional 'default' field.
+    """
+    try:
+        parsed = json.loads(keys_json)
+    except json.JSONDecodeError as e:
```
johnbieren: why? Doesn't it make sense to just call `json.loads` and let the function do its thing and raise its exception if it fails? I am pretty sure if you pass non json to `json.loads` the exception will say something about invalid json. It isn't like this hurts, but I don't understand why we want to maintain our own version of code that imported functions already do

elenagerman: good catch! thank you @johnbieren !
simplified

scripts/python/tasks/managed/collect_task_params.py
elenagerman: good point! thank you!
refactored

johnbieren (approved):

ach912 (approved):
lgtm

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-21T08:41:04.470Z |

Approved

PR #893, opened 2026-07-15
feat(RELEASE-2486): add create-github-release python script

elenagerman:
## Describe your changes
Add standalone Python script for create-github-release managed task.

## Relevant Jira
[RELEASE-2486](https://redhat.atlassian.net/browse/RELEASE-2486)

Assisted-by: Claude

[RELEASE-2486]: https://redhat.atlassian.net/browse/RELEASE-2486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


seanconroy2021 (dismissed):
LGTM

scripts/python/helpers/container_image.py
```diff
@@ -0,0 +1,77 @@
...
+import json
+import tarfile
+from pathlib import Path
+from typing import Any
+
+import file as file_helper
+from logger import logger
+
+
+def load_snapshot(snapshot_path: Path) -> dict[str, Any]:
```
johnbieren: I don't understand this. Why do we have a helper that simply calls an existing helper?

scripts/python/helpers/container_image.py
```diff
@@ -0,0 +1,77 @@
...
+    """Load and return the snapshot JSON from *snapshot_path*."""
+    return file_helper.load_json_dict(snapshot_path)
+
+
+def get_first_component_image(snapshot: dict[str, Any]) -> str:
```
johnbieren: Can't you just use `snapshot.first_component(snapshot)["container_image"]`?

scripts/python/helpers/container_image.py
```diff
@@ -0,0 +1,77 @@
...
+    first = components[0]
+    if not isinstance(first, dict):
+        raise TypeError("snapshot component[0] must be an object")
+    image_url = str(first.get("containerImage", "")).strip()
+    if not image_url:
+        raise ValueError("Unable to get image url from snapshot: containerImage is empty")
+    return image_url
+
+
+def extract_binaries_from_layers(
```
johnbieren: I think this would fit better in an existing binaries helper (extract_artifacts maybe?). I think this entire container_image.py helper can go away, using existing funcs for the first two and moving this one

scripts/python/tasks/managed/create_github_release.py
```diff
@@ -0,0 +1,269 @@
...
+import tempfile
+from pathlib import Path
+
+import container_image
+import skopeo
+import tekton
+from logger import logger
+from vcs import github
+
+PROG = "create_github_release.py"
```
johnbieren: unused

scripts/python/tasks/managed/create_github_release.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+def _run_gh_command(
```
johnbieren: Shouldn't this go in the github helper https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/vcs/github.py ?

scripts/python/tasks/managed/create_github_release.py
```diff
@@ -0,0 +1,269 @@
...
+    gh_token: str,
+) -> str | None:
+    """Check if a release with the given version tag exists.
+
+    Return the release URL if it exists, or None if it does not.
+    Raise RuntimeError on API errors to avoid incorrectly attempting creation.
+    """
+    tag = f"v{release_version}"
+    result = _run_gh_command(
+        ["gh", "api", f"repos/{owner_repo}/releases/tags/{tag}"],
```
johnbieren: Isn't this a different endpoint than the task currently uses? https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/create-github-release/create-github-release.yaml#L163

scripts/python/tasks/managed/create_github_release.py
```diff
@@ -0,0 +1,269 @@
...
+
+        extracted_dir = tmp_dir / image_binaries_path
+        if not extracted_dir.is_dir():
+            raise ValueError(
+                f"Image {image_url} does not contain the '{image_binaries_path}' directory"
+            )
+
+        copied_count = copy_binaries_to_temp(extracted_dir, binaries_tmp)
+        if copied_count == 0:
+            raise ValueError(f"No binary files found in '{image_binaries_path}' directory")
```
johnbieren: I don't see the existing task fail if no binary files are found

scripts/python/helpers/container_image.py
swickersh: I started to leave this comment yesterday.  When running through Cursor:
```
I'd actually suggest the opposite direction. Moving extract_binaries_from_layers into extract_artifacts.py would go the wrong way, for a few reasons:

extract_artifacts.py is a heavy, task-specific orchestrator, not a generic domain helper. Even though it lives under helpers/, it's really a full task script — it has its own argparse, main(), ThreadPoolExecutor, module-level env-var config (REDHAT_WORKLOADS_TOKEN_MOUNT, CONTENT_DIR), calls select-oci-auth/skopeo itself, and depends on disk_image_utils.
```

That was one of the earlier helpers and it was written more as a helper "script" than set of related helper functions to import.
Likely we need to go back and tweak it to the more established pattern we have here now.(in a totally separate PR)

johnbieren: tbh I'd be fine with that rework being done in this PR but otherwise sounds good to me

swickersh: Yep that works too

scripts/python/tasks/managed/create_github_release.py
elenagerman: good catch! thank you!
fixed

scripts/python/tasks/managed/create_github_release.py
elenagerman: My implementation uses the tag-specific endpoint instead.
This was an intentional change based on earlier feedback in this PR; QoDo if I remember correctly.
the list endpoint has pagination issues and can miss an existing release if it's not in the first page of results. 
The tag-specific endpoint directly checks for the exact tag and handles 404 properly.

Do you want me to revert to match the original behavior (list endpoint with filtering)?

scripts/python/tasks/managed/create_github_release.py
elenagerman: good point!
moved

scripts/python/tasks/managed/create_github_release.py
elenagerman: good catch! thank you!
removed

scripts/python/helpers/container_image.py
elenagerman: my bad, missed it..
thank you @johnbieren !
I completely removed container_image.py then and using existing helper instead

scripts/python/helpers/container_image.py
elenagerman: absolutely!
fixed, thank you!

scripts/python/helpers/container_image.py
elenagerman: good catch, thank you @johnbieren !
just a leftover after some code duplication removal.
fixed

scripts/python/tasks/managed/create_github_release.py
johnbieren: I think it is much better to copy the existing logic as is first. Converting to python is already a huge change. You can then make any improvements you/qodo thought of in that process as a followup. If things break, I want it to be obvious whether it was due to the python conversion or some other optimization

elenagerman: sure! reverted

johnbieren (dismissed):

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-03T06:59:40.560Z |

Approved

davidmogar (approved):

davidmogar (approved):

PR #896, opened 2026-07-16
feat(KONFLUX-12379): convert index image tasks to python

theflockers:
Convert update-index-image and inspect-target-index
from inline bash to standalone Python scripts with
unit tests in release-service-utils.

Assisted-by: Claude
Signed-Off-By: Leandro Mendes

### DIFF
--- a/scripts/python/tasks/internal/inspect_target_index.py
+++ b/scripts/python/tasks/internal/inspect_target_index.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+"""Inspect a built FBC target index image using skopeo."""
+
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any
+
+from logger import logger
+from subprocess_cmd import run_cmd
+
+import tekton
+
+
+def read_credential(path: Path) -> str:
+    """Read a registry credential from a mounted secret file.
+
+    Args:
+        path: Path to the credential file (user:password format).
+
+    Returns:
+        The credential string.
+
+    """
+    return path.read_text(encoding="utf-8").strip()
+
+
+def inspect_image(source_index: str, credential: str) -> dict[str, Any]:
+    """Inspect a container image with skopeo and return its sha and per-arch digests.
+
+    Args:
+        source_index: Image pullspec to inspect (e.g. quay.io/redhat/index:v4.13).
+        credential: Registry credential in user:password format.
+
+    Returns:
+        Dict with "sha" (overall image digest) and "digests" (list of per-arch digests).
+
+    """
+    result = run_cmd(
+        [
+            "skopeo",
+            "inspect",
+            "--retry-times",
+            "3",
+            "--creds",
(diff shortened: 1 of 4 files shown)


scripts/python/tasks/internal/inspect_target_index.py
```diff
@@ -0,0 +1,91 @@
...
+from subprocess_cmd import run_cmd
+
+import tekton
+
+
+def read_credential(path: Path) -> str:
+    """Read a registry credential from a mounted secret file.
+
+    Args:
+        path: Path to the credential file (user:password format).
```
swickersh: Registry credentials can leak into Tekton logs on skopeo failure (inspect_target_index.py)

--creds user:pass is in the run_cmd argv. On failure, CalledProcessError stringifies the full command, and main() logs it:

Should we do an auth file like this: https://github.com/konflux-ci/release-service-utils/blob/main/scripts/python/helpers/extract_artifacts.py#L205

scripts/python/tasks/managed/update_index_image.py
```diff
@@ -0,0 +1,280 @@
...
+    """
+    updated = copy.deepcopy(snapshot)
+    components = updated.get("components", [])
+
+    for i, component in enumerate(components):
+        repos = component.get("repositories", [])
+        if not repos:
+            continue
+
+        tag = repos[0].get("tags", [""])[0]
```
swickersh: My cursor suggested this. 
Empty tags: [] raises IndexError (update_index_image.py)

`tag = repos[0].get("tags", [""])[0]`

Default only applies when tags is missing. Empty list → crash. Bash/jq returns null and skips. Safer:

```
tags = repos[0].get("tags") or []
tag = tags[0] if tags else ""
```

scripts/python/tasks/internal/inspect_target_index.py
theflockers: fixed according to `extract_artifacts.py`

scripts/python/tasks/managed/update_index_image.py
theflockers: fixed.

swickersh (approved):

seanconroy2021 (approved):
LGTM

scripts/python/tasks/internal/inspect_target_index.py
johnbieren, line 1: I don't see an internal catalog task by the same name - what one is this replacing?

scripts/python/tasks/managed/update_index_image.py
johnbieren, line 1: What managed task is this for? Is the managed one going to be renamed? I don't know that using the same name is strictly required, but it would be nice to at least mention it in the header

theflockers: They were to replace the ones (shell ones) added for KONFLUX-12379, but they ultimately got removed. I am closing this PR.

PR #900, opened 2026-07-17
fix: deduplicate architectures in get-image-architectures

Amanyd:
When an image index contains both gzip and zstd:chunked manifests per architecture (dual-compression), return only the non-zstd entry per platform. If all entries for a platform are zstd, keep the first one.

Ref: https://github.com/konflux-ci/architecture/blob/main/ADR/0070-dual-compression-for-container-builds.md

### DIFF
--- a/utils/get-image-architectures
+++ b/utils/get-image-architectures
@@ -86,5 +86,23 @@ elif [ $(jq -r '.mediaType' <<< $RAW_OUTPUT) == "application/vnd.docker.distribu
 else
     # Multi arch
     manifests=$(jq '.manifests[] += {multiarch: true}' <<< "$RAW_OUTPUT")
-    jq -cr '.manifests[]' <<< "$manifests"
+    # Deduplicate: when an image index contains both gzip and zstd:chunked
+    # manifests per architecture (dual-compression), keep only the non-zstd
+    # entry per (architecture, os). If all entries for a platform are zstd,
+    # keep the first one so the error surfaces downstream rather than
+    # silently returning nothing.
+    # Ref: ADR 0070 (konflux-ci/architecture)
+    jq -cr '.manifests[]' <<< "$manifests" | jq -cs '
+        group_by(.platform.architecture + "/" + .platform.os)
+        | map(
+            if (.[0].platform.architecture == null or .[0].platform.os == null or
+                .[0].platform.architecture == "unknown" or .[0].platform.os == "unknown")
+            then .[]
+            else
+                (map(select(.annotations["io.github.containers.compression.zstd"] == null)) | .[0])
+                // .[0]
+            end
+        )
+        | .[]
+    '
 fi


eskultety:
@happybhati @swickersh hey guys, do you think we could get this reviewed any time soon. I wouldn't ping this normally but since this is work for Google Summer of Code other work depends on we'd like to make progress. Many thanks in advance!
Reference to other origin release PR that led to this: https://github.com/konflux-ci/release-service-catalog/pull/2349

happybhati:
Nice, saw the coordination with #2349 — makes sense to centralize here. One thing to check before merging: if an index ever has multiple manifests with missing/unknown platform info (e.g. buildx attestation-manifest entries), this filter's `group_by` key would collapse them together and silently drop all but one. Worth checking if this can be guarded (e.g. skip grouping when architecture/os is null or "unknown") since this now covers more callers than the original 6 in the ADR.

swickersh (dismissed):
This lgtm
I'd like to see test coverage for it but I'm seeing that we don't have tests set up for any code in this directory - so that would be unreasonable to request that of you IMO.

Amanyd:
@swickersh   Yes sure, I'll add a guard to skip grouping when architecture or os is null or "unknown", those entries will pass through unchanged, and only entries with valid platform info get deduplicated.

swickersh (approved):

johnbieren (approved):

davidmogar (approved):

eskultety:
I see we've received a few ACK, can someone please approve the final CI job and merge? That would be amazing, thanks a lot!

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-03T06:58:39.325Z |

Approved

eskultety:
@Amanyd we may need a rebase here.

Amanyd:
@johnbieren @eskultety rebased onto the latest main, could you /ok-to-test again?

johnbieren:
We use a merge queue, so rebasing actually does nothing unless you have a conflict
/ok-to-test

eskultety:
> We use a merge queue, so rebasing actually does nothing unless you have a conflict /ok-to-test

@johnbieren so why was it removed from the merge queue then?

johnbieren:
> > We use a merge queue, so rebasing actually does nothing unless you have a conflict /ok-to-test
> 
> @johnbieren so why was it removed from the merge queue then?

Because it did not pass status checks in queue. Typically that happens when an e2e test fails in the merge queue, but here it looks like it timed out. My guess would be that the plr was kueued and we hit a timeout waiting

Amanyd:
Test coverage follow-up:  #954 pytest cases with a mocked skopeo, runs in CI

PR #901, opened 2026-07-17
feat(RELEASE-2500): convert marketplacesvm-push-disk-images to python

swickersh:
Move the managed task bash into a standalone script with pytest
coverage so the catalog can call it from the utils image. Also port
memory-throttle.sh's cgroup usage check to a reusable Python helper
and call it before each component's oras pull, so the converted
task still pauses new work under memory pressure like the other
bash tasks that use memory-throttle.sh.

Assisted-by: Cursor AI

### DIFF
--- a/scripts/python/helpers/test_memory_throttle.py
+++ b/scripts/python/helpers/test_memory_throttle.py
@@ -1,0 +1,172 @@
+"""Tests for memory_throttle."""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import memory_throttle as mt
+import pytest
+
+
+def _write(path: Path, value: str) -> None:
+    path.write_text(value, encoding="utf-8")
+
+
+# --- read_cgroup_memory ---
+
+
+def test_read_cgroup_memory_v2(tmp_path: Path) -> None:
+    """Cgroups v2 files are preferred and parsed as (current, max)."""
+    v2_current = tmp_path / "memory.current"
+    v2_max = tmp_path / "memory.max"
+    _write(v2_current, "1000\n")
+    _write(v2_max, "4000\n")
+
+    result = mt.read_cgroup_memory(
+        v2_current=v2_current,
+        v2_max=v2_max,
+        v1_current=tmp_path / "missing-v1-current",
+        v1_max=tmp_path / "missing-v1-max",
+    )
+    assert result == (1000, 4000)
+
+
+def test_read_cgroup_memory_v2_unlimited_falls_back_to_v1(tmp_path: Path) -> None:
+    """A v2 max of 'max' (unlimited) is ignored in favor of v1 values."""
+    v2_current = tmp_path / "memory.current"
+    v2_max = tmp_path / "memory.max"
+    v1_current = tmp_path / "usage_in_bytes"
+    v1_max = tmp_path / "limit_in_bytes"
+    _write(v2_current, "1000")
+    _write(v2_max, "max")
+    _write(v1_current, "2000")
+    _write(v1_max, "8000")
+
+    result = mt.read_cgroup_memory(
+        v2_current=v2_current, v2_max=v2_max, v1_current=v1_current, v1_max=v1_max
+    )
+    assert result == (2000, 8000)
+
+
+def test_read_cgroup_memory_unavailable(tmp_path: Path) -> None:
+    """None is returned when no cgroup interface is readable."""
+    missing = tmp_path / "missing"
+    result = mt.read_cgroup_memory(
+        v2_current=missing, v2_max=missing, v1_current=missing, v1_max=missing
+    )
+    assert result is None
+
+
+def test_read_cgroup_memory_non_numeric_content(tmp_path: Path) -> None:
+    """Non-numeric file content (corrupt/unexpected) is treated as unreadable."""
+    v2_current = tmp_path / "memory.current"
+    v2_max = tmp_path / "memory.max"
+    missing = tmp_path / "missing"
+    _write(v2_current, "not-a-number")
+    _write(v2_max, "4000")
+
+    result = mt.read_cgroup_memory(
+        v2_current=v2_current, v2_max=v2_max, v1_current=missing, v1_max=missing
+    )
+    assert result is None
+
+
+# --- get_memory_usage_percent / format_bytes / get_memory_stats ---
+
+
+def test_get_memory_usage_percent() -> None:
+    """Usage percent is computed as an integer floor division."""
+    assert mt.get_memory_usage_percent(read_memory=lambda: (500, 1000)) == 50
+    assert mt.get_memory_usage_percent(read_memory=lambda: (333, 1000)) == 33
+
+
+def test_get_memory_usage_percent_unavailable() -> None:
+    """None is returned when the reader has no data."""
+    assert mt.get_memory_usage_percent(read_memory=lambda: None) is None
+
+
+@pytest.mark.parametrize(
+    ("num_bytes", "expected"),
+    [
+        (512, "512B"),
+        (2545, "2Ki"),
+        (5 * 1024 * 1024 + 777, "5Mi"),
+        (3 * 1024 * 1024 * 1024 + 123456, "3Gi"),
+    ],
+)
+def test_format_bytes(num_bytes: int, expected: str) -> None:
+    """Byte counts are formatted with the largest whole unit, truncated."""
+    assert mt.format_bytes(num_bytes) == expected
+
+
+def test_get_memory_stats() -> None:
+    """Stats string reports used/limit and the usage percentage."""
+    assert mt.get_memory_stats(
+        read_memory=lambda: (512 * 1024 * 1024, 1024 * 1024 * 1024)
+    ) == ("512Mi/1Gi (50%)")
+
+
+def test_get_memory_stats_unavailable() -> None:
+    """'unavailable' is reported when the reader has no data."""
+    assert mt.get_memory_stats(read_memory=lambda: None) == "unavailable"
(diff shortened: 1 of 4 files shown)


scripts/python/helpers/test_memory_throttle.py
```diff
@@ -0,0 +1,158 @@
...
+    assert mt.get_memory_usage_percent(read_memory=lambda: None) is None
+
+
+@pytest.mark.parametrize(
+    ("num_bytes", "expected"),
+    [
+        (512, "512B"),
+        (2048, "2Ki"),
+        (5 * 1024 * 1024, "5Mi"),
+        (3 * 1024 * 1024 * 1024, "3Gi"),
```
mmalina: Isn't it a bit weird to have all the examples be exactly multiplies of 1024? :) The normal case will be that it's not like that - so something like 2545 or whatever.

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,440 @@
...
+    run_cmd: RunCmd = subprocess_cmd.run_cmd,
+    wait_for_memory: Callable[..., None] = memory_throttle.wait_for_memory,
+) -> int:
+    """Execute the full marketplacesvm disk-image push workflow."""
+    credential_files = validate_credentials(secrets_dir)
+    set_cloud_credentials(credential_files)
+
+    snapshot_file = data_dir / snapshot_path
+    snapshot = load_json_dict(snapshot_file)
+    starmap_file = write_starmap_file(snapshot, snapshot_file)
```
mmalina: A minor thing, but it's not really worth doing this if there are no components - so maybe only do this once you're sure there are components? i.e. move this down after the `raise ValueError`.

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
mmalina, line 1: It seems that if oras fails, the error log will be suppressed. From claude:
```
  Subprocess stdout/stderr is fully swallowed on failure, task-wide.

  subprocess_cmd.run_cmd (the helper everything in this PR uses to shell out) hardcodes
  stdout=subprocess.PIPE, and defaults stderr=subprocess.PIPE unless the caller passes
  stderr_path= to redirect it to a file. In marketplacesvm_push_disk_images.py, no caller ever
  passes stderr_path:

  - oras_utils.oras_pull(pullspec, download_dir) in prepare_component — no stderr_path, so both
  select-oci-auth and oras pull's stderr are captured into memory only.
  - validate_staged_structure() — run_cmd(["pushsource-ls", ...], check=True) — same.
  - run_marketplacesvm_push() — run_cmd([..., "marketplacesvm_push_wrapper", "--debug", ...],
  check=True) — same, and this is the one that matters most since --debug is presumably where the
  real diagnostic output lives.

  On failure, check=True raises subprocess.CalledProcessError, but CalledProcessError.__str__()
  does not include the captured stdout/stderr — it's just "Command '[...]' returned non-zero exit
  status 1." The actual output sits in exc.stdout/exc.stderr attributes that nothing in this file
  ever reads.

  Net effect: if oras pull fails (bad auth, network error, missing artifact) or the
  marketplacesvm_push_wrapper fails, the task will die with a content-free "exit status 1" and the
  operator gets zero information about why — a real regression from the bash version, where that
  output streamed straight into the pod log.

  Fix: thread a stderr_path through oras_pull, validate_staged_structure, and
  run_marketplacesvm_push (writing to a shared temp file)
```

scripts/python/helpers/test_memory_throttle.py
swickersh: Thanks - using more random numbers now.

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
swickersh: good suggestion - moved. thanks

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
swickersh: added a log_command_failure() helper. thanks

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,466 @@
...
+
+
+def validate_credentials(secrets_dir: Path) -> list[Path]:
+    """Validate marketplace credential JSON files and return their paths.
+
+    Each ``*.json`` file under *secrets_dir* must contain ``marketplace_account``
+    and ``auth`` keys. Raises ``RuntimeError`` when no files exist or validation
+    fails.
+    """
+    logger.info("DEBUG: Sanity validation of credentials")
```
querti: Should the DEBUG be there?

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,466 @@
...
+    starmap_file = write_starmap_file(snapshot, snapshot_file)
+
+    workdir.mkdir(parents=True, exist_ok=True)
+    with tempfile.TemporaryDirectory(dir=workdir) as base_dir_name:
+        base_dir = Path(base_dir_name)
+        disk_imgs_dir = base_dir / "starmap" / "CLOUD_IMAGES"
+        disk_imgs_dir.mkdir(parents=True, exist_ok=True)
+
+        prepare_components(
+            [c for c in components if isinstance(c, Mapping)],
```
querti: According to Claude, this is a behavior difference where bash version would raise for non JSON objects, but Python version silently drops them. Might be worth checking/verifying

querti (commented):
LGTM other than some minor comments. Also, could you please ensure 100% line coverage?

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
swickersh: removed the misleading debug prefix. It's just info.
thanks

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
swickersh: validating up front now. thanks

mmalina (dismissed):

ach912 (dismissed):
lgtm

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,468 @@
...
+        log_command_failure(exc)
+        raise
+
+
+def run_marketplacesvm_push(
+    base_dir: Path,
+    starmap_file: Path,
+    *,
+    pre_push: bool,
+    run_cmd: RunCmd = subprocess_cmd.run_cmd,
```
johnbieren: I think this should be hardcoded

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,468 @@
...
+
+def run(
+    *,
+    data_dir: Path,
+    snapshot_path: str,
+    pre_push: bool,
+    concurrent_limit: int,
+    secrets_dir: Path,
+    workdir: Path,
+    oras_pull: Callable[..., None] = oras_utils.oras_pull,
```
johnbieren: I think this should be hardcoded

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,468 @@
...
+    run_cmd: RunCmd = subprocess_cmd.run_cmd,
+    wait_for_memory: Callable[..., None] = memory_throttle.wait_for_memory,
```
johnbieren: I think this should be hardcoded

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
```diff
@@ -0,0 +1,468 @@
...
+def run(
+    *,
+    data_dir: Path,
+    snapshot_path: str,
+    pre_push: bool,
+    concurrent_limit: int,
+    secrets_dir: Path,
+    workdir: Path,
+    oras_pull: Callable[..., None] = oras_utils.oras_pull,
+    run_cmd: RunCmd = subprocess_cmd.run_cmd,
```
johnbieren: I think this should be hardcoded

johnbieren (commented):
Thanks for porting memory-throttle!

scripts/python/tasks/managed/marketplacesvm_push_disk_images.py
swickersh: Why? It looks like this dependency injection pattern is used just as much as "hardcoding" + monkeypatching in the repo (e.g. get_advisory_severity.py, pulp_push_disk_images.py's run_cmd vs. direct_sign_index_image.py, rh_direct_sign_image.py). I'm happy to change it, mostly just curious.

johnbieren: https://github.com/konflux-ci/release-service-utils/pull/868#discussion_r3587125932

mmalina: Sorry I missed this, I agree :)

swickersh: Thanks for the context. Fixed.

johnbieren (approved):

davidmogar (approved):

PR #912, opened 2026-07-21
fix(KFLUXSPRT-8470): make pulp task wait timeout configurable

bhills-rh:
wait_for_task() hardcoded a 120s timeout with no override, causing false failures when a Pulp task queues behind other work before executing. Add --pulp-task-timeout-seconds (default 300s, up from 120s) and log state transitions with elapsed time to make the next recurrence easier to diagnose.

Assisted-by: Claude

### DIFF
--- a/pubtools-pulp-wrapper/pulp_push_wrapper.py
+++ b/pubtools-pulp-wrapper/pulp_push_wrapper.py
@@ -48,6 +48,14 @@ EXODUS_ENV_VARS_STRICT = (
 EXODUS_ENV_VARS_OTHERS = ("EXODUS_GW_TIMEOUT",)
 
 
+def _positive_int(value: str) -> int:
+    """Parse *value* as an argparse type, rejecting non-positive integers."""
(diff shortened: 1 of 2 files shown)


swickersh:
Some discussion in slack around this still. The longer timeout might not be the best solution. Maybe we start with improved logging.

johnbieren (approved):

johnbieren:
It'd be nice if you could squash this down to a commit or two instead of five

bhills-rh:
> It'd be nice if you could squash this down to a commit or two instead of five

Done :heavy_check_mark:

swickersh (approved):

ach912 (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-27T10:02:54.082Z |

Approved

PR #914, opened 2026-07-22
feat(RELEASE-2487): convert create-pyxis-image task logic to Python

querti:
Rewrite the create-pyxis-image bash script as a Python module with
full unit test coverage. Components are processed in parallel via
ThreadPoolExecutor, grouped by digest to prevent Pyxis race conditions.

New shared helpers: memory_throttle (cgroup-aware memory pressure
throttling, copied from PR https://github.com/konflux-ci/release-service-utils/pull/901), oras_utils, image_ref, and
pyxis_api GraphQL URL resolution.

Refactors pyxis/create_container_image.py to expose a dataclass and
a reusable create_or_update() entry point, decoupling it from
argparse so the new task can call it as a library.

Assisted-by: Cursor

### DIFF
--- a/pyxis/create_container_image.py
+++ b/pyxis/create_container_image.py
@@ -44,6 +44,7 @@ so these images are available from both registries
 """
 
 import argparse
+from dataclasses import dataclass
 from urllib.parse import quote
 from datetime import datetime
 import json
@@ -60,6 +61,31 @@ MANIFEST_LIST_TYPES = [
     "application/vnd.oci.image.index.v1+json",
     "application/vnd.docker.distribution.manifest.list.v2+json",
 ]
+
+
+@dataclass(frozen=True)
+class ContainerImageArgs:
+    """Typed argument bundle for create_or_update and its helpers.
+
+    All fields mirror the CLI flags accepted by ``setup_argparser``.
+    The class is frozen so it can be safely shared across threads.
+    """
+
+    pyxis_url: str
+    certified: str
+    tags: str
+    is_latest: str
+    oras_manifest_fetch: str
+    name: str
+    media_type: str
+    digest: str
+    architecture_digest: str
+    architecture: str
+    rh_push: str = "false"
+    append_tags: str = "false"
+    dockerfile: str = ""
+    metadata: str = ""
+    verbose: bool = False
 
 
 def setup_argparser() -> Any:  # pragma: no cover
@@ -323,8 +349,11 @@ def repository_digest_values(args):
     return result
 
 
-def create_container_image(args, parsed_data: Dict[str, Any], tags: List[str]):
-    """Create a new containerImage entry in a pyxis instance."""
+def create_container_image(args, parsed_data: Dict[str, Any], tags: List[str]) -> str:
+    """Create a new containerImage entry in a pyxis instance.
+
+    :return: The ``_id`` of the newly created ContainerImage.
+    """
     LOGGER.info("Creating new container image")
 
     constructed_tags = construct_tags(tags)
@@ -361,6 +390,7 @@ def create_container_image(args, parsed_data: Dict[str, Any], tags: List[str]):
     # Make sure container metadata was successfully added to Pyxis
     if "_id" in rsp:
         emit_id(rsp["_id"])
+        return rsp["_id"]
     else:
         raise Exception("Image metadata was not successfully added to Pyxis.")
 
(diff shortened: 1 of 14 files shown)


scripts/python/tasks/managed/create_pyxis_image.py
```diff
@@ -0,0 +1,545 @@
...
+        "components": [results[i] for i in sorted(results) if results[i] is not None],
+    }
+
+    output_path = data_dir / pyxis_data_path
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    output_path.write_text(json.dumps(json_output, indent=2), encoding="utf-8")
+    logger.info("Pyxis data written to %s", output_path)
+
+
+def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
```
happybhati, line 504: I'm wondering if this could be done using `tekton.require_env("PARAM_*") / tekton.result_paths_from_env(...)` instead of `argparse` with CLI flags? Just noticed the other converted tasks (`request_advisory_creation.py`, `filter_already_released_advisory_rpms.py`, `validate_single_component.py`, etc.) all read their Tekton params/results from env vars via the shared tekton `helper` rather than CLI args. Might be worth aligning with that so the catalog-side task YAML can follow the same `command: ["python3", "-m", ...]` + env var wiring pattern as the rest. What do you think?

querti: It's a bit of a split. Some Python takss use args and some use env vars. I think it's largely determined by how the original task was structured and if it already used ENV vars. Since it's not consistent across the repo, would it be OK if we skipped this conversion for now? Then, once all tasks are converted, we could revisit this conversation and refactor to whichever way we prefer. WDYT?

pyxis/create_container_image.py
seanconroy2021: Should this not be raise SystemExit(main())

pyxis/test_create_container_image.py
```diff
@@ -836,6 +837,141 @@ def test_construct_tags__append_mode_empty_existing(mock_datetime):
     ]
 
 
+@patch("create_container_image.create_container_image")
+@patch("create_container_image.find_image")
+@patch("create_container_image.prepare_parsed_data")
+def test_create_or_update__new_image(mock_prepare, mock_find, mock_create):
```
seanconroy2021: All test functions in this file are missing -> None. This is pre-existing but should be an easy fix.

scripts/python/tasks/managed/create_pyxis_image.py
```diff
@@ -0,0 +1,545 @@
...
+            metadata_path=metadata_path if metadata else None,
+        )
+
+        component_json: dict[str, Any] = {
+            "containerImage": container_image,
+            "componentIndex": component_index,
+            "pyxisImages": [],
+        }
+
+        # Bug fix: the original bash script re-initialized COMPONENT_JSON inside
```
seanconroy2021, line 370: Do we really need the context of bash here? The explanation at the end would be enough right?

pyxis/create_container_image.py
querti: Fixed.

pyxis/test_create_container_image.py
querti: Fixed.

scripts/python/tasks/managed/create_pyxis_image.py
querti: Hmm, I don't know, the Python conversion changes the behavior of the original bash script in a way that I believe is a bugfix. But it is a change in behavior and I felt that it should be documented somehow. I feel that without the bash context the comment would lose its meaning.

WDYT?

seanconroy2021: It's more of NIT than anything up to you. Just seem if the issue is fixed; we dont need to mention it anymore :)

seanconroy2021 (dismissed):
LGTM - otherr than the open comments by Happy :)

src/tasks/managed/create_pyxis_image/create_pyxis_image.py
```diff
@@ -0,0 +1,545 @@
...
+        "components": [results[i] for i in sorted(results) if results[i] is not None],
+    }
+
+    output_path = data_dir / pyxis_data_path
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    output_path.write_text(json.dumps(json_output, indent=2), encoding="utf-8")
+    logger.info("Pyxis data written to %s", output_path)
+
+
+def _parse_args(argv: list[str] | None = None) -> argparse.Namespace:
```
happybhati: Sure, thanks for explaining.

happybhati (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-26T09:34:44.324Z |

Approved

davidmogar (approved):

davidmogar (approved):

PR #921, opened 2026-07-27
feat(RELEASE-2468): add apply_mapping.py script

swickersh:
Add a standalone python script that replicates the apply-mapping managed task's bash logic: merging mapping data into the snapshot, expanding tag templates (including incrementer and component-incrementer variables), extracting image metadata via skopeo, and converting registry URLs.

New helpers were added for reuse: json_merge (jq-style object multiply and deep array-union merge), image_architectures (wraps get-image-architectures), and date_format (wraps date). skopeo.py gained list_tags and extra inspect options, and image_ref.py gained registry URL conversion helpers.

The catalog task will be updated to call this script instead of its inline bash implementation.

Assisted-by: Cursor AI


scripts/python/helpers/date_format.py
```diff
@@ -0,0 +1,30 @@
...
+Python, this helper shells out to ``date`` so behavior matches exactly.
+"""
+
+from __future__ import annotations
+
+from subprocess_cmd import run_cmd_text
+
+
+def current_timestamp() -> str:
+    """Return the current local date/time formatted as ``%Y%m%d %T``."""
```
midnightercz, line 15: I understand you instruct agent to convert the existing bash to python, but it doesn't have to be literally. Can't you simply use dateime module?

scripts/python/helpers/date_format.py
```diff
@@ -0,0 +1,30 @@
...
+    return run_cmd_text(["date", "+%Y%m%d %T"]).strip()
+
+
+def format_date(date_string: str, date_format: str) -> str:
```
midnightercz, line 19: where is this used? I see it only in tests

scripts/python/helpers/image_architectures.py
```diff
@@ -0,0 +1,43 @@
...
+"""
+
+from __future__ import annotations
+
+import json
+
+from subprocess_cmd import run_cmd
+
+
+def get_image_architectures(
```
midnightercz, line 18: here again, I don't think we need helper module for one function. Is this created purelly by agent? Should we maybe adjust AGENTS.md?

scripts/python/helpers/image_architectures.py
```diff
@@ -0,0 +1,43 @@
...
+
+    Each returned dict has at least ``platform`` (with ``architecture`` and
+    ``os``) and ``digest`` keys; multi-arch images yield one dict per
+    manifest, while single-arch images and OCI artifacts yield exactly one.
+
+    Raises:
+        RuntimeError: if the underlying command fails.
+
+    """
+    cmd = ["get-image-architectures"]
```
midnightercz, line 33: Whole code bellow makes sense but I wonder if then isn't better to put it into that script

scripts/python/helpers/json_merge.py
```diff
@@ -0,0 +1,110 @@
+"""JSON merge helpers mirroring ``jq`` merge semantics used by release tasks."""
+
+from __future__ import annotations
```
midnightercz, line 10: There's jq binding for python. I wonder if it makes sense to "mirror" it like this. As here you have some functions that you use for code in this PR. Then in the future other might need to use more jq functionality and will add next methods here + relevant tests. And eventually this module can become so big that we reach the state where it will be just easier to use existing jq library binging in python.

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+    # don't support separate tags per arch, so we just use the first digest listed.
+    arch_infos = get_arch_fn(container_image)
+    arch = arch_infos[0]["platform"]["architecture"]
+    os_name = arch_infos[0]["platform"]["os"]
+    first_digest = arch_infos[0]["digest"]
+    image_with_digest = f"{container_image.rsplit('@', 1)[0]}@{first_digest}"
+
+    # Raw manifest inspection works for every artifact type and gives us annotations
+    # and the config media type, which tells us whether standard `inspect` will work.
+    raw_manifest = _inspect_json(inspect_fn, image_with_digest, no_tags=True, raw=True)
```
midnightercz: from here + whole if-else up to `oci_version_raw =` that could be separated method, ideally in some helper module. I think it should be like arch processing above. So one method which extract all info from the manifest and return some complex structure (or dict) and the set component details in here.

midnightercz (changes requested):

scripts/python/tasks/managed/test_apply_mapping.py
```diff
@@ -0,0 +1,1360 @@
...
+
+def _fake_inspect(raw_response: dict, standard_response: dict | None = None):
+    """Build an inspect_fn stub keyed on the ``raw`` flag."""
+
+    def _inspect(ref: str, **kwargs: Any) -> Any:
+        if kwargs.get("raw"):
+            return _completed(json.dumps(raw_response))
+        return _completed(json.dumps(standard_response or {}))
+
+    return _inspect
```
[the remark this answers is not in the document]
swickersh: done

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+    build_sha = _digest_sha(container_image)
+    passed_timestamp_format = component.get("timestampFormat") or default_timestamp_format
+    release_timestamp = format_date_fn(current_timestamp, passed_timestamp_format)
+
+    # The build-date label and Created values are not the same per architecture, but we
+    # don't support separate tags per arch, so we just use the first digest listed.
+    arch_infos = get_arch_fn(container_image)
+    arch = arch_infos[0]["platform"]["architecture"]
+    os_name = arch_infos[0]["platform"]["os"]
+    first_digest = arch_infos[0]["digest"]
```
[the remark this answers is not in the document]
swickersh: done

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+    """Merge ``original`` (Snapshot) and ``mapping`` components by name.
+
+    Only components present in both lists are kept, mirroring
+    ``group_by(.name) | select(length > 1)``. Matching components are merged
+    with :func:`json_merge.jq_multiply`, giving the mapping's values
+    precedence. Results are ordered by component name.
+    """
+    groups: dict[str | None, list[dict]] = {}
+    for component in list(original) + list(mapping):
+        groups.setdefault(component.get("name"), []).append(component)
```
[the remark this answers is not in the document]
swickersh: done

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+        default=False,
+        help="Append the resolved {{ timestamp }} value to each repository's tag list",
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv: list[str] | None = None) -> int:
+    """Parse arguments, resolve Tekton result paths, and run the mapping."""
+    setup_ca_cert()
+    args = _parse_args(argv)
```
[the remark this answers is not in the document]
swickersh: This is the exact same pattern used in filter_already_released_images.py (setup_ca_cert() called directly in main(), tested via mock.patch)

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+        url = repository.get("url") or ""
+        repo_tags = repository.get("tags") or []
+        tags_pre_substitution = json_merge.unique_sorted(
+            list(default_component_tags) + list(repo_tags)
+        )
+        tags = translate_tags(
+            tags_pre_substitution,
+            substitute_map,
+            labels,
+            url,
```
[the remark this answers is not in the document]
swickersh: checked this against the current production bash task (apply-mapping.yaml in release-service-catalog) line-by-line — it computes all_repos/calls translate_tags with the pre-conversion URL in bash too, in the identical order.

scripts/python/tasks/managed/apply_mapping.py
```diff
@@ -0,0 +1,593 @@
...
+
+        # This block is temporary to support both quay.io and registry.redhat.io.
+        # It should be removed once all repositories are migrated to registry.redhat.io.
+        if url.startswith(_QUAY_TEMP_PREFIXES):
+            url = image_ref.convert_to_registry(url)
+
+        if url.startswith(_REGISTRY_PREFIXES):
+            repository["rh-registry-repo"] = url
+            repository["registry-access-repo"] = image_ref.convert_to_registry_access(url)
+            repository["url"] = image_ref.convert_to_quay(url)
```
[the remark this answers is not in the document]
swickersh: Also exact parity with the bash task, which unconditionally sets this key too (confirmed against the catalog's own test-apply-mapping-flatpak-registries.yaml, which expects "" for flatpak repos). Downstream consumers already expect this field to exist as "" today; omitting the key would be an actual behavior change for an "Optional/Informational" severity item.

scripts/python/helpers/date_format.py
swickersh: in apply_mapping: https://github.com/konflux-ci/release-service-utils/pull/921/changes#diff-51e1260366e8dcb7a5c2b25d9a20dd4c046f0293b3a32aa5fbb43225ad972cc0R305

scripts/python/helpers/image_architectures.py
swickersh: Yes it's created by ai agent.
But IMO the use of a helper doesn't have much to do with it only containing one function. It has more to do with re-usability.  If other tasks (looks like 7 tasks in catalog) need to know the image architecture then this is good for DRY.

scripts/python/helpers/json_merge.py
swickersh: yea that's a good point. Converted to the python jq binding

scripts/python/helpers/date_format.py
swickersh: I'm leaning on Cursor a bit with this and it wrote a quick test to check. The suggestion was to replace the date subprocess call with Python's datetime.strftime(), since it would avoid shelling out. That works fine for the few custom formats actually in use (e.g. %Y%m%d), but the default format (no timestampFormat set, which ~94% of RPAs rely on) is %s — Unix epoch.

Turns out Python's strftime('%s') doesn't just lack native support for %s, it silently gives the wrong answer: it ignores tzinfo entirely and formats using local wall-clock fields, so a UTC-aware datetime comes out off by the local UTC offset:

```
dt = datetime.fromisoformat("2024-01-15T00:00:00Z")
dt.strftime('%s')     -> 1705294800  (wrong, off by local UTC offset)
int(dt.timestamp())   -> 1705276800  (correct, matches `date -d ... +%s`)
```
It's fixable, but only by special-casing %s separately from every other format directive, which reintroduces the "reimplement GNU date's rules" complexity we were trying to avoid by shelling out in the first place.

scripts/python/helpers/image_architectures.py
swickersh: get-image-architectures(bash) and its newline-delimited-JSON output is consumed directly by 8 other Tekton tasks across release-service-catalog via bash/jq. Changing its output format to make it more Python-friendly would be a breaking change with a wide blast radius, well outside this PR's scope. The thin Python wrapper mirrors the same pattern already used for skopeo.py (wrapping an external binary rather than reimplementing it).

scripts/python/tasks/managed/apply_mapping.py
swickersh: Good suggestion, created separate method but didn't extract to a helper module. I don't see other consumers of the helper yet.

scripts/python/helpers/date_format.py
midnightercz: Yeah, it makes sense I wasn't loaded as it was too big

scripts/python/helpers/image_architectures.py
midnightercz: How about to call it image utils then? Imaging you put every single function to new file, Then you'll have 300+ files, 300 directories for tests and so on. I think we should try to find balance on file size and number of files

midnightercz (dismissed):

scripts/python/helpers/image_architectures.py
swickersh: I don't see where I or anyone suggested putting every single function into a new file. Related functions would be in shared helper files. In this one instance, yes so far there's only one function in the file. 

I'll remove the helper though, and if/when other converted tasks need this same logic we can create the helper then if it makes sense.

midnightercz: Yes, that's what I'm little worried about as we don't write code manually but rather instruct agents to do that, this won't be removed ever because agents will basically write new files for new helper methods as they'll see this is helper just for image architectures.

swickersh: One tricky thing we face is that many of us are simultaneously converting bash scripts to python in `release-service-utils`and we don't know (without a lot of extra leg work) what functionality can be reused between many of the new python scripts.
I suggested at some point that **after** all tasks are converted, we go back through the utils repo and sweep for common functionality that can be combined in helper files and also we could look for duplicate helper files that basically do the same thing.

ach912 (dismissed):
lgtm

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-17T08:30:40.314Z |

Approved

PR #922, opened 2026-07-27
feat(RELEASE-2498): add get_ocp_version standalone script

happybhati:
This commit adds a python script to replicate the inline bash in the get-ocp-version managed task in the catalog repo. The task will be updated to use this python module instead.

-` get_ocp_version.py:` inspect FBC fragment images via skopeo, resolve multi-arch manifests via `get-image-architectures`, and validate all components share the same OCP version

- `test_get_ocp_version.py: `pytest cases covering single-arch, multi-arch (OCI index + Docker v2s2), version mismatch, and invalid format

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/get_ocp_version.py
+++ b/scripts/python/tasks/managed/get_ocp_version.py
@@ -1,0 +1,132 @@
+#!/usr/bin/env python3
+"""Validate that all FBC fragment components in a snapshot target one OCP version.
+
+For each snapshot component, inspect its FBC fragment image with
+`skopeo inspect` and read the `org.opencontainers.image.base.name`
+annotation to get the OCP version tag. Multi-arch images (OCI index or
+Docker manifest-list) are resolved to a single platform's manifest first via
+`get-image-architectures`. Every component must report the same `vX.Y`
+version, which is written to the Tekton result.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import re
+from pathlib import Path
+
+import skopeo
+import tekton
+from file import load_json_dict
+from logger import logger
+from subprocess_cmd import run_cmd_text
+
+PROG = "get_ocp_version.py"
+
+_VERSION_PATTERN = re.compile(r"^v[0-9]+\.[0-9]+$")
+_MULTI_ARCH_MEDIA_TYPES = {
+    "application/vnd.oci.image.index.v1+json",
+    "application/vnd.docker.distribution.manifest.list.v2+json",
+}
+
+
+def _base_name_tag(manifest: dict) -> str:
+    """Return the tag portion of a manifest's base-image annotation.
+
+    The `org.opencontainers.image.base.name` annotation has the form
+    `registry/path:vX.Y`; only the text after the last colon is kept,
+    mirroring the original `cut -d: -f2` behavior.
+    """
+    annotations = manifest.get("annotations") or {}
+    base_name = annotations.get("org.opencontainers.image.base.name") or ""
+    return base_name.rsplit(":", 1)[-1] if base_name else ""
+
+
+def resolve_ocp_version(fbc_fragment: str) -> str:
+    """Return the OCP version tag for *fbc_fragment*, resolving multi-arch images."""
+    manifest = json.loads(skopeo.inspect(fbc_fragment, raw=True, check=True).stdout)
+
+    if manifest.get("mediaType") in _MULTI_ARCH_MEDIA_TYPES:
+        logger.info("  Multiplatform image detected, extracting manifest")
+        arch_output = run_cmd_text(["get-image-architectures", fbc_fragment])
+        platforms = [json.loads(line) for line in arch_output.splitlines() if line.strip()]
+        manifest_image_sha = platforms[0]["digest"]
+        fbc_fragment = f"{fbc_fragment.rsplit('@', 1)[0]}@{manifest_image_sha}"
+        manifest = json.loads(skopeo.inspect(fbc_fragment, raw=True, check=True).stdout)
+
+    return _base_name_tag(manifest)
+
+
+def validate_ocp_versions(snapshot: dict) -> str:
+    """Validate every component targets the same, correctly formatted OCP version.
+
+    Return the unified `vX.Y` version string. Raise `ValueError` if a
+    component's version is malformed or a version mismatch is found.
+    """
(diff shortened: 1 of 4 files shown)


querti (dismissed):

seanconroy2021 (dismissed):
LGTM - Can you close out the Qodo comments :)

scripts/python/tasks/managed/get_ocp_version.py
```diff
@@ -0,0 +1,141 @@
...
+PROG = "get_ocp_version.py"
+
+_VERSION_PATTERN = re.compile(r"^v[0-9]+\.[0-9]+$")
+_MULTI_ARCH_MEDIA_TYPES = {
+    "application/vnd.oci.image.index.v1+json",
+    "application/vnd.docker.distribution.manifest.list.v2+json",
+}
+
+
+def _skopeo_inspect_raw(image_ref: str) -> dict:
```
johnbieren: Why do we need a function to just call an imported helper?

scripts/python/tasks/managed/get_ocp_version.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    components = snapshot.get("components") or []
+    logger.info("Found %d FBC components to validate", len(components))
+    if not components:
+        raise ValueError("No components found in snapshot")
+
+    validated_version = ""
+    for i, component in enumerate(components):
+        logger.info("Processing component %d...", i)
+        fbc_fragment = component.get("containerImage", "")
```
johnbieren: IIUC, `resolve_ocp_version(fbc_fragment)` will call `_skopeo_inspect_raw(fbc_fragment)` which will fail if `fbc_fragment` is `""` so why define a default which will end up just mess with the error message (something about skopeo inspecting empty string instead of simply component had no `containerImage`)?

happybhati: Good catch, fixed — now it checks for containerImage explicitly and raises a clear error (Component {i} has no containerImage) instead of letting an empty string reach skopeo. Added tests for the missing/empty cases too.

scripts/python/tasks/managed/get_ocp_version.py
happybhati: It's called twice here — once for the initial manifest, once after resolving the multi-arch digest — so I pulled it out to avoid duplicating the error-check + parse logic. `update_fbc_catalog.py` does the same thing with `inspect_image_created()` for the same reason. Can inline it if you'd still prefer that though.

scripts/python/tasks/managed/get_ocp_version.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    components = snapshot.get("components") or []
+    logger.info("Found %d FBC components to validate", len(components))
+    if not components:
+        raise ValueError("No components found in snapshot")
+
+    validated_version = ""
+    for i, component in enumerate(components):
+        logger.info("Processing component %d...", i)
+        fbc_fragment = component.get("containerImage")
```
johnbieren: this works, but `component["containerImage"]` lets you remove the next two lines too and use python normal exceptions

scripts/python/tasks/managed/get_ocp_version.py
johnbieren: That's fair, but I think it would be nice to do what Lubo did here https://github.com/konflux-ci/release-service-utils/pull/930/changes#diff-096731e0baf73cda3aec9a6f5b8e14913af53eb26811684ace329123311f6a7fR34 so we can reuse the checking of the return code, and then you don't really need this function at all

happybhati: Ah yes, thanks, makes sense.
Done — dropped the wrapper. Added `check=True` to `skopeo.inspect()` (same pattern Lubo added to `skopeo.copy()` in #930), so it raises `CalledProcessError` on failure directly and both call sites just do `json.loads(skopeo.inspect(...).stdout)` inline now.

scripts/python/tasks/managed/get_ocp_version.py
happybhati: Done — using component["containerImage"] now, missing key raises KeyError naturally.
Thanks :)

johnbieren (approved):
/retest

happybhati:
e2e failing with ongoing issue
```
⚠️ Warning: Could not get component PR from annotations: {"message":"waiting for spec.containerImage to be set by ImageRepository with annotation image-controller.appstudio.redhat.com/update-component-image"}
Waiting 20 seconds before retry...
```

happybhati:
All required tests are passing and ready to be merged.

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-16T07:44:49.661Z |

Approved

happybhati:
Reached secondary rate limit trying again

happybhati:
Follow-up for anyone finding this later: while wiring this script up in the catalog task ([release-service-catalog#2414](https://github.com/konflux-ci/release-service-catalog/pull/2414)), we found that `get-ocp-version` hasn't actually been invoked by the `fbc-release` pipeline since **2025-09-19**, in [`c52f9300`](https://github.com/konflux-ci/release-service-catalog/commit/c52f93003b9e9c28f44184f56681495b8bf261c3) (`fix(KONFLUX-10248): restore multi-OCP version support`). That change replaced the single-global-OCP-version check with per-component handling in `prepare-fbc-parameters`/`add-fbc-contribution`, and dropped `get-ocp-version` from the pipeline (its sibling `update-ocp-tag` was deleted outright at the time; `get-ocp-version`'s task directory was left behind orphaned).

Since there's no pipeline path back to this task, we're closing the catalog PR and [RELEASE-2498](https://redhat.atlassian.net/browse/RELEASE-2498) without wiring `get_ocp_version.py` up rather than merging dead code. Leaving this script here since this PR is already merged — happy to remove it in a follow-up cleanup if the team prefers.


[RELEASE-2498]: https://redhat.atlassian.net/browse/RELEASE-2498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #923, opened 2026-07-27
feat(RELEASE-2493): add filter_already_released_advisory_images script

happybhati:
This commit adds a python script to replicate the inline bash in the filter-already-released-advisory-images managed task in the catalog repo. The task will be updated to use this python module instead.

- `filter_already_released_advisory_images_managed.py`: transform snapshot components to arch-specific images, classify mapped repositories as stage/production, and submit them to the internal pipeline via `internal_request.create()`

- `test_filter_already_released_advisory_images_managed.py:` 42 pytest cases covering repo classification, skipFilter, arch-resolution failures, and InternalRequest success/failure paths

Note: the script is named with a `_managed` suffix (not just `filter_already_released_advisory_images.py`) because `scripts/python/tasks/internal/` already has an unrelated internal task script with that exact base name; since neither directory is a real Python package, pytest couldn't tell the two `test_filter_already_released_advisory_images.py` files apart without the rename.

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
+++ b/scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
@@ -1,0 +1,342 @@
+#!/usr/bin/env python3
+"""Filter advisory-published images from a snapshot before advisory creation.
+
+For each snapshot component, resolve architecture-specific image digests via
+`get-image-architectures`, classify the component's mapped repository URLs as
+pending (stage) or production advisories, and submit the resulting entries to
+the `filter-already-released-advisory-images` internal pipeline via an
+InternalRequest. Components the internal pipeline reports as still needing
+release are kept in the snapshot; fully-released snapshots are reduced to an
+empty component list and the run is marked skippable.
+"""
+
+from __future__ import annotations
+
+import base64
+import dataclasses
+import gzip
+import json
+import os
+import re
+import subprocess
+from pathlib import Path
+from typing import Any
+
+import internal_request
+import subprocess_cmd
+import tekton
+from file import load_json_dict
+from logger import logger
+
+PROG = "filter_already_released_advisory_images_managed.py"
+
+_PENDING_PATTERN = re.compile(r"quay\.io/redhat-pending/|quay\.io/rh-flatpaks-stage/")
+_PROD_PATTERN = re.compile(r"quay\.io/redhat-prod/|quay\.io/rh-flatpaks-prod/")
+_PIPELINE = "filter-already-released-advisory-images"
+_PIPELINERUN_UID_LABEL = "internal-services.appstudio.openshift.io/pipelinerun-uid"
+
+
+@dataclasses.dataclass(frozen=True)
+class ResultPaths:
+    """Tekton result file paths written by the filtering workflow."""
+
+    result: Path
+    skip_release: Path
+    environment: Path
+    latest_advisory_url: Path
+    latest_advisory_internal_url: Path
+
+
+@dataclasses.dataclass(frozen=True)
+class FilterConfig:
+    """Input configuration for the filtering workflow."""
+
+    snapshot_file: Path
+    rpa_file: Path
+    data_file: Path
+    results_file: Path
+    pipeline_run_uid: str
+    task_git_url: str
+    task_git_revision: str
+    synchronously: bool
+
+
+def _repo_url_category(url: str) -> str:
+    """Classify a mapped repository URL as "pending", "prod", or "orphan"."""
+    if _PENDING_PATTERN.search(url):
+        return "pending"
+    if _PROD_PATTERN.search(url):
+        return "prod"
+    return "orphan"
+
+
+def determine_environment(snapshot: dict[str, Any]) -> tuple[str, str]:
+    """Classify every mapped repository URL and pick the advisory environment.
+
+    Return ``(environment, advisory_secret_name)``. Raise `ValueError` when
+    repositories mix pending and production URLs, contain orphaned URLs, or
+    when no mapped repository is found at all.
+    """
+    found = {"pending": False, "prod": False, "orphan": False}
+    for component in snapshot.get("components", []):
+        for repo in component.get("repositories") or []:
+            url = repo.get("url") or ""
+            if not url:
+                continue
+            found[_repo_url_category(url)] = True
+
+    logger.info("Repository status:")
+    logger.info("- Pending repositories: %s", found["pending"])
+    logger.info("- Production repositories: %s", found["prod"])
+    logger.info("- Orphan repositories: %s", found["orphan"])
+
+    if found["pending"] and found["prod"]:
+        raise ValueError("cannot publish to both redhat-pending and redhat-prod repositories")
+    if not found["pending"] and not found["prod"]:
+        raise ValueError(
+            "you must publish to either redhat-pending or redhat-prod repositories"
+        )
+    if found["orphan"]:
+        raise ValueError(
+            "you must publish to either redhat-pending or redhat-prod repositories"
+        )
+
+    if found["pending"]:
+        return "stage", "create-advisory-stage-secret"
+    return "production", "create-advisory-prod-secret"
+
+
+def _get_image_architectures(image: str) -> list[dict[str, Any]]:
+    """Return parsed architecture/digest objects for *image*.
+
+    Raises `subprocess.CalledProcessError` when architecture resolution fails.
+    """
+    output = subprocess_cmd.run_cmd_text(["get-image-architectures", image])
+    return [json.loads(line) for line in output.splitlines() if line.strip()]
+
+
+def transform_component(component: dict[str, Any]) -> list[dict[str, Any]]:
+    """Expand one snapshot component into arch-specific transformed entries.
+
+    Return one minimal entry per (mapped repository, resolved architecture)
+    pair, containing only the fields the internal pipeline needs. Raises
+    `subprocess.CalledProcessError` when architecture resolution fails.
+    """
+    name = component.get("name", "")
+    image = component.get("containerImage", "")
+    digests = _get_image_architectures(image)
+
+    entries: list[dict[str, Any]] = []
+    for repo in component.get("repositories") or []:
+        repo_url = repo.get("rh-registry-repo", "")
+        tags = repo.get("tags")
+        for arch in digests:
+            entries.append(
+                {
+                    "name": name,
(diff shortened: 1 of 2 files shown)


scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,342 @@
...
+from pathlib import Path
+from typing import Any
+
+import internal_request
+import subprocess_cmd
+import tekton
+from file import load_json_dict
+from logger import logger
+
+PROG = "filter_already_released_advisory_images_managed.py"
```
midnightercz, line 31: unused variable, it looks like this is claude or other agent flaw as it also generated file like this for me where there's PROG = ... but it's actually not used.
And interestingly qodo doesn't mind

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,342 @@
...
+        params={
+            "transformedSnapshot": transformed_snapshot,
+            "origin": origin,
+            "advisory_secret_name": advisory_secret_name,
+            "internalRequestPipelineRunName": config.pipeline_run_uid,
+            "taskGitUrl": config.task_git_url,
+            "taskGitRevision": config.task_git_revision,
+        },
+        labels={_PIPELINERUN_UID_LABEL: config.pipeline_run_uid},
+        sync=config.synchronously,
```
midnightercz, line 213: This should be always true shouldn't it?

midnightercz (changes requested):

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,342 @@
...
+    )
```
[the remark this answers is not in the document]
mmalina: I would dismiss this. It is based on a line in AGENTS.md, but we should remove it.

scripts/python/tasks/managed/test_filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,692 @@
...
+        rpa_file=tmp_path / "rpa.json",
+        data_file=tmp_path / "data.json",
+        results_file=tmp_path / "results.json",
+        pipeline_run_uid="uid-1",
+        task_git_url="http://localhost",
+        task_git_revision="main",
+        synchronously=True,
+    )
+    defaults.update(overrides)
+    return task.FilterConfig(**defaults)
```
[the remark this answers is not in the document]
happybhati: Good catch, added type hints for both. Fixed.

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
happybhati: Yeah, dead code. Removed it, thanks for catching that.

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
happybhati: You're right that the code assumes it's true (no polling if it's not). This matches the bash task exactly though - same immediate read of results right after creating the IR, no wait loop either way. None of the real pipelines override the default, so in practice it's always true today. Kept it configurable to match the existing param 1:1, but happy to just hardcode sync=True here if you'd rather not carry that assumption forward, let me know.

midnightercz: Can you file maybe JIRA for that? To remove the param or change the code as now it allows to run the code in invalid way

mmalina: yeah, it never made sense to have that as parameter, because the implementation would always work just with one or the other, but not both. I would suggest to just remove the parameter as part of this work. but if you want to keep it 1:1, that's fine with me.

happybhati: Good call — removed the param entirely, internal_request.create()'s own sync=True default handles it now. Just pushed. I'll drop the synchronously param from the catalog task too once the new utils image is available (konflux-ci/release-service-catalog#2415).

happybhati:
I am wondering why all e2e skipped and passing; I was expecting advisory releated will be executed.

johnbieren:
> I am wondering why all e2e skipped and passing; I was expecting advisory releated will be executed.

The smart test selection looks for the changed python script to be either used in a task or imported in a python script used in a task. Because this is a brand new script, it isn't in any task in the development branch of release-service-catalog, so they all skip. Once the task is converted to use this script, future changes to it should trigger advisories e2e tests

happybhati:
> > I am wondering why all e2e skipped and passing; I was expecting advisory releated will be executed.
> 
> The smart test selection looks for the changed python script to be either used in a task or imported in a python script used in a task. Because this is a brand new script, it isn't in any task in the development branch of [release-service-catalog](https://issues.redhat.com/browse/RELEASE-service-catalog), so they all skip. Once the task is converted to use this script, future changes to it should trigger advisories e2e tests

Ah, okay, thanks for the clarification!

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,336 @@
...
+    if check_skip_filter(config.data_file):
+        logger.info(
+            "skipFilter is true in data.json, skipping filtering and passing "
+            "snapshot unchanged."
+        )
+        results.skip_release.write_text("false", encoding="utf-8")
+        results.result.write_text("Success", encoding="utf-8")
+        return
+
+    ir_results = submit_filter_request(
```
midnightercz: nitpick, technically you're also waiting for results not just submitting the request.

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,336 @@
...
+        entries,
+        origin=origin,
+        advisory_secret_name=advisory_secret_name,
+        config=config,
+    )
+
+    advisory_url = ir_results.get("advisory_url", "")
```
midnightercz: I saw this in other code as well when fetching data from IR. Why ".get"? Aren't we expecting result to have advisory_url? Shouldn't the script fail if advisory_url is empty?

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,336 @@
...
+        raise ValueError(
+            "you must publish to either redhat-pending or redhat-prod repositories"
+        )
+
+    if found["pending"]:
+        return "stage", "create-advisory-stage-secret"
+    return "production", "create-advisory-prod-secret"
+
+
+def _get_image_architectures(image: str) -> list[dict[str, Any]]:
```
midnightercz: little confusing name

scripts/python/tasks/managed/test_filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,680 @@
...
+        written_snapshot = json.loads(config.snapshot_file.read_text(encoding="utf-8"))
+        assert written_snapshot["components"] == []
+        results_file = json.loads(config.results_file.read_text(encoding="utf-8"))
+        assert results_file == {
+            "advisory": {"url": "http://advisory", "internal_url": "http://internal"}
+        }
+
+    def test_partial_components_released(self, tmp_path: Path) -> None:
+        """Some components still need release: filtered snapshot, skip_release=false."""
+        config = _config(tmp_path)
```
midnightercz: these two lines seem to be repeating over few tests, it could be worth of writing some setup method or fixture

midnightercz (approved):

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
happybhati: Good catch, thanks! Renamed it to `run_filter_request` since it does wait for and fetch the result too.

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
happybhati: Agreed, thanks! Renamed to `_resolve_image_architectures`.

scripts/python/tasks/managed/test_filter_already_released_advisory_images_managed.py
happybhati: Good idea, thanks! Added a `config`/`results` fixture so those lines aren't repeated in every test.

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
happybhati: Fair point, thanks! Switched to direct indexing there so it fails loudly instead of silently writing an empty string if the pipeline ever doesn't set it.

midnightercz (dismissed):

elenagerman:
this script is not executable

scripts/python/tasks/managed/filter_already_released_advisory_images_managed.py
```diff
@@ -0,0 +1,351 @@
...
+
+def check_skip_filter(data_file: Path) -> bool:
+    """Return True when data.json sets `skipFilter` to boolean or string true."""
+    if not data_file.is_file():
+        return False
+    value = load_json_dict(data_file).get("skipFilter")
+    return value is True or value == "true"
+
+
+def _fetch_ir_results(ir_name: str) -> dict[str, Any]:
```
elenagerman: Reimplements internal_request.fetch_results which already exists. Other scripts use the helper.

happybhati: Thanks @elenagerman — addressed both in `3cfa624`:

- Made the script executable (`chmod +x`).
- Removed `_fetch_ir_results()` and switched `run_filter_request()` to use the existing `internal_request.fetch_results()` helper instead, matching the pattern used in `push_disk_images.py` / `request_advisory_creation.py`. Updated the test mocks accordingly.

elenagerman (approved):

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-09-01T11:29:47.943Z |

Approved

PR #927, opened 2026-07-28
fix(RELEASE-2749): sign windows staging binaries with machine store

swickersh:
The e2e Windows signing cert now lives in the Local Machine store to survive host reboots and password rotations without an RDP session. Add /sm to signtool for the e2e-konflux account only, detected from the quay nonprod url; production's HSM-backed cert stays in the Current User store and is unaffected.

Assisted-by: Cursor AI

### DIFF
--- a/scripts/python/helpers/sign_windows.py
+++ b/scripts/python/helpers/sign_windows.py
@@ -45,6 +45,19 @@ CONTENT_DIR = Path(os.environ.get("CONTENT_DIR", "/shared/artifacts"))
 logger = logging.getLogger(__name__)
 
 
+def _is_staging_quay_url(quay_url: str) -> bool:
+    """Return True if quay_url targets the nonprod (staging) artifact path.
+
+    Staging releases route through the nonprod Quay path (set based on
+    release intention in release-service-catalog's
+    tasks/managed/push-artifacts-to-cdn/push-artifacts-to-cdn.yaml) and sign
+    with a certificate in the Local Machine store, so signtool needs the
+    /sm flag only for those. Production keeps using its HSM-backed cert in
+    the Current User store.
+    """
+    return "/nonprod" in quay_url
+
+
 def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     """Parse and return CLI arguments."""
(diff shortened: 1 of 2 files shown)


scripts/python/helpers/sign_windows.py
```diff
@@ -172,6 +187,7 @@ def run(quay_url: str, pipeline_run_uid: str) -> None:
             unsigned_digest=unsigned_digest,
             pipeline_run_uid=pipeline_run_uid,
             windows_temp_dir=windows_temp_dir,
+            use_machine_store=_is_staging_quay_url(quay_url),
```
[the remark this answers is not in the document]
swickersh: the nonprod destination is only used by the e2e account.

scripts/python/helpers/sign_windows.py
```diff
@@ -45,6 +45,18 @@
 logger = logging.getLogger(__name__)
 
 
+def _is_staging_quay_url(quay_url: str) -> bool:
+    """Return True if quay_url targets the nonprod (staging) artifact path.
+
+    Staging releases route through the nonprod Quay path (see
+    push-artifacts-to-cdn.yaml, which sets this based on release intention)
```
johnbieren: I know you're off on PTO now and this will likely be merged before then, but maybe as a followup, this says `see push-artifacts-to-cdn.yaml`. A pipeline definition? Task? Internal or managed? It would be nice to be a little more specific

johnbieren (dismissed):

scripts/python/helpers/sign_windows.py
swickersh: good point. Updated

johnbieren (approved):

querti (approved):

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-30T13:44:01.906Z |

Approved

johnbieren:
The title on this is longer than 72 chars once the merge queue appends to it. I'm not sure we are going to be able to merge it until Scott is back

PR #928, opened 2026-07-29
feat(RELEASE-2477): convert collect-data task to python

querti:
Add collect_data.py script and comprehensive unit tests.

Refactor get_resource module to expose get_resource() and
get_resource_dict() functions, making it importable from Python
instead of only usable as a CLI entrypoint. Also add a
ResourceFetchError exception for structured error handling.

The check-data-keys validation (previously a separate Tekton step)
is incorporated directly into the collect_data script

Assisted-by: Cursor

### DIFF
--- a/scripts/python/tasks/managed/collect_data.py
+++ b/scripts/python/tasks/managed/collect_data.py
@@ -1,0 +1,408 @@
+#!/usr/bin/env python3
+"""Collect release data from K8s resources and merge into data.json.
+
+Fetch Release, ReleasePlan, ReleasePlanAdmission, ReleaseServiceConfig, and
+Snapshot CRs; merge collector results with spec.data from those resources
+(priority: RPA > RP > Release > collectors); resolve pipeline ref metadata;
+validate disallowed data key sources; and write JSON files plus Tekton results.
+"""
+
+from __future__ import annotations
+
+import dataclasses
+import json
+import os
+from pathlib import Path
+from typing import Any
+
+import http_client
+from authentication import setup_ca_cert
+from file import resolve_path_under_base
+from get_resource import get_resource_dict
+from vcs.github import owner_repo_from_url
+import tekton
+from logger import logger
+
+_DISALLOWED_RELEASE_NOTE_KEYS: tuple[str, ...] = (
+    "releaseNotes.product_id",
+    "releaseNotes.product_name",
+    "releaseNotes.product_version",
+    "releaseNotes.product_stream",
+    "releaseNotes.cpe",
+    "releaseNotes.allow_custom_live_id",
+)
+
+_RESULT_ENV_VARS: dict[str, str] = {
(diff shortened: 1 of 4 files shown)


mmalina:
> The check-data-keys validation (previously a separate Tekton step)
> is incorporated directly into the collect_data script

Can you clarify this? AFAICS, check-data-keys is a complete task with its own python script. Or is this something else?

scripts/python/tasks/managed/collect_data.py
```diff
@@ -0,0 +1,408 @@
...
+}
+
+
+def _resolve_commit_sha(org: str, repo: str, revision: str) -> str:
+    """Query the GitHub API for the full commit SHA, falling back to 'unknown'."""
+    try:
+        api_url = f"https://api.github.com/repos/{org}/{repo}/commits/{revision}"
+        response_text = http_client.get_text(api_url, timeout=10)
+        return json.loads(response_text).get("sha") or "unknown"
+    except Exception:
```
mmalina, line 150: So we will completely swallow the exception and not log what the error was? Could we include the exception text with the warning perhaps?

querti:
> > The check-data-keys validation (previously a separate Tekton step)
> > is incorporated directly into the collect_data script
> 
> Can you clarify this? AFAICS, check-data-keys is a complete task with its own python script. Or is this something else?

I meant this step: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/collect-data/collect-data.yaml#L323 , sorry if it was unclear.

scripts/python/tasks/managed/collect_data.py
querti: The warning log has exc_info=True, so it should print the traceback

mmalina: ok, thanks, I was wondering what that would do :)

mmalina:
> > > The check-data-keys validation (previously a separate Tekton step)
> > > is incorporated directly into the collect_data script
> > 
> > 
> > Can you clarify this? AFAICS, check-data-keys is a complete task with its own python script. Or is this something else?
> 
> I meant this step: https://github.com/konflux-ci/release-service-catalog/blob/development/tasks/managed/collect-data/collect-data.yaml#L323 , sorry if it was unclear.

ah, thanks, I should have checked the task. so this was check-data-key-sources - checking if the keys came from allowed source (R, RP, RPA). Whereas the check-data-keys task checks the data structure against the schema.

mmalina (approved):

theflockers (approved):

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-06T13:16:16.247Z |

Approved

PR #929, opened 2026-07-30
fix(KONFLUX-13739): strip tag/digest in fbc query

theflockers:
fbc_opt_in is a repository-level attribute. Strip
tag or digest from pull specs before querying
Pyxis to avoid incorrect URL paths.

Assisted-by: Claude

### DIFF
--- a/scripts/python/helpers/image_ref.py
+++ b/scripts/python/helpers/image_ref.py
@@ -104,6 +104,19 @@ def resolve_quay_digest_to_git_sha(digest: str, container_image: str) -> str | N
         return None
 
 
+def strip_tag_and_digest(pull_spec: str) -> str:
+    """Return the repository portion of a pull spec, without tag or digest.
+
+    Strips the digest (``@sha256:...``) first, then strips the tag only
+    from the last path segment so a registry port (e.g. ``registry:5000``)
+    is never confused with a tag separator.
+    """
+    repo = pull_spec.split("@")[0]
+    parts = repo.split("/")
+    parts[-1] = parts[-1].rsplit(":", 1)[0]
+    return "/".join(parts)
+
+
 def pyxis_url_for_pull_spec(pyxis_url: str, pull_spec: str) -> str:
     """Build the Pyxis repository/tag API URL for `pull_spec`.
 
--- a/scripts/python/tasks/internal/check_fbc_opt_in.py
+++ b/scripts/python/tasks/internal/check_fbc_opt_in.py
@@ -57,13 +57,7 @@ def get_fbc_opt_in(
     """
     try:
         # fbc_opt_in is a repository-level attribute; strip tag and digest.
-        # Strip digest first, then strip the tag only from the last path
-        # segment so a registry port (e.g. registry:5000) is never confused
-        # with a tag separator.
-        repo_spec = pull_spec.split("@")[0]
-        parts = repo_spec.split("/")
-        parts[-1] = parts[-1].rsplit(":", 1)[0]
-        repo_spec = "/".join(parts)
+        repo_spec = image_ref.strip_tag_and_digest(pull_spec)
         if repo_spec != pull_spec:
             logger.info(
                 "Pull spec %s includes a tag/digest; stripped to %s for repo query",
--- a/scripts/python/tasks/internal/check_fbc_opt_in.py
+++ b/scripts/python/tasks/internal/check_fbc_opt_in.py
@@ -13,16 +13,15 @@ from collections.abc import Callable
 from pathlib import Path
 from typing import Any
 
-import requests
-from requests.auth import AuthBase
-from requests_kerberos import HTTPKerberosAuth, OPTIONAL
-
 import authentication
 import file
 import http_client
 import image_ref
+import requests
 import tekton
 from logger import logger
+from requests.auth import AuthBase
+from requests_kerberos import OPTIONAL, HTTPKerberosAuth
 
 PROG = "check_fbc_opt_in.py"
 
@@ -32,11 +31,11 @@ def parse_container_images(value: str) -> list[str]:
     # Input is a JSON array string; keep validation strict so malformed input fails fast.
     data = json.loads(value)
     if not isinstance(data, list):
-        raise ValueError("container images must be a JSON array")
+        raise TypeError("container images must be a JSON array")
     out: list[str] = []
     for item in data:
         if not isinstance(item, str):
-            raise ValueError("container images must be strings")
+            raise TypeError("container images must be strings")
         stripped = item.strip()
         if not stripped:
             raise ValueError("container images must be non-empty")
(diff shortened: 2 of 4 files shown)


scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -183,9 +192,10 @@ def main() -> int:
             service_account_mount,
             iib_services_config_mount,
         )
-    except (ValueError, tekton.CheckStepError) as e:
-        # Surface validation and step failures as a clear one-line SystemExit message.
-        raise SystemExit(f"{PROG}: {e}") from e
+    except (TypeError, ValueError, tekton.CheckStepError) as e:
+        logger.error("%s: %s", PROG, e)
+        rpath.write_text(json.dumps([]), encoding="utf-8")
+        return 0
```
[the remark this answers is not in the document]
theflockers: addressed — logger.error now uses exc_info=True to include the full traceback in Tekton/CI logs for debugging."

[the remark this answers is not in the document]
theflockers: ● "Addressed — the except clause now catches Exception broadly (matching the pattern used by other internal tasks like create_advisory and push_artifacts_to_cdn), with exc_info=True for full traceback logging."

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -51,17 +50,28 @@ def get_fbc_opt_in(
...
+        # segment so a registry port (e.g. registry:5000) is never confused
+        # with a tag separator.
+        repo_spec = pull_spec.split("@")[0]
+        parts = repo_spec.split("/")
+        parts[-1] = parts[-1].rsplit(":", 1)[0]
+        repo_spec = "/".join(parts)
+        if repo_spec != pull_spec:
+            logger.info(
+                "Pull spec %s includes a tag/digest; stripped to %s for repo query",
+                pull_spec,
```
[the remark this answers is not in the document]
theflockers: ● "The original code already logged this same INFO line — we only changed the stripping logic, not the logging. The message fires once per image, matching the existing per-image INFO logs in run_check. This is consistent with the prior behavior."

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -183,9 +192,10 @@ def main() -> int:
             service_account_mount,
             iib_services_config_mount,
         )
-    except (ValueError, tekton.CheckStepError) as e:
-        # Surface validation and step failures as a clear one-line SystemExit message.
-        raise SystemExit(f"{PROG}: {e}") from e
+    except Exception as e:
+        logger.error("%s: %s", PROG, e, exc_info=True)
+        rpath.write_text(json.dumps([]), encoding="utf-8")
+        return 0
```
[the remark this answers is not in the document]
theflockers: ● "If the Tekton result path is unwritable, the task is fundamentally misconfigured — no amount of error handling can write a result to an unwritable path. This applies equally to the success path's rpath.write_text() on line 205. Tekton guarantees the result paths are writable at
  runtime; guarding against this would be defending against a broken runtime, not application logic."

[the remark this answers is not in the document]
theflockers: "The fail-closed design is intentional and consistent with get_fbc_opt_in() itself, which returns False (opt-out) on any query failure. Writing [] means no images are opted in — the safe, conservative outcome. Downstream tasks treat missing opt-ins as opt-out, so no work is
  silently lost. The logger.error with exc_info=True makes failures clearly visible in Tekton logs for debugging."

src/tasks/internal/check_fbc_opt_in/check_fbc_opt_in.py
```diff
@@ -183,9 +192,10 @@ def main() -> int:
             service_account_mount,
             iib_services_config_mount,
         )
-    except (ValueError, tekton.CheckStepError) as e:
-        # Surface validation and step failures as a clear one-line SystemExit message.
-        raise SystemExit(f"{PROG}: {e}") from e
+    except Exception as e:
+        logger.error("%s: %s", PROG, e, exc_info=True)
+        rpath.write_text(json.dumps([]), encoding="utf-8")
```
querti, line 195: These seem like undocumented changes.  What does this do?

scripts/python/tasks/internal/check_fbc_opt_in.py
```diff
@@ -51,17 +50,28 @@ def get_fbc_opt_in(
...
     """
     try:
-        # Build the Pyxis endpoint for this image pull spec.
+        # fbc_opt_in is a repository-level attribute; strip tag and digest.
+        # Strip digest first, then strip the tag only from the last path
+        # segment so a registry port (e.g. registry:5000) is never confused
+        # with a tag separator.
+        repo_spec = pull_spec.split("@")[0]
+        parts = repo_spec.split("/")
+        parts[-1] = parts[-1].rsplit(":", 1)[0]
```
lslebodn: wrapping "parsing" in class might improve understanding of the code e.g.
```
image_name = ContainerImageFQDN(pull_spec) 

repo_spec = image_name.registry + image_name.repository
```
or maybe even some  method/property there. 

But feel free to close.

src/tasks/internal/check_fbc_opt_in/check_fbc_opt_in.py
theflockers: Internal requests should exit 0 to be able to set the result in case of error. That is a regular convention for internal tasks. Claude caught that this task was not consistent with the other internal tasks.

scripts/python/tasks/internal/check_fbc_opt_in.py
theflockers: added a function to strip the pull_spec.

querti (approved):

scripts/python/helpers/image_ref.py
```diff
@@ -104,6 +104,19 @@ def resolve_quay_digest_to_git_sha(digest: str, container_image: str) -> str | N
         return None
 
 
+def strip_tag_and_digest(pull_spec: str) -> str:
+    """Return the repository portion of a pull spec, without tag or digest.
```
lslebodn: It turns registry + repository and not just repository

theflockers: comment fixed.

scripts/python/tasks/internal/check_fbc_opt_in.py
lslebodn: TYVM. Feel free the resolve the thread. I do not have permissions.

scripts/python/helpers/image_ref.py
lslebodn: TYVM. Feel free the resolve the thread. I do not have permissions.

elenagerman (dismissed):

elenagerman (approved):

src/tasks/internal/check_fbc_opt_in/check_fbc_opt_in.py
```diff
@@ -59,9 +59,16 @@ def get_fbc_opt_in(
...
+        repo_spec = image_ref.strip_tag_and_digest(pull_spec)
+        if repo_spec != pull_spec:
+            logger.info(
+                "Pull spec %s includes a tag/digest; stripped to %s for repo query",
+                pull_spec,
+                repo_spec,
+            )
         body = http_client.get_text(
-            image_ref.pyxis_url_for_pull_spec(pyxis_url, pull_spec),
+            image_ref.pyxis_url_for_pull_spec(pyxis_url, repo_spec),
```
johnbieren, line 71: is pyxis_url_for_pull_spec poorly named? it also looks to me that that function already strips some stuff. Can you help me understand this a little more?

theflockers: The previous versions expects a wrong input (full pull_spec containing tags or image sha) but the correct is receiving the `repo` part only because `fbc_opt_in` flag lives there, not in the tag/image part.

After we strip the tags keeping the repo level, then `pyxis_url_for_pull_spec` adjust it to the expected pyxis `repository` url, for instance:

Receives: `r.io/repo/img:v1` return `https://pyxis/repository/repo/img/tag/v1`.

The current change transforms the input to `r.io/repo` then `pyxis_url_for_pull_spec`  returns `https://pyxis/repository/repo`, as we don't need the tags.

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-09-02T12:30:06.901Z |

Approved

PR #940, opened 2026-08-04
fix(KFLUXSPRT-8616): resolve binary collision when multiple files share the same os/arch

happybhati:
## Summary

Fixes a bug where components with multiple file entries sharing the same `(os, arch)` tuple (e.g. rhel8 and rhel9 Linux/amd64 binaries for oc-mirror) produce identical output tarballs because both archives are extracted into the same intermediate directory, causing the second to overwrite the first.

**Root cause**: `push_unsigned.py::_unpack_file_entries()` resolves the target directory via `oras_utils.os_arch_dir()`, which returns `linux/<arch>/` for all Linux entries regardless of the source archive. When two entries share `(linux, amd64)`, both extract into `linux/amd64/` and the second overwrites the first. `compress_artifacts.py` then re-tars the corrupted directory for both entries, producing identical output.

**Fix**: 
- Detect duplicate `(os, arch)` entries via a `Counter` computed once over the combined `files[]` and `staged.files[]` entries in `run()`, and isolate each into a per-source subdirectory (`linux/amd64/<archive-stem>/`)
- Update `_compress_file_entry()` to prefer the per-source subdirectory when it exists, with backward-compatible fallback for single-entry components
- Add `archive_stem()` helper to `oras_utils.py` for consistent `.tar.gz` extension stripping

## Key Changes

- `oras_utils.py`: New `archive_stem()` helper (strips `.tar.gz` extension)
- `push_unsigned.py`: `run()` computes `os_arch_counts` over combined file entries; `_unpack_file_entries()` uses per-archive subdirectories when collisions are detected
- `compress_artifacts.py`: `_compress_file_entry()` checks for per-source subdirectory before compressing

## Testing

- 3 new unit tests for `archive_stem()` parametrized cases
- 2 new tests for `_unpack_file_entries()`: multi-rhel isolation + different-binary-name isolation
- 1 new test for `_compress_file_entry()`: verifies distinct tarballs from per-source subdirectories
- All 101 tests pass with no regressions

## Affected Use Case

The oc-mirror-2.0 RPA has both rhel8 and rhel9 entries for each architecture, all with `os: linux`. This caused the rhel8 binary to overwrite the rhel9 binary during the unpack step, resulting in identical binaries being published to CDN.

Fixes: KFLUXSPRT-8616

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/compress_artifacts.py
+++ b/scripts/python/helpers/compress_artifacts.py
@@ -34,6 +34,7 @@ import shutil
 import subprocess
 import tarfile
 import zipfile
+from collections import Counter
 from pathlib import Path
 
 import disk_image_utils
@@ -104,6 +105,7 @@ def _compress_file_entry(
     ready_dir: Path,
     *,
     is_disk_image_component: bool = False,
+    os_arch_counts: Counter[tuple[str, str]] | None = None,
 ) -> str:
     """Compress one file entry into ready_dir and return the (possibly normalized) source path.
 
@@ -134,6 +136,11 @@ def _compress_file_entry(
     )
     if arch_dir is None:
         raise RuntimeError(f"Unknown OS '{os_name}' in {array_name}[] entry (arch: {arch})")
+
+    if os_arch_counts and os_arch_counts[(os_name, arch)] > 1:
+        per_source_dir = arch_dir / oras_utils.archive_stem(source_filename)
+        if per_source_dir.is_dir():
+            arch_dir = per_source_dir
 
     if not arch_dir.is_dir() or not any(arch_dir.iterdir()):
         raise RuntimeError(f"Architecture directory is empty or not found: {arch_dir}")
@@ -188,6 +195,10 @@ def compress_component(component: dict, snapshot: dict) -> dict:
 
     is_disk_image = disk_image_utils.is_disk_image_component(component)
 
+    os_arch_counts: Counter[tuple[str, str]] = Counter(
+        (e.get("os", ""), e.get("arch", "")) for e in files_entries + staged_entries
+    )
+
     normalized_files = []
     if files_entries:
         logger.info(
@@ -195,7 +206,12 @@ def compress_component(component: dict, snapshot: dict) -> dict:
         )
         for entry in files_entries:
             normalized_source = _compress_file_entry(
-                entry, "files", component_dir, ready_dir, is_disk_image_component=is_disk_image
+                entry,
+                "files",
+                component_dir,
+                ready_dir,
+                is_disk_image_component=is_disk_image,
+                os_arch_counts=os_arch_counts,
             )
             normalized_entry = dict(entry)
             # no-op for mac/linux, .zip correction for windows
@@ -213,6 +229,7 @@ def compress_component(component: dict, snapshot: dict) -> dict:
                 component_dir,
                 ready_dir,
                 is_disk_image_component=is_disk_image,
+                os_arch_counts=os_arch_counts,
             )
 
     updated_component = dict(component)
--- a/scripts/python/helpers/oras_utils.py
+++ b/scripts/python/helpers/oras_utils.py
@@ -131,6 +131,19 @@ def oras_push(tag: str, directory: Path, subdirectory: str, component_name: str)
     return match.group(1)
 
 
+def archive_stem(name: str) -> str:
+    """Strip the ``.tar.gz`` extension to produce a stem suitable for subdirectory names.
+
+    Example::
+
+        >>> archive_stem("oc-mirror-rhel9-linux-amd64.tar.gz")
+        'oc-mirror-rhel9-linux-amd64'
+    """
+    if name.endswith(".tar.gz"):
+        return name[: -len(".tar.gz")]
+    return Path(name).stem
+
+
 def os_arch_dir(
     os_name: str, arch: str, *, mac_windows_base: Path, linux_base: Path
 ) -> Path | None:
--- a/scripts/python/helpers/push_unsigned.py
+++ b/scripts/python/helpers/push_unsigned.py
@@ -28,6 +28,7 @@ import logging
 import os
 import shutil
 import tarfile
+from collections import Counter
 from pathlib import Path
 
 import disk_image_utils
@@ -88,8 +89,17 @@ def _unpack_file_entries(
     unsigned_dir: Path,
     *,
     is_disk_image_component: bool = False,
+    os_arch_counts: Counter[tuple[str, str]] | None = None,
 ) -> None:
     """Extract each archive from entries into its OS/arch subdirectory under unsigned_dir.
+
+    When multiple entries share the same ``(os, arch)`` (e.g. rhel8 and rhel9
+    variants), each archive is extracted into a per-source subdirectory
+    (``<os>/<arch>/<archive_stem>/``) to prevent files from colliding.
+
+    *os_arch_counts* should be pre-computed over the **combined** ``files[]`` and
+    ``staged.files[]`` entries so that collisions spanning both arrays are detected.
+    When ``None``, a local count is computed from *entries* alone (backward compat).
 
     Files are moved directly (without unpacking) when either:
     - *is_disk_image_component* is True (set when contentType: disk-image), or
@@ -97,6 +107,13 @@ def _unpack_file_entries(
       .raw.gz, .vhd.gz).
     All other files are treated as tar archives and extracted.
     """
+    if os_arch_counts is None:
+        os_arch_counts = Counter()
+        for entry in entries:
+            key = (entry.get("os", ""), entry.get("arch", ""))
+            if key[0] and key[1]:
+                os_arch_counts[key] += 1
+
     for entry in entries:
         source = entry.get("source", "")
         os_name = entry.get("os", "")
@@ -116,6 +133,12 @@ def _unpack_file_entries(
         )
         if target_dir is None:
             continue
+
+        if os_arch_counts[(os_name, arch)] > 1:
+            target_dir = target_dir / oras_utils.archive_stem(archive_name)
+            logger.info(
+                "  Multiple entries for (%s, %s); isolating to %s", os_name, arch, target_dir
+            )
 
         target_dir.mkdir(parents=True, exist_ok=True)
         if is_disk_image_component or disk_image_utils.is_disk_image_file(archive_name):
@@ -176,17 +199,28 @@ def run(quay_url: str, pipeline_run_uid: str) -> None:
             (component_dir / "linux").mkdir(parents=True, exist_ok=True)
 
         is_disk_image = disk_image_utils.is_disk_image_component(component)
+
+        files_entries = component.get("files") or []
+        staged_entries = (component.get("staged") or {}).get("files") or []
+        os_arch_counts: Counter[tuple[str, str]] = Counter()
+        for entry in files_entries + staged_entries:
+            key = (entry.get("os", ""), entry.get("arch", ""))
+            if key[0] and key[1]:
+                os_arch_counts[key] += 1
+
         _unpack_file_entries(
-            component.get("files") or [],
+            files_entries,
             component_dir,
             unsigned_dir,
             is_disk_image_component=is_disk_image,
+            os_arch_counts=os_arch_counts,
         )
         _unpack_file_entries(
-            (component.get("staged") or {}).get("files") or [],
+            staged_entries,
             component_dir,
             unsigned_dir,
             is_disk_image_component=is_disk_image,
+            os_arch_counts=os_arch_counts,
         )
 
         supp_hold = component_dir / "supplementary"
--- a/scripts/python/helpers/test_compress_artifacts.py
+++ b/scripts/python/helpers/test_compress_artifacts.py
@@ -502,3 +502,144 @@ def test_main_exception_returns_error() -> None:
     with mock.patch.object(compress_artifacts, "run", side_effect=RuntimeError("oras fail")):
         rc = compress_artifacts.main(["compress_artifacts.py", "--quay-url", "quay.io/org"])
     assert rc == 1
+
+
+# ---------------------------------------------------------------------------
+# multi-entry (os, arch) — per-source subdirectory support
+# ---------------------------------------------------------------------------
+
+
+def test_compress_file_entry_no_false_match_single_entry(tmp_path: Path) -> None:
+    """A single (os, arch) entry ignores a coincidental subdir matching archive_stem."""
+    comp_dir = tmp_path / "component"
+    ready_dir = comp_dir / "ready_for_distribution"
+    ready_dir.mkdir(parents=True)
+
+    arch_dir = comp_dir / "linux" / "amd64"
+    arch_dir.mkdir(parents=True)
+    (arch_dir / "real-binary").write_bytes(b"the-binary")
+
+    # Coincidental subdirectory whose name matches archive_stem
+    fake_subdir = arch_dir / "product-linux-amd64"
+    fake_subdir.mkdir()
+    (fake_subdir / "decoy").write_bytes(b"should-not-appear-alone")
+
+    from collections import Counter
+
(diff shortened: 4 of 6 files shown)


scripts/python/helpers/push_unsigned.py
```diff
@@ -91,12 +91,24 @@ def _unpack_file_entries(
...
+    variants), each archive is extracted into a per-source subdirectory
+    (``<os>/<arch>/<archive_stem>/``) to prevent files from colliding.
+
     Files are moved directly (without unpacking) when either:
     - *is_disk_image_component* is True (set when contentType: disk-image), or
     - the filename has an unambiguous disk-image suffix (.qcow2, .iso, .iso.gz,
       .raw.gz, .vhd.gz).
     All other files are treated as tar archives and extracted.
     """
+    from collections import Counter
```
swickersh: I'd move the import to the top please.

swickersh (commented):
We have scenarios today where teams release to same os/arch but have DIFFERENT binary names like a client and installer.  It would be good to ensure that user case is tested.  So for test_push_unsigned:

```
def test_unpack_file_entries_different_binaries_same_arch_isolation(tmp_path: Path) -> None:
    """Different-named binaries sharing (os, arch) don't cross-contaminate each other."""
    comp_dir = tmp_path / "prod"
    comp_dir.mkdir()

    installer_archive = comp_dir / "product-installer-linux-amd64.tar.gz"
    _make_tar(installer_archive, {"installer": b"installer-binary"})

    client_archive = comp_dir / "product-client-linux-amd64.tar.gz"
    _make_tar(client_archive, {"client": b"client-binary"})

    entries = [
        {"source": "/releases/product-installer-linux-amd64.tar.gz", "os": "linux", "arch": "amd64"},
        {"source": "/releases/product-client-linux-amd64.tar.gz", "os": "linux", "arch": "amd64"},
    ]

    push_unsigned._unpack_file_entries(entries, comp_dir, comp_dir / "unsigned")

    installer_dir = comp_dir / "linux" / "amd64" / "product-installer-linux-amd64"
    client_dir = comp_dir / "linux" / "amd64" / "product-client-linux-amd64"
    assert (installer_dir / "installer").exists()
    assert (client_dir / "client").exists()
    # the key assertion: no cross-contamination
    assert not (installer_dir / "client").exists()
    assert not (client_dir / "installer").exists()
```

and in test_compress_artifacts:

```
def test_compress_file_entry_per_source_subdir_different_binaries(tmp_path: Path) -> None:
    """Different-named binaries sharing (os, arch) don't cross-contaminate output tarballs."""
    comp_dir = tmp_path / "component"
    ready_dir = comp_dir / "ready_for_distribution"
    ready_dir.mkdir(parents=True)

    installer_dir = comp_dir / "linux" / "amd64" / "product-installer-linux-amd64"
    installer_dir.mkdir(parents=True)
    (installer_dir / "installer").write_bytes(b"installer-binary")

    client_dir = comp_dir / "linux" / "amd64" / "product-client-linux-amd64"
    client_dir.mkdir(parents=True)
    (client_dir / "client").write_bytes(b"client-binary")

    compress_artifacts._compress_file_entry(
        {"source": "/releases/product-installer-linux-amd64.tar.gz", "os": "linux", "arch": "amd64"},
        "files",
        comp_dir,
        ready_dir,
    )
    compress_artifacts._compress_file_entry(
        {"source": "/releases/product-client-linux-amd64.tar.gz", "os": "linux", "arch": "amd64"},
        "files",
        comp_dir,
        ready_dir,
    )

    import tarfile

    installer_out = ready_dir / "product-installer-linux-amd64.tar.gz"
    client_out = ready_dir / "product-client-linux-amd64.tar.gz"

    with tarfile.open(str(installer_out), "r:gz") as tf:
        installer_names = tf.getnames()
    with tarfile.open(str(client_out), "r:gz") as tf:
        client_names = tf.getnames()

    # the key assertion: each tarball contains only its own binary
    assert installer_names == ["installer"]
    assert client_names == ["client"]
```

scripts/python/helpers/oras_utils.py
```diff
@@ -131,6 +131,21 @@ def oras_push(tag: str, directory: Path, subdirectory: str, component_name: str)
     return match.group(1)
 
 
+def archive_stem(name: str) -> str:
+    """Strip archive extensions to produce a stem suitable for subdirectory names.
+
+    Handles double extensions like ``.tar.gz`` and ``.tar.bz2``::
```
swickersh: .tar.bz2 isn't an input tarball we support. Only .tar.gz according to our docs and current RPAs unless something change I'm unaware of

scripts/python/helpers/push_unsigned.py
```diff
@@ -91,12 +91,24 @@ def _unpack_file_entries(
...
+
+    os_arch_counts: Counter[tuple[str, str]] = Counter()
```
swickersh: This Counter is scoped to a single call of _unpack_file_entries. But run() calls this function twice per component — once for files[], once for staged.files[] — passing the same component_dir/unsigned_dir both times. For Linux entries, both calls resolve to component_dir/linux/<arch> regardless of which array they came from.

If a collision only exists across the two arrays — e.g. one linux/amd64 entry in files[] and a different linux/amd64 entry in staged.files[] — neither call ever sees count > 1 for that (os, arch) pair, so isolation never triggers and both entries extract flat into the same directory. That's the exact bug this PR fixes, just split across files[]/staged.files[] instead of within one array, and since files[] feeds Developer Portal/CGW while staged.files[] feeds CDN/Pulp, it could mean content meant for one delivery channel leaks into the artifact shipped to the other.

Suggested fix: compute os_arch_counts once over the combined files[] + staged.files[] entries in run() and pass it into _unpack_file_entries (or merge both entry lists before calling this function once), rather than letting each call build its own Counter from only the entries it was given.

happybhati:
Thanks Scott for the thorough review — all feedback addressed in the latest push:

1. **Import moved** — `from collections import Counter` is now at module level
2. **`.tar.bz2` removed** — `archive_stem()` only handles `.tar.gz` per current RPA support
3. **Counter scope fixed** — `os_arch_counts` is now computed once in `run()` over the combined `files[] + staged.files[]` entries and passed to `_unpack_file_entries()`, so cross-array collisions are caught
4. **Cross-contamination tests added** — both `test_unpack_file_entries_different_binaries_same_arch_isolation` and `test_compress_file_entry_per_source_subdir_different_binaries` are in, verifying installer/client binaries stay isolated

All 101 tests pass.

scripts/python/helpers/push_unsigned.py
happybhati: Done — moved to module-level imports.

scripts/python/helpers/oras_utils.py
happybhati: Good call — removed .tar.bz2 handling and its test case. Only .tar.gz is supported per current RPAs.

scripts/python/helpers/push_unsigned.py
happybhati: Excellent catch — fixed by computing os_arch_counts once in run() over the combined files[] + staged.files[] entries and passing it to _unpack_file_entries().

scripts/python/helpers/compress_artifacts.py
```diff
@@ -135,6 +135,10 @@ def _compress_file_entry(
     if arch_dir is None:
         raise RuntimeError(f"Unknown OS '{os_name}' in {array_name}[] entry (arch: {arch})")
 
+    per_source_dir = arch_dir / oras_utils.archive_stem(source_filename)
+    if per_source_dir.is_dir():
+        arch_dir = per_source_dir
+
```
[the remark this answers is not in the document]
swickersh: This one sounds legit.
Should probably implement and add a test case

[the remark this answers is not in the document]
happybhati: Implemented — `_compress_file_entry()` now accepts an explicit `os_arch_counts` parameter instead of inferring from filesystem state. `compress_component()` computes the Counter over combined `files[] + staged.files[]` and threads it through, matching the approach in `push_unsigned.run()`. Added `test_compress_file_entry_no_false_match_single_entry` to cover the false-match scenario.

ashwindasr:
thanks @happybhati @swickersh for working on this. was checking in to see if there was any blockers to getting this merged?

happybhati:
> thanks @happybhati @swickersh for working on this. was checking in to see if there was any blockers to getting this merged?

Our e2e was down for the past couple of days. I will try to test things again and would probably need another round of review.  Sorry for the delay :/

happybhati:
Waiting for limit to get reset; e2e failing.
```
🔴 error: failed to create repository hacbs-release-tests/catalog-utils-e2e-5ebb67a4-12e2-4fab-aa14-178f1741d89e
Response: {
  "message": "You have exceeded a secondary rate limit and have been temporarily blocked from content creation. Please retry your request again later. For more on scraping GitHub and how it may affect your rights, please review our Terms of Service (https://docs.github.com/en/site-policy/github-terms/github-terms-of-service) If you reach out to GitHub Support for help, please include the request ID 2DA8:BA5E4:63F0E4:14F4894:6A7F2289 and timestamp 2026-08-14 14:13:29 UTC.",
  "documentation_url": "/rest/overview/rate-limits-for-the-rest-api#about-secondary-rate-limits",
  "status": "403"
}
Error message: You have exceeded a secondary rate limit and have been temporarily blocked from content creation. Please retry your request again later. For more on scraping GitHub and how it may affect your rights, please review our Terms of Service (https://docs.github.com/en/site-policy/github-terms/github-terms-of-service) If you reach out to GitHub Support for help, please include the request ID 2DA8:BA5E4:63F0E4:14F4894:6A7F2289 and timestamp 2026-08-14 14:13:29 UTC.
```

happybhati:
Github is down!

ashwindasr:
/retest

GH seems to be back online

happybhati:
```
 error: failed to create repository hacbs-release-tests/catalog-utils-e2e-58e0a9e3-9ef0-4e22-be01-954530f0d465
Response: {
  "message": "You have exceeded a secondary rate limit and have been temporarily blocked from content creation. Please retry your request again later. For more on scraping GitHub and how it may affect your rights, please review our Terms of Service (https://docs.github.com/en/site-policy/github-terms/github-terms-of-service) If you reach out to GitHub Support for help, please include the request ID D84E:28299:441FB37:E6F631B:6A8487ED and timestamp 2026-08-18 16:27:25 UTC.",
  "documentation_url": "/rest/overview/rate-limits-for-the-rest-api#about-secondary-rate-limits",
  "status": "403"
}
```

/retest

swickersh (approved):

happybhati:
/retest
```
🔴 error: failed to create repository hacbs-release-tests/catalog-utils-e2e-48a1e229-f4f5-47b8-b77c-06b1d2d1ba1f
Response: {
	"message": "API rate limit exceeded for user ID 157044195. If you reach out to GitHub Support for help, please include the request ID CDFD:4178B:1632AA6:49E09ED:6A8DB538 and timestamp 2026-08-25 15:31:04 UTC. For more on scraping GitHub and how it may affect your rights, please review our Terms of Service (https:\/\/docs.github.com\/en\/site-policy\/github-terms\/github-terms-of-service)",
	"documentation_url": "https://docs.github.com/en/rest/using-the-rest-api/getting-started-with-the-rest-api#rate-limiting",
	"status": "403"
```

swickersh:
unfortunately I think there is one bug remaining.

Binary/generic advisory PURLs still collapse on (component, arch, os), so this PR does not fully fix the collision case.

filename_for_binary_or_generic() takes the first matching file, and _merge_updated_artifacts() dedupes by component|architecture|os. Disk-image already expands one row per file; binary/generic does not.

Concrete example: rhodf-cli ships noobaa + odf in one component with the same (os, arch). After this PR, Pulp/CDN will publish six distinct files, but the advisory will still emit three PURLs, all pointing at the first file of each pair.

Publish path is fine. PURL generation needs a follow-up to key binary/generic artifacts by source filename the same way disk-image already does.

happybhati:
@swickersh the same `(os, arch)` collision also affected advisory PURL generation (only the first file per arch/os got a PURL). Added a fix here mirroring the disk-image per-file expansion pattern, with regression tests. LMK if it makes sense to keep in this PR, or if you'd rather split it into its own Jira/PR.

swickersh:
> @swickersh the same `(os, arch)` collision also affected advisory PURL generation (only the first file per arch/os got a PURL). Added a fix here mirroring the disk-image per-file expansion pattern, with regression tests. LMK if it makes sense to keep in this PR, or if you'd rather split it into its own Jira/PR.

Looks good to keep it here. Same collision, one layer up, and I’d rather not ship distinct CDN files with collapsed advisory PURLs. Fine that it was pre-existing — thanks for folding it in!

src/helpers/oras_utils/oras_utils.py
```diff
@@ -131,6 +131,19 @@ def oras_push(tag: str, directory: Path, subdirectory: str, component_name: str)
...
+    """Strip the ``.tar.gz`` extension to produce a stem suitable for subdirectory names.
+
+    Example::
+
+        >>> archive_stem("oc-mirror-rhel9-linux-amd64.tar.gz")
+        'oc-mirror-rhel9-linux-amd64'
+    """
+    if name.endswith(".tar.gz"):
+        return name[: -len(".tar.gz")]
+    return Path(name).stem
```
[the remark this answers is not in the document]
happybhati: Only `.tar.gz` is a supported archive format for these RPA `files[]`/`staged.files[]` entries today — that's why `archive_stem()` only special-cases `.tar.gz` and otherwise falls back to `Path(name).stem` (this was actually simplified from a broader multi-suffix version per Scott's feedback above, specifically to avoid speculative handling for formats we don't ingest). Since `.tar.bz2`/`.tar.xz` can't appear as a `source` value in practice, the stem collision this flags isn't reachable with current inputs.

If/when we add support for another compressed archive format, we should revisit `archive_stem()` with the `Path(name).suffixes`-based approach suggested here so the per-source isolation stays collision-free. Leaving as-is for now to keep this PR scoped to the reported bug.

happybhati:
@swickersh are we good with this PR ? I will rebase one last time to verify test again so far all test passing. and temporary catalog PR also Passed all check with Utils tmp image.
https://github.com/konflux-ci/release-service-catalog/pull/2519

swickersh (approved):

src/helpers/content_gateway/tests/test_content_gateway.py
```diff
@@ -115,3 +115,52 @@ def test_filename_for_binary_or_generic_empty_when_no_match() -> None:
...
+            component,
+            architecture="aarch64",
+            operating_system="linux",
+        )
+        == []
+    )
+
+
+def test_filename_for_binary_or_generic_uses_first_of_multiple_matches() -> None:
+    """The singular helper still returns just the first match, for compatibility."""
```
johnbieren: So the singular helper is still used? I don't know what "for compatibility" means here

happybhati: Good catch, thanks - it wasn't actually used anywhere, the "for compatibility" docstring was misleading. Removed `filename_for_binary_or_generic` entirely (function, export, and its dedicated tests) and moved the relevant coverage onto `filenames_for_binary_or_generic`, which is the one real callers use. Pushed as an amend; ran the full suite locally (3290 passed) plus ruff/black clean.

happybhati:
Addressed Johnny's feedback above (removed the unused `filename_for_binary_or_generic` helper). Also re-tested the newly built utils image end-to-end against the CDN pipeline on our validation catalog PR (konflux-ci/release-service-catalog#2519) — all 55 checks green, including `push-artifacts-to-cdn-e2e-test`. Should be ready for review/approval again. Thanks a lot for all the reviews so far! No Rush 🙇

johnbieren (approved):

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Timestamp** | 2026-09-11T14:54:47.539Z |

swickersh (approved):

PR #942, opened 2026-08-05
fix(KONFLUX-13766): double-quote version strings in advisory YAML

seanconroy2021:
PyYAML writes "9.9" or "1" with single quotes not double, which differs from the template output. Added a custom dumper to force double quotes when a string looks like a number. int/bool fields are not affected.

Assisted-by: Claude

### DIFF
--- a/scripts/python/helpers/advisory_data.py
+++ b/scripts/python/helpers/advisory_data.py
@@ -264,11 +264,28 @@ def template_context_merge(
     }
 
 
+class _Dumper(yaml.SafeDumper):
+    pass
+
+
+def _quote_ambiguous_strings(dumper: yaml.Dumper, data: str) -> yaml.ScalarNode:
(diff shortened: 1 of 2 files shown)


FilipNikolovski (approved):
lgtm

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-11T10:51:45.197Z |

Approved

seanconroy2021:
So was testing this and PyYAML leaves "1.0.1" unquoted since it's not ambiguous. So have added a second check to quote any string that's just digits and dots regardless of whether YAML would mistype it.

FilipNikolovski (approved):

johnbieren (approved):

PR #945, opened 2026-08-06
fix(RELEASE-2747): extract_artifacts.py better symlink checks

johnbieren:
This commit updates extract_artifacts.py to not fail whenever there is a symlink in the extracted directory tree and instead only fail if a symlink in the extracted directory is a wanted file from the RPA.

Assisted-By: Cursor

### DIFF
--- a/scripts/python/helpers/extract_artifacts.py
+++ b/scripts/python/helpers/extract_artifacts.py
@@ -97,27 +97,49 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
     return unique_files, unique_dirs
 
 
+def _normalize_tar_member_name(name: str) -> str:
+    """Normalize a tar member path for prefix matching and safe extraction."""
+    normalized = name.lstrip("/")
+    while normalized.startswith("./"):
+        normalized = normalized[2:]
+    return normalized
+
+
 def _safe_extract_layer(
-    tf: tarfile.TarFile, image_path: str, target_dir: Path, layer_name: str
+    tf: tarfile.TarFile,
+    image_path: str,
+    target_dir: Path,
+    layer_name: str,
 ) -> bool:
     """Extract members under image_path/ from a layer tarfile with path safety checks.
 
-    Returns True if any matching members were found, False otherwise.
-    Raises RuntimeError for unsafe entries (path traversal, symlinks, hardlinks, devices).
+    Symlink, hardlink, and device entries are always skipped (not extracted). Later
+    layers may replace a skipped special with a regular file; final wanted-path
+    validation happens in process_component after all layers are merged. Path
+    traversal is rejected for every member that is extracted.
+
+    Returns True if any matching members were found under image_path, False otherwise.
     """
     target_real = target_dir.resolve()
     found = False
     for member in tf.getmembers():
-        if not (member.name == image_path or member.name.startswith(f"{image_path}/")):
+        normalized = _normalize_tar_member_name(member.name)
+        if not (normalized == image_path or normalized.startswith(f"{image_path}/")):
             continue
         found = True
         if member.issym() or member.islnk() or member.isdev():
-            raise RuntimeError(
-                f"Layer {layer_name} contains unsupported entry type: {member.name}"
+            logger.debug(
+                "Skipping unsupported entry type in layer %s: %s",
+                layer_name,
+                member.name,
             )
-        member_real = (target_dir / member.name).resolve()
+            continue
+        # Use the normalized relative path so absolute tar names (e.g. /releases/x)
+        # do not resolve outside target_dir and still land under the extract root.
+        member_real = (target_dir / normalized).resolve()
         if member_real != target_real and target_real not in member_real.parents:
             raise RuntimeError(f"Layer {layer_name} contains unsafe path: {member.name}")
+        member.name = normalized
         tf.extract(member, path=str(target_dir), filter="data")
     return found
 
@@ -245,13 +267,19 @@ def process_component(component: dict) -> None:
 
             for wanted in wanted_files:
                 src = tmp_dir / wanted
-                if src.is_file():
-                    shutil.copy2(str(src), str(destination / src.name))
-                else:
+                # Validate final merged state after all layers; never follow symlinks.
+                if src.is_symlink():
+                    raise RuntimeError(
+                        f"File '{wanted}' declared in RPA resolved to a symlink; "
+                        "wanted artifacts must be regular files"
+                    )
+                if not src.is_file():
                     logger.error("Expected file not found in container: %s", wanted)
                     raise RuntimeError(
-                        f"File '{wanted}' declared in RPA was not found in any container layer"
+                        f"File '{wanted}' declared in RPA was not found in any "
+                        "container layer"
                     )
+                shutil.copy2(str(src), str(destination / src.name))
     finally:
         shutil.rmtree(tmp_dir, ignore_errors=True)
 
--- a/scripts/python/helpers/test_extract_artifacts.py
+++ b/scripts/python/helpers/test_extract_artifacts.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import io
 import json
 import logging
 import tarfile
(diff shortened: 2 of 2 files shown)


scripts/python/helpers/extract_artifacts.py
```diff
@@ -97,24 +97,50 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
...
+    Returns True if any matching members were found under image_path, False otherwise.
     """
+    wanted_set = set(wanted_files)
     target_real = target_dir.resolve()
     found = False
     for member in tf.getmembers():
-        if not (member.name == image_path or member.name.startswith(f"{image_path}/")):
+        normalized = _normalize_tar_member_name(member.name)
+        if not (normalized == image_path or normalized.startswith(f"{image_path}/")):
             continue
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/helpers/extract_artifacts.py
```diff
@@ -97,27 +97,56 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
...
-    Returns True if any matching members were found, False otherwise.
-    Raises RuntimeError for unsafe entries (path traversal, symlinks, hardlinks, devices).
+    Symlink, hardlink, and device entries that are not in wanted_files are skipped
+    (not extracted). Those same types raise if the member path is a wanted artifact
+    — RPA-requested files must be regular files. Path traversal is rejected for
+    every member that is extracted.
+
+    Returns True if any matching members were found under image_path, False otherwise.
     """
+    wanted_set = set(wanted_files)
```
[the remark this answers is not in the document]
johnbieren: fixed

scripts/python/helpers/extract_artifacts.py
```diff
@@ -97,27 +97,56 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
...
+            logger.debug(
+                "Skipping unsupported entry type in layer %s: %s",
+                layer_name,
+                member.name,
             )
-        member_real = (target_dir / member.name).resolve()
+            continue
+        # Use the normalized relative path so absolute tar names (e.g. /releases/x)
+        # do not resolve outside target_dir and still land under the extract root.
+        member_real = (target_dir / normalized).resolve()
```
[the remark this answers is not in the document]
johnbieren: this is overkill

scripts/python/helpers/extract_artifacts.py
```diff
@@ -97,27 +97,56 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
...
-        if not (member.name == image_path or member.name.startswith(f"{image_path}/")):
+        normalized = _normalize_tar_member_name(member.name)
+        if not (normalized == image_path or normalized.startswith(f"{image_path}/")):
             continue
         found = True
         if member.issym() or member.islnk() or member.isdev():
-            raise RuntimeError(
-                f"Layer {layer_name} contains unsupported entry type: {member.name}"
+            if normalized in wanted_set:
+                raise RuntimeError(
```
mmalina: This sounds like a legitimate concern from Claude, if an edge case:

```
  Main finding: premature per-layer failure can break legitimate images

  In _safe_extract_layer, when a symlink/hardlink/device member's path exactly
  matches a wanted file, it raises immediately:

  if member.issym() or member.islnk() or member.isdev():
      if normalized in wanted_set:
          raise RuntimeError(...)

  process_component calls this once per layer, iterating layer_digests in
  manifest order — which for OCI images is base-layer-first, top-layer-last
  (union filesystem semantics: later layers shadow earlier ones).

  Problem: if a base layer happens to have a symlink at the wanted path (e.g.
  left over from an earlier build step, or a placeholder that gets replaced
  later), the code raises and aborts the whole extraction on that base layer —
  even if a later layer legitimately overwrites that path with a real regular
  file, which is exactly what would end up on disk in the real container.

  This directly undercuts the PR's own goal (stop failing extraction due to
  incidental symlinks). Notice the file already has a proper mechanism for this:
  after all layers are extracted, process_component does a final
  src.is_symlink() check on the fully-merged result — that's the correct place
  to enforce "wanted files must not be symlinks," since it reflects final state.
  The per-layer raise inside _safe_extract_layer is redundant with that check
  and, worse, is more strict than the final check because it fires on
  transient/shadowed state rather than final state.

  Telling detail: the PR adds
  test_safe_extract_layer_multilayer_wanted_symlink_overwrite_fails, which tests
  exactly the opposite ordering (real file in layer1, symlink overwrite in
  layer2 → correctly fails). There's no test for
  base-layer-symlink-then-real-file-in-top-layer, which is the scenario that
  would currently regress.

  Suggested fix: drop the immediate raise in _safe_extract_layer for wanted-path
  symlinks/hardlinks/devices (just skip them like any other non-wanted special
  file), and rely solely on the existing post-extraction src.is_symlink() /
  is_file() checks in process_component, which already correctly validate final
  state after all layers are merged.
```

Unless Claude is missing something here, why isn't the existing check for symlink at the end sufficient here (possibly extending it to check for hardlink and dev as well)?

scripts/python/helpers/extract_artifacts.py
```diff
@@ -97,27 +97,56 @@ def _get_source_paths(component: dict) -> tuple[list[str], list[str]]:
...
-    Returns True if any matching members were found, False otherwise.
-    Raises RuntimeError for unsafe entries (path traversal, symlinks, hardlinks, devices).
+    Symlink, hardlink, and device entries that are not in wanted_files are skipped
+    (not extracted). Those same types raise if the member path is a wanted artifact
+    — RPA-requested files must be regular files. Path traversal is rejected for
+    every member that is extracted.
+
+    Returns True if any matching members were found under image_path, False otherwise.
     """
+    wanted_set = {_normalize_tar_member_name(w) for w in wanted_files}
```
mmalina: Minor:
   
  - wanted_set = {_normalize_tar_member_name(w) for w in wanted_files} is
  recomputed on every _safe_extract_layer call (once per layer × per extract
  dir). Cheap in absolute terms, but could be hoisted and passed in as a
  pre-built set from process_component to avoid repeated work.

scripts/python/helpers/extract_artifacts.py
```diff
@@ -245,13 +276,19 @@ def process_component(component: dict) -> None:
...
+                        "wanted artifacts must be regular files"
+                    )
+                if not src.is_file():
                     logger.error("Expected file not found in container: %s", wanted)
                     raise RuntimeError(
-                        f"File '{wanted}' declared in RPA was not found in any container layer"
+                        f"File '{wanted}' declared in RPA was not found in any "
+                        "container layer"
                     )
+                shutil.copy2(str(src), str(destination / src.name), follow_symlinks=False)
```
mmalina: - shutil.copy2(..., follow_symlinks=False) right after an explicit if
  src.is_symlink(): raise is dead weight — follow_symlinks only changes behavior
  when the source is a symlink, which is already excluded by that point.
  Harmless, just redundant.

scripts/python/helpers/extract_artifacts.py
johnbieren: fixed

scripts/python/helpers/extract_artifacts.py
johnbieren: fixed

scripts/python/helpers/extract_artifacts.py
johnbieren: good catch, fixed

mmalina (approved):

kurlov:
@johnbieren once this PR gets merged when it will be available on stage/prod? This bug blocks ACS binary publishing migration and I wonder when should I test publishing on stage env

johnbieren:
> @johnbieren once this PR gets merged when it will be available on stage/prod? This bug blocks ACS binary publishing migration and I wonder when should I test publishing on stage env

Shortly after this merges, I will open the catalog PR to bump the image used in the task. Once that merges, it will be available in the development branch of release-service-catalog. It will then go to staging the next Monday and production a week later. ACS may want to switch (temporarily) to the development branch in their RPA if they want the fix sooner (but it isn't available at all until I get the catalog PR based on this one merged)

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-12T09:30:00.106Z |

Approved

johnbieren:
> @johnbieren once this PR gets merged when it will be available on stage/prod? This bug blocks ACS binary publishing migration and I wonder when should I test publishing on stage env

catalog PR https://github.com/konflux-ci/release-service-catalog/pull/2457

PR #952, opened 2026-08-10
feat(RELEASE-2519): add push_snapshot.py script

querti:
Add a python script to replicate the functionality of the inline bash script in the push-snapshot managed task in the catalog repo. The task will be updated to use this python module instead.

Add an image_ref helper for parsing image references, and extend oras_utils with an oras_cp wrapper and an auth_file option for oras_resolve. Also add memory_throttle (cgroup-aware memory pressure throttling), copied from PR konflux-ci#901.

Assisted-by: Cursor


scripts/python/helpers/image_ref.py
```diff
@@ -13,6 +13,58 @@
...
+    'quay.io/org/repo'
+    """
+    without_digest = ref.split("@", 1)[0]
+    last_segment = without_digest.rsplit("/", 1)[-1]
+    if ":" in last_segment:
+        without_digest = without_digest.rsplit(":", 1)[0]
+    return without_digest
+
+
+def digest(ref: str) -> str:
```
seanconroy2021: I don't see this function being called anywhere except in tests, so why do we need it?

scripts/python/helpers/image_ref.py
```diff
@@ -13,6 +13,58 @@
...
+    'sha256:abc123'
+
+    Raises ``ValueError`` when the reference has no ``@`` digest marker.
+    """
+    if "@" not in ref:
+        raise ValueError(f"reference has no digest: {ref!r}")
+    return ref.split("@", 1)[1]
+
+
+def repo_path(ref: str) -> str:
```
seanconroy2021: same here don't see this function being called anywhere except in tests, so why do we need it?

src/tasks/managed/push_snapshot/tests/test_push_snapshot.py
```diff
@@ -0,0 +1,898 @@
...
+            [], 0, stdout=json.dumps({"mediaType": idx}), stderr=""
+        )
+
+        push_snapshot.run(snapshot_file, data_file, results_dir, 5, 0, False)
+
+        for c in run_mocks.push_image.call_args_list:
+            job = c[0][0]
+            assert job.platform == "linux/amd64"
+
+    def test_migration_artifacts(self, tmp_path: Path, run_mocks) -> None:
```
seanconroy2021, line 758: Could add a test for copy_bundle_migrations=True with no migration annotations on the component. Which would test push_migration_artifact isn't called.

scripts/python/tasks/managed/push_snapshot.py
```diff
@@ -0,0 +1,646 @@
...
+    oras has very limited support for selecting the right auth entry from
+    a multi-entry credentials file, so we create a custom auth file with
+    just one entry keyed by the full repository path (instead of the bare
+    registry hostname) so that ``oras resolve`` and ``cosign copy`` can
+    match credentials correctly.
+    """
+    reg = image_ref.registry(container_image)
+    source_repo = image_ref.repository(container_image)
+    auth_json_str = select_oci_auth(container_image)
+    auth_data = json.loads(auth_json_str) if auth_json_str else {}
```
seanconroy2021: Isn't this dead code? select_oci_auth already returns at least {} so auth_json_str is never false. Could just use json.loads(auth_json_str) directly. Same for create_dest_auth_file.

scripts/python/helpers/image_ref.py
querti: Good catch, it must have been a leftover from an earlier iteration. Removed.

scripts/python/helpers/image_ref.py
querti: Removed as well.

src/tasks/managed/push_snapshot/tests/test_push_snapshot.py
querti: Good point, added.

scripts/python/tasks/managed/push_snapshot.py
querti: Good catch, removed pointless guard from both places.

seanconroy2021 (dismissed):
lgtm

davidmogar (dismissed):

davidmogar (dismissed):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-31T12:38:29.825Z |

Approved

davidmogar (approved):

PR #956, opened 2026-08-11
feat(RELEASE-2520): add reduce_snapshot.py script

querti:
Add a Python script to replace the inline bash logic of the reduce-snapshot managed task in the catalog repo.

The two original steps (ensure-required-labels-present and reduce) are consolidated into a single Python module that handles both label validation and snapshot filtering.

Fix a contradictory behavior in the original shell script: when the built component was not found in the snapshot, the script printed "Using original Snapshot" but then exited 1 without actually writing the snapshot file. The Python version now does what the message says — it falls back to the original snapshot and writes it out successfully.

Assisted-by: Cursor


midnightercz (dismissed):

scripts/python/tasks/managed/reduce_snapshot.py
```diff
@@ -0,0 +1,160 @@
...
+        return reduced
+
+    logger.warning(
+        "Reduced Snapshot has %d components (expected 1). "
+        "Verify that the Snapshot contains the built component: %s. "
+        "Using original Snapshot.",
+        len(filtered),
+        component_name,
+    )
+    return None
```
midnightercz: I know it's probably like this in original bash script, but it's little wierd that you return None instead of original and in parent method you overwrite original with reduced if reduced. It would make sense to just return snapshot, original or reduced.

querti: Good point, fixed.

happybhati (dismissed):

scripts/python/tasks/managed/reduce_snapshot.py
```diff
@@ -0,0 +1,156 @@
...
+import json
+import os
+from pathlib import Path
+
+import file
+import tekton
+from get_resource import get_resource
+from logger import logger
+
+PROG = "reduce_snapshot.py"
```
johnbieren: unused

johnbieren (commented):
Are you going to open a PR to remove https://github.com/conforma/cli/blob/main/hack/reduce-snapshot.sh?

scripts/python/tasks/managed/reduce_snapshot.py
querti: removed

querti:
> Are you going to open a PR to remove https://github.com/conforma/cli/blob/main/hack/reduce-snapshot.sh?

Looks like conforma repo itself still uses this script:
 https://github.com/conforma/cli/blob/main/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml#L354 
 https://github.com/conforma/cli/blob/main/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml#L300

johnbieren:
> > Are you going to open a PR to remove https://github.com/conforma/cli/blob/main/hack/reduce-snapshot.sh?
> 
> Looks like conforma repo itself still uses this script: https://github.com/conforma/cli/blob/main/tasks/verify-conforma-konflux-ta/0.1/verify-conforma-konflux-ta.yaml#L354 https://github.com/conforma/cli/blob/main/tasks/verify-enterprise-contract/0.1/verify-enterprise-contract.yaml#L300

Ack, thanks for checking. Can you just let them know on slack then that for the reduce-snapshot task we are going to maintain python going forward and not be using their script?

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-08-19T12:41:19.415Z |

Approved

midnightercz (approved):

